Issue #3, opened 2015-02-16
docker compose setup for development and testing


dnephin:
Hello,

I would like to contribute [Dockerfiles](https://www.docker.com/) and a [docker-compose.yaml](https://github.com/docker/fig) for interactive development and testing.  

There are a few advantages to a setup like this, notably:
- you can closely mimic production in development by having separate "hosts" for each service (memcache, postgres, etc). You can also ensure that versions and configuration of these services matches production configuration.
- you're able to get started with a project by just running a couple of commands (no manual steps to setup config and other services on your desktop.
  
  ```
  git clone ...
  docker-compose up
  ```

Is this something you would be interesting in accepting a pull request for?

benoitg:
I have no objection in principle.

However, not being familiar with docker, I have no idea how hard it would be for you to keep it in sync with our own version management (a combinasion of Fabric, Bower and pip)

dnephin:
It compliments bower and pip quite well. In a lot of ways it fulfils the same role as many of these fabric tasks.

Instead of duplicating them, I could have it just run the fabric task to do the installation of packages. The downside to that approach is that changes to the `fabfile.py` would require rebuilding the whole image, and an ssh server has to be installed in the containers.

I might start there and see how it looks.

benoitg:
There is no way around containers running a ssh server.  Several of the fabric tasks are crucial to the operation of an assembl server, not just it's installation.

dnephin:
Cool, I will go with that approach.  Since most of the fabric tasks split installation of dependencies from the build work, it should be possible to structure the `Dockerfile` so that rebuilds are pretty fast (unless the `fabfile.py` changes, which causes everything to be rebuilt).

jmatsushita:
Any interest in continuing this thread? The current complex installation setup might limit adoption. I know it prevents me from spinning up a test server and I'd really like to test! :)

maparent:
Good day, Jun! 

[...]

PR #4, opened 2015-02-21
docker compose

dnephin:
First steps for #3

I believe this setup is mostly working. I was able to launch the webapp and login (so the database is connected), but I only see an empty screen with "There are no discussions on this server yet", so I must be missing some fixture data.

I've still got more work to do here, but I wanted to send out a PR early to see if you had any feedback.

**Next steps**:
- start resolving the TODOs
- add a script that runs the test suite against this setup

**To use this setup**:

Install docker (requires 1.5): https://docs.docker.com/installation/ (for ubuntu I'd recommend https://docs.docker.com/installation/ubuntulinux/#docker-maintained-package-installation to get the latest)

To run the docker-compose containers

``` bash
virtualenv dc-venv
source dc-venv/bin/activate
pip install requests==2.4.3 docker-compose==1.1.0-rc2
docker build -t assembl_virtuoso -f dockerfiles/virtuoso/Dockerfile .
docker-compose build assembl
docker-compose up -d

# Get the ip of the webapp container
docker inspect -f '{{ .NetworkSettings.IPAddress }}' assembl_assembl_1
```

The `docker build` step is only necessary until https://github.com/docker/fig/issues/834 is fixed. A lot of this could be automated by using [tox](https://pypi.python.org/pypi/tox).


maparent:
Bonjour, Daniel!
Awesome work you did. I tried running it, and ran into some issues with the last step because my image's repository name includes my assembl directory's name (in my case assembl-feature) and other steps assume assembl. Do you think it's possible to fix that?
(I'll retry with the assembl directory name on my side.)

Here's extracts from the trace:

    $ docker-compose up -d
    ...
    Status: Downloaded newer image for redis:latest
    Creating assemblfeature_celeryimap_1...
    Pulling image assembl_assembl:latest...
    Pulling repository assembl_assembl
    Traceback (most recent call last):
      File "/Users/maparent/OpenSource/assembl-feature/dc-venv/bin/docker-compose", line 9, in <module>
        load_entry_point('docker-compose==1.1.0rc2', 'console_scripts', 'docker-compose')()
    ...
      File "/Users/maparent/OpenSource/assembl-feature/dc-venv/lib/python2.7/site-packages/compose/progress_stream.py", line 50, in print_output_event
        raise StreamOutputError(event['errorDetail']['message'])
    compose.progress_stream.StreamOutputError: Error: image library/assembl_assembl:latest not found
    ...
    $ docker images
    REPOSITORY               TAG                 IMAGE ID            CREATED             VIRTUAL SIZE
    assemblfeature_assembl   latest              6b3ddba1a59b        7 hours ago         3.788 GB
    assembl_virtuoso         latest              7110e557b8f7        7 hours ago         3.53 GB
    ...

As for your TODOs: I've added pandoc to the fabfile. I'm less sure about needing ruby-dev, since we build rbenv; I'll experiment more with that.
I'll look into separating the requirements in sections. That said, couldn't the database upgrade be done in the webapp layer instead? Then we wouldn't have to install the python machinery in the virtuoso layer. 
We may have to fix to a virtuoso commit, there are known bugs with 7.2.0.1. I'll get back to you on that.
I'll also look into tox (though I'm less sure when.)
As for being left with an empty screen, the issue is not fixtures but creating a basic superuser who can create discussions. Look for the assembl-add-user section in the readme.

We'll get there, and with many thanks for your hard work!

Best,
Marc-Antoine Parent

maparent:
OK, I did try within an assembl directory, and ran the process to completion.
However, the assembl_assembl process seems to only start one of the celery process, as opposed to the webapp. (At least that's what the command said in `docker ps`, and I could not access the assembl web app. It may be an artefact of my setup? I'm using boot2docker 1.5.0 on Mac OS X 10.10.2.)
More important, I get the impression from reading the dockerfile that you are starting pserve directly, and not through supervisord.  I understand that you're probably trying to avoid starting virtuoso. Tonight, I'll put preferences in the .ini file to control each supervisor process' startup independently, that should allow you to use supervisord.
(Even if there's no other python tool in the virtuoso layer, I suggest you use supervisord even there; virtuoso is prone to silent crashes, and it's important to be able to auto-restart it. You might avoid a virtualenv through using the supervisord from apt-get in that case.)
Cheers,
Marc-Antoine

maparent:
ok, something went wrong with my setup, and after a reboot I have distinct assembl_celeryimap_1 and assembl_assembl_1 processes. I still cannot access the application, but that's clearly because I still need to read up on docker networking. One mystery solved.
The notifications and change_router still seem to be missing, so I guess my supervisord comment stands, unless there is a fundamental reason to separate things that I do not know about?

paolina92:

PR #32, opened 2016-05-11
standalone docker image

dachary:
For test purposes it is convenient to have an all-in-one docker image, including redis, memcached and the database server. So that it can be run with:

    docker run --port 8000:80 --v /opt/assembl:/var/lib/database assembl
    firefox localhost:8000

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,50 @@
+FROM    ubuntu:16.04
+
+ENV     DEBIAN_FRONTEND noninteractive
+ARG	GITREPO=https://github.com/imaginationforpeople/assembl.git
+ARG	GITBRANCH=develop
+
+RUN     apt-get update && apt-get install -y \
+            fabric \
+            git \
+            openssh-server \
+            pandoc \
+            sudo \
+            net-tools \
+            monit \
+            nginx \
+            uwsgi \
+            uwsgi-plugin-python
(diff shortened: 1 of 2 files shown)


dachary:
This is different from #4 which separates redis, memcached, database in different containers.

dachary:
It does not actually work because virtuoso fails to bootstrap for reasons that are beyond my level of expertise. Since virtuoso is being abandoned for postgresql, it seems futile to try to resolve the problem. However, since the switch to postgresql is dated yesterday, it's likely that trying to make it work within a container will uncover some minor issues. It's best to postpone this work for later and re-open this pull request when things are stable.

dachary:
@maparent the build command in the comment is wrong (I was fixing it but you were quicker to merge than I expected ;-)

    sudo docker build --build-arg GITREPO=https://github.com/dachary/assembl.git --build-arg GITBRANCH=wip-docker --tag assembl --no-cache docker

dachary:
@maparent I tested this Dockerfile to work with

    sudo docker run -d -p 8080:80 --name assembl assembl

However, your original Dockerfile was also working. Whatever problem you had with it was environmental. The modifications I made to the Dockerfile are minor cleanups.

maparent:
Docker comment: sorry for being fast, maybe another pull request with the comment?
Test: obviously I had the wrong run command. It works with yours. Success, and bravo for being the first external contributor! I need to add a contributors file now ;-)

PR #40, opened 2016-08-26
Delete messages

swergas:
This corresponds to the task https://app.asana.com/0/158951227126092/158937659088697


assembl/static/js/app/common/collectionManager.js
```diff
@@ -412,8 +412,10 @@ var CollectionManager = Marionette.Object.extend({
       this.executeRequest = function() {
 
         var that = this,
-            allMessageStructureCollectionPromise = this.collectionManager.getAllMessageStructureCollectionPromise(),
             ids = [];
+        if ( messagesStructureCollectionPromise === undefined ){
```
paolina92: typeof messagesStructureCollectionPromise === "undefined" -> est plus prudent car passe sur plus de navigateurs

assembl/static/js/app/models/langstring.js
```diff
@@ -264,10 +280,21 @@ var LangString = Base.Model.extend({
    */
   original: function() {
     var originals = this.get("entries").filter(function(e) {return !e.isMachineTranslation();});
-    if (originals.length > 1) {
+    if ( originals.length == 1 ){
```
paolina92: plutôt ===

assembl/static/js/app/views/confirmModal.js
```diff
@@ -0,0 +1,53 @@
...
+      icon: this.icon,
+      contentText: this.contentText,
+      cancelText: this.cancelText,
+      submitText: this.submitText
+    };
+  },
+
+  // called by Backbone.Modal when users clicks on submitEl
+  submit: function(ev) {
+    if ( _.isFunction(this.onSubmit) ){
```
paolina92: J'ai pas compris le if pour tester si c'est une fn

assembl/static/js/app/views/message.js
```diff
@@ -600,7 +606,7 @@ var MessageView = Marionette.LayoutView.extend({
     }
 
 
-    if (this.model.get("publication_state") != "PUBLISHED") {
+    if (this.model.get("publication_state") == "MODERATED_TEXT_ON_DEMAND" || this.model.get("publication_state") == "MODERATED_TEXT_NEVER_AVAILABLE") {
```
paolina92: plutôt ===

assembl/static/js/app/views/message.js
```diff
@@ -646,6 +652,8 @@ var MessageView = Marionette.LayoutView.extend({
       html_export_url = Ctx.getApiV2DiscussionUrl("posts/" + this.model.getNumericId() + "/html_export");
     }
 
+    var user_can_delete_this_message = ( Ctx.getCurrentUserId() == Ctx.extractId(this.model.get('idCreator')) && Ctx.getCurrentUser().can(Permissions.DELETE_MY_POST) ) || Ctx.getCurrentUser().can(Permissions.DELETE_POST);
```
paolina92: ===

assembl/static/js/app/views/message.js
```diff
@@ -734,7 +744,7 @@ var MessageView = Marionette.LayoutView.extend({
     this.isCompleteDataLoaded();
     //Check if the message is moderated
     
-    if (this.model.get('publication_state') !== 'PUBLISHED'){
+    if (this.model.get('publication_state') == "MODERATED_TEXT_ON_DEMAND" || this.model.get("publication_state") == "MODERATED_TEXT_NEVER_AVAILABLE"){
```
paolina92: idem

assembl/static/js/app/views/message.js
```diff
@@ -749,250 +759,261 @@ var MessageView = Marionette.LayoutView.extend({
...
-          //Only show the translation view *iff* the message was translated by the backend
-          var translationView = new MessageTranslationView({messageModel: this.model, messageView: this});
-          this.translationRegion.show(translationView);
-          this.translationRegion.$el.removeClass("hidden");
-        } else if (this.translationRegion.$el) {
-          this.translationRegion.$el.addClass("hidden");
+      //Translation view should only be shown when the message is in full view or in preview mode. Otherwise,
+      //do not show it
+      //Also, only the body translation triggers the translation view
+      if (this.viewStyle == this.availableMessageViewStyles.FULL_BODY ||
```
paolina92: idem

assembl/static/js/app/views/message.js
```diff
@@ -749,250 +759,261 @@ var MessageView = Marionette.LayoutView.extend({
...
+              //console.log("dotdotdot initialized on message", that.model.id);
+              //console.log(isTruncated, orgContent);
+              if (isTruncated)
               {
-                if (Ctx.debugRender) {
-                  console.log("there may be a problem with the dotdotdot of message ", that.model.id, "so we will maybe re-render it");
-                }
-
-                if (++that.reRendered < 5) // we use this to avoid infinite loop of render() calls
+                that.$(".ellipsis > a.readMore, .ellipsis > p > a.readMore").removeClass('hidden');
```
paolina92: pas top le selecteur

assembl/static/js/app/views/message.js
```diff
@@ -749,250 +759,261 @@ var MessageView = Marionette.LayoutView.extend({
...
+              }
+              else
+              {
+                that.$(".ellipsis > a.readMore, .ellipsis > p > a.readMore").addClass('hidden');
```
paolina92: idem

assembl/static/js/app/views/message.js
```diff
@@ -749,250 +759,261 @@ var MessageView = Marionette.LayoutView.extend({
...
 
-                  setTimeout(function() {
-                    that.render();
-                  }, 500);
-                }
-                else
-                {
-                  if (Ctx.debugRender) {
-                    console.log("no, we won't re-render it because we already tried several times: ", that.reRendered);
+                  if (++that.reRendered < 5) // we use this to avoid infinite loop of render() calls
```
paolina92: ca m'a pas l'air très propre ça mais surement compliqué à corriger

assembl/static/js/app/views/message.js
```diff
@@ -749,250 +759,261 @@ var MessageView = Marionette.LayoutView.extend({
...
+                  {
+                    if (Ctx.debugRender) {
+                      console.log("yes, we will re-render => tries: ", that.reRendered);
+                    }
+
+                    setTimeout(function() {
```
paolina92: Pareil c'est pas super le settimout... Compliqué à corriger aussi j'imagine

assembl/static/js/app/views/message.js
```diff
@@ -749,250 +759,261 @@ var MessageView = Marionette.LayoutView.extend({
...
-          var current_navigation_state = that.messageListView.getContainingGroup().model.get('navigationState');
+        that.messageListView.requestPostRenderSlowCallback(function() {
 
-          //console.log("current_navigation_state:", current_navigation_state);
-          if (current_navigation_state === 'about')
-          {
-            that.listenToOnce(Assembl.vent, 'DEPRECATEDnavigation:selected', applyEllipsis);
-            return;
-          }
+          setTimeout(function() {
```
paolina92: idem

assembl/static/js/app/views/message.js
```diff
@@ -1524,6 +1545,52 @@ var MessageView = Marionette.LayoutView.extend({
...
+      ).then(function(data) {
+        Growl.showBottomGrowl(
+          Growl.GrowlReason.SUCCESS,
+          i18n.gettext('Message has been successfully deleted.'),
+          { delay: 12000 }
+        );
+
+        // Refresh the messageList
+        that.messageListView.render();
+        setTimeout(function(){
```
paolina92: Pourquoi un setTimeout ?

assembl/static/js/app/views/messageDeletedByUser.js
```diff
@@ -0,0 +1,118 @@
...
+    var that = this;
+
+    if ("subject" in options){
+      this.subject = options.subject;
+    }
+
+    if ("body" in options){
+      this.body = options.body;
+    }
+
```
paolina92: l'enchainement de la ligne 41 à la ligne 49 est bizarre

assembl/static/js/app/views/messageDeletedByUser.js
```diff
@@ -0,0 +1,118 @@
...
+    var agentNameView = new AgentViews.AgentNameView({
+      model: this.creator
+    });
+    this.name.show(agentNameView);
+  },
+
+  onRender: function(){
+    this.$el.attr("id", "message-" + this.model.get('@id'));
+    
+    if (this.template == "#tmpl-loader") {
```
paolina92: idem ===

assembl/static/js/app/views/messageFamily.js
```diff
@@ -158,11 +166,17 @@ var MessageFamilyView = Marionette.ItemView.extend({
...
     }
 
-    this._messageView = new messageViewClass({
-      model: this.model,
-      messageListView: this.messageListView,
-      messageFamilyView: this
-    });
+    var publication_state = this.model.get('publication_state');
+    if ( publication_state && publication_state in MessageModel.DeletedPublicationStates ){
+      if ( publication_state == MessageModel.PublicationStates.DELETED_BY_USER ){
```
paolina92: idem

assembl/static/js/app/views/confirmModal.js
swergas: Si parmi les paramètres à la construction de l'objet on a donné un onSubmit qui est une fonction, alors on peut l'exécuter.

assembl/static/js/app/views/message.js
swergas: c'est le code qu'il y avait avant, donc je serai partisan qu'un éventuel refactoring soit en dehors de la PR

assembl/static/js/app/views/message.js
swergas: c'est le code qu'il y avait avant, donc je serai partisan qu'un éventuel refactoring soit en dehors de la PR

assembl/static/js/app/views/message.js
swergas: c'est le code qu'il y avait avant, donc je serai partisan qu'un éventuel refactoring soit en dehors de la PR

assembl/static/js/app/views/messageDeletedByUser.js
swergas: Pourquoi ? Tu recommanderais quoi ? C'est un pattern qui a été utilisé à beaucoup d'endroits dans le code.

assembl/alembic/versions/41176a6a6758_tombstonable_content.py
```diff
@@ -0,0 +1,56 @@
...
+        for dp in dps2:
+            db.add(m.DiscussionPermission(
+                discussion_id = dp.discussion_id,
+                role_id = dp.role_id,
+                permission_id = p_delete_post.id))
+
+
+def downgrade(pyramid_env):
+    with context.begin_transaction():
+        op.drop_column('content', 'tombstone_date')
```
ayyazdaniaryan: It's good practice to do a proper downgrade, in case of system errors on upgrades. So removing the extra DiscussionPermissions would be ideal.

assembl/static/js/app/common/collectionManager.js
```diff
@@ -344,7 +344,7 @@ var CollectionManager = Marionette.Object.extend({
    * @returns {Promise}
    * @function app.common.collectionManager.CollectionManager.getMessageFullModelRequestWorker
   */
-  getMessageFullModelRequestWorker: function(collectionManager) {
+  getMessageFullModelRequestWorker: function(collectionManager, messagesStructureCollectionPromise) {
```
ayyazdaniaryan: @swergas Can you comment as to why this was now injected into the request worker? I knew this, and even I forgot :p

assembl/static/js/app/views/message.js
ayyazdaniaryan: @paolina92 I think this has to do with the scrollTo? Or maybe waiting for other components to render?

assembl/static/js/app/views/message.js
```diff
@@ -1524,6 +1546,52 @@ var MessageView = Marionette.LayoutView.extend({
...
+    // We could try to minimize context switching for the user, by scrolling the viewport to the message the user wants to delete, as soon as the confirmation popin opens, using this line of code:
+    // that.messageListView.scrollToMessage(that.model, false, false);
+    
+    var onSubmit = function(ev){
+      var analytics = Analytics.getInstance();
+      analytics.trackEvent(analytics.events.MESSAGE_LIKED);
+      // we could use that.model.destroy() instead, and add || method == "delete" to models/message.js::sync()
+      var message_delete_url = that.model.getApiV2Url();
+      Promise.resolve(
+        $.ajax(message_delete_url, {
```
ayyazdaniaryan: @swergas There was no other way of calling the DELETE route with backbone? Or was this method easier to resolve? If so, let's comment in the message model in the front-end that the backbone `Backbone.Model.destroy` method will not call this route, and instead, use the `$.ajax(url, {method: 'delete'})` method.

assembl/static/js/app/views/message.js
```diff
@@ -1524,6 +1546,52 @@ var MessageView = Marionette.LayoutView.extend({
...
+        );
+      });
+    };
+    var confirm = new ConfirmModal({
+      contentText: i18n.gettext('Are you sure you want to delete this message?'),
+      cancelText: i18n.gettext('No'),
+      submitText: i18n.gettext('Yes'),
+      onSubmit: onSubmit,
+    });
+    Assembl.slider.show(confirm);
```
ayyazdaniaryan: Should we use the `@app.common.context.setCurrentModalView` as well? Will the confirmation box be closed by any other view other than itself? If not, then no worries.

assembl/static/js/app/views/messageDeletedByUser.js
```diff
@@ -0,0 +1,118 @@
...
+  },
+
+  subject: "",
+  body: i18n.gettext("This message has been deleted by its author."),
+
+  initialize: function(options) {
+    var that = this;
+
+    if ("subject" in options){
+      this.subject = options.subject;
```
ayyazdaniaryan: `Backbone`y way of doing this is:
`this.subject = options.subject || null;`

assembl/static/js/app/views/messageListPostQuery.js
```diff
@@ -374,6 +389,33 @@ var PostQuery = function() {
       });
     };
 
+    this.getResultMessageStructureCollectionPromise = function() {
+      // The "deleted messages" filter needs a "all message structure collection" which is different from the default one (because the default one may not include all deleted messages)
+      // Maybe we should use a Subset of the getAllMessageStructureCollectionPromise instead
+      // TODO: Have a more generic mecanism than a list of ifs. For example add a usesNonStandardMessageStructure boolean method in AbstractPostFilter
+      if ( ("only_deleted_posts" in this._query) || ("also_deleted_posts" in this._query) ){
```
ayyazdaniaryan: I would highly recommend making an `enum` type object that contains these strings. Will keep the code DRY, and we can document the strings wherever they are

assembl/views/api/post.py
```diff
@@ -168,6 +168,66 @@ def get_posts(request):
...
+    # if deleted == 'true':
+    #     posts = posts.filter(PostClass.not_tombstone_condition())
+    #     ideaContentLinkQuery = ideaContentLinkQuery.filter(PostClass.not_tombstone_condition())
+    # end v3
+
+    # v4
+    deleted = request.GET.get('deleted', None)
+    if deleted is None:
+        if not ids:
+            deleted = 'false'
```
ayyazdaniaryan: Again, `enum` types. Any time we use strings, maintainability becomes a hassle. Make a `dict` type that contains these strings, document their usages/cases. And call the dict keys instead.

assembl/views/api/post.py
```diff
@@ -313,16 +373,54 @@ def get_posts(request):
...
     else:
         posts = posts.order_by(Content.id)
-    print str(posts)
+    # print str(posts)
 
     no_of_posts = 0
     no_of_posts_viewed_by_user = 0
 
+
+    def post_and_its_whole_line_of_descent_are_deleted(post):
```
ayyazdaniaryan: Add docstrings like:
`
"""
Counts the root post and every post that is it's children, AND their children (therefore, the entire tree of posts under the root post) are deleted or not.
:param: post: The root post to count from
:type: post: assembl.models.Content
:rtype: boolean
"""
`

Very useful to start documenting the back-end as we go along in Sphinx style RST in order to improve overall code documentation. See here for more examples: http://thomas-cokelaer.info/tutorials/sphinx/docstring_python.html

assembl/views/api/post.py
```diff
@@ -313,16 +373,54 @@ def get_posts(request):
...
+            return False
+        children = post.children
+        if len(children) == 0:
+            return True
+        for child in children:
+            if not post_and_its_whole_line_of_descent_are_deleted(child):
+                return False
+        return True
+
+    def post_or_one_of_line_of_descent_is_deleted(post):
```
ayyazdaniaryan: Same. Document key methods :D

assembl/views/api2/post.py
```diff
@@ -39,6 +39,44 @@ def show_similar_posts(request):
...
+    for extract in extracts_to_remove:
+        extract.delete()
+
+    
+    if user_id == instance.creator_id and P_DELETE_MY_POST in permissions:
+        instance.publication_state = PublicationStates.DELETED_BY_USER
+    elif P_DELETE_POST in permissions:
+        instance.publication_state = PublicationStates.DELETED_BY_ADMIN
+
+    instance.is_tombstone = True
```
ayyazdaniaryan: Let's make this a method on the instance? Also...does the view_config save the model and commit to db after the view has been completed?

assembl/views/api2/post.py
```diff
@@ -39,6 +39,44 @@ def show_similar_posts(request):
...
+
+    return {
+        "result": "Post has been successfully deleted.",
+        "removed_extracts": number_of_extracts
+    }
+
```
ayyazdaniaryan: Based on another comment above, if we don't do a `$.ajax` call to this view, and instead have `Backbone.Model.destroy` make this call, then is this affected at all?

assembl/static/js/app/views/messageDeletedByUser.js
manthis: Peux-tu donner plus de précision sur ton commentaire @paolina92 ?

swergas: OK je vais utiliser https://github.com/assembl/assembl/pull/40#discussion_r76826783

assembl/alembic/versions/41176a6a6758_tombstonable_content.py
swergas: Thanks for spotting it! I will add this.

assembl/static/js/app/views/message.js
swergas: I wrote it this way first and we thought about the other way togeter afterwards, both should work I guess, but I thought there is a small risk that this becomes a bit tricky and produce unanticipated effects. So as we are a bit in a hurry, I prefered to keep it as is rather than redo full AQ. Let's talk about it together when we have a moment!

assembl/static/js/app/views/message.js
swergas: No I think it's OK we don't need it.

assembl/static/js/app/views/message.js
ayyazdaniaryan: No problem @swergas. As you wish :)

assembl/alembic/versions/41176a6a6758_tombstonable_content.py
```diff
@@ -0,0 +1,74 @@
...
+import sqlalchemy as sa
+import transaction
+
+
+def upgrade(pyramid_env):
+    with context.begin_transaction():
+        op.add_column('content', sa.Column('tombstone_date', sa.DateTime))
+
+
+    from assembl import models as m
```
manthis: Wouldn't it be better to declare imports at the beginning of file for readability purposes?

assembl/alembic/versions/41176a6a6758_tombstonable_content.py
```diff
@@ -0,0 +1,74 @@
...
+
+def downgrade(pyramid_env):
+    # Remove tombstone_date column from content table
+    with context.begin_transaction():
+        op.drop_column('content', 'tombstone_date')
+
+
+    # Remove P_DELETE_MY_POST and P_DELETE_POST permissions, as well as their appearances in DiscussionPermission table (their activation on roles)
+    from assembl import models as m
+    from assembl.auth import P_DELETE_MY_POST, P_DELETE_POST
```
manthis: Wouldn't it be better to declare imports at the beginning of file for readability purposes?

assembl/models/idea.py
```diff
@@ -999,11 +999,30 @@ class RootIdea(Idea):
 
     @property
     def num_posts(self):
-        """ In the root idea, this is the count of all mesages in the system """
+        """ In the root idea, num_posts is the count of all non-deleted mesages in the discussion """
+        from .post import Post
```
manthis: Wouldn't it be better to declare imports at the beginning of file for readability purposes?

assembl/models/idea.py
```diff
@@ -999,11 +999,30 @@ class RootIdea(Idea):
...
+            Post.discussion_id == self.discussion_id,
+            Post.hidden==False,
+            Post.tombstone_condition()
+        ).count()
+        return int(result)
+
+    @property
+    def num_read_posts(self):
+        """ In the root idea, num_posts is the count of all non-deleted read mesages in the discussion """
         from .post import Post
```
manthis: Wouldn't it be better to declare imports at the beginning of file for readability purposes?

assembl/static/js/app/views/message.js
```diff
@@ -720,7 +730,8 @@ var MessageView = Marionette.LayoutView.extend({
     },
 
   render: function() {
-    //Is this for inheritence??
+    //This code was used to get the Marionette Render method.
+    //TODO: Check to see if this is still valid code. AY
```
manthis: TODO to be checked and removed priori to merging

assembl/static/js/app/views/message.js
```diff
@@ -749,250 +760,261 @@ var MessageView = Marionette.LayoutView.extend({
...
+        console.log("isMessageTranslated: ", this.isMessageTranslated);
+        console.log("Discrepency? ", !(this.isMessageTranslated !== this.showAnnotations));
+        console.log("_body.value: ", this._body.value());
+        console.log("_body.isMachineTranslation(): ", this._body.isMachineTranslation());
+      }
 
-    if (Ctx.debugAnnotator) {
-      console.log("showAnnotations: ", this.showAnnotations);
-    }
+      if (Ctx.debugAnnotator) {
```
manthis: Prefer a specific branch for esthetic commits

assembl/static/js/app/views/message.js
```diff
@@ -749,250 +760,261 @@ var MessageView = Marionette.LayoutView.extend({
...
+        console.log("showAnnotations: ", this.showAnnotations);
+      }
 
-    if (!this.showAnnotations) {
-      this.removeAnnotations();
-    }
+      if (!this.showAnnotations) {
+        this.removeAnnotations();
+      }
```
manthis: Idem

assembl/static/js/app/views/message.js
```diff
@@ -749,250 +760,261 @@ var MessageView = Marionette.LayoutView.extend({
...
-    // do not render the whole thing if only the like_count changed.
-    // it may kill the message being edited.
-    if (this.changeIsPartialRender()) {
-      return;
-    }
+      // do not render the whole thing if only the like_count changed.
+      // it may kill the message being edited.
+      if (this.changeIsPartialRender()) {
+        return;
+      }
```
manthis: Idem

assembl/static/js/app/views/message.js
manthis: Agreed, refactoring is out of PR

assembl/static/js/app/views/message.js
manthis: Out of PR

assembl/static/js/app/views/messageListPostQuery.js
manthis: Agreed

assembl/views/api/post.py
```diff
@@ -313,16 +373,68 @@ def get_posts(request):
...
+        :returns: True if post or one of its line of descent (direct and indirect answers) is deleted, False otherwise.
+        :rtype: bool
+        :param post: The post you want to analyse
+        :type post: assembl.models.Post
+        
+        """
+        if post.is_tombstone:
+            return True
+        children = post.children
+        if len(children) == 0:
```
manthis: I think this test is useless. When they are no children, you won't enter the loop and consequently False will be returned.

assembl/views/api/post.py
```diff
@@ -313,16 +373,68 @@ def get_posts(request):
...
+
     for query_result in posts:
         score, viewpost, likedpost = None, None, None
         if not isinstance(query_result, (list, tuple)):
             query_result = [query_result]
         post = query_result[0]
+
+        # The response should not include deleted posts which do not break the structure of threads (these are deleted posts which have not received any direct or indirect non-deleted answer)
+        # Look for non-deleted (direct or indirect) children, and if there is none, we know that removing it from the results will not break the structure
+        # TODO: use recursion at the SQL query level instead
```
manthis: Is this TODO something we can do during PR? Or should it become a User Story of its own?

assembl/views/api/post.py
```diff
@@ -313,16 +373,68 @@ def get_posts(request):
...
+        ignore_this_post = False
+        if deleted == 'false':
```
manthis: Not sure to understand why we use Strings instead of boolean. Is there a reason I can't see yet? In case we would use booleans, the second test would be useless and we could just replace it by "else".

assembl/models/idea.py
maparent: Not sure in this case, but imports in-function are often done to avoid circular imports.

manthis: When trying to avoid circular import you usually move your imports to the end of the file: http://effbot.org/zone/import-confusion.htm (see circular imports). Or maybe I don't understand how you prevent circular import this way. In this case, I'd be interested to understand how you do that. Besides, in my point of view, using an import in the middle of a function will cause that function to take longer. And finally, all python style guides encourage to define imports at the beginning of a file. But discussion is opened on this topic.

maparent: a) Import in the middle of a file does not take longer in Python. It is executed once. I have seen this style in the wild.
b) not familiar with moving imports at the end, worth trying, could be cleaner.

benoitg: It not always faster or slower.  Like many things in life, it depends ;)

Import in the function body is going to be faster (and possibly consume a lot les memory) if it avoids some modules being loaded (lazy loading modules) but significantly slower if the function is small and executed in a tight loop (there is function call overhead from the import statement even if the module is loaded only once).

In any event, we did it to avoid circular loading.  Sqlalchemy declarative models requires some hard to predict module loading orders.

Should we decide to make it a standard to not load module in function calls, we must move ALL imports to the end, otherwise we will have NEITHER the readability/documentation advantages of grouped imports, OR the lazy loading/documentation locality advantages of in function imports.

manthis: Thanks for this explanation. It makes things clearer. I do agree with you that in order to prevent circular import, it is better to avoid declaring imports at the beginning of the file. So the two options are either in the function bodies or at the end of the file. I do believe that you guys had good reasons to do it this way. So no problem for me to leave it like as is. Thanks guys.

PR #41, opened 2016-09-06
Idea panel attachment

ayyazdaniaryan:
https://app.asana.com/0/26508243143256/158193194307145

Things noted to not work after merging develop onto idea-panel-attachment:
- The arrow dropdown menu gives more options than just delete on ideal panel image
- The image is stretched, instead of static

Known issue:
- The arrow on an image should NOT show when the user has not the edit_idea permission
- The image should not be clickable in an idea panel


assembl/static/js/app/models/attachments.js
```diff
@@ -214,8 +215,12 @@ var AttachmentModel = Base.Model.extend({
         that = this;
     return d.destroy({
       success: function(model, response){
-        console.log('in document destroy success callback');
         return Base.Model.prototype.destroy.call(that);
+      },
+      error: function(model, response){
+        console.warn("Could NOT destroy document model id " + d.id
+                     + "\nContinuing to destroy attachment id " + this.id);
+        return Base.Model.prototype.destroy.call(that);
```
paolina92: May be it's better (in the futur) to let the view handle the model's error and display an error message in order to inform the user.

assembl/static/js/app/models/attachments.js
```diff
@@ -370,8 +393,152 @@ var AttachmentCollection = Base.Collection.extend({
...
+        });
+      }
+
+      correctedNumberModelsCollection = correctedNumberModelsCollection.collection;
+    }
+
+    _.each(correctedNumberModelsCollection, function(model){
+      if (!that.isTypeLimitCorrect(model)){
+        model.typeLimitReached = true;
+        Growl.showBottomGrowl(
```
paolina92: May be it's better to use Growl in the View, not in the model?

ayyazdaniaryan: I would have liked that. But, at the time of the attachment model creation, the attachment view for that model is not yet created. So, during this check, the view isn't created. Unless we have a parent Event Aggregator at the topmost view, and THAT view would listen for an event from these models, and that view throws a growl.
This was a quick win to show a growl from the model, given that the growl is not bound to the attachment view. Please discuss.

assembl/static/js/app/models/attachments.js
ayyazdaniaryan: Hum...that's interesting. This error callback still tries to destroy the attachment model, even if the document model is not destroyed. So, from user perspective, the attachment would not be visible anymore. `console.warn` will be caught by Sentry, and that is something that we as developers would see.

But, I see your point. If the server is down, and the document model is not destroyed, AND attachment model is also not, then a growl or something should be shown to the user. Maybe now a good time to figure out a way to throw events at an aggregator?

assembl/static/js/app/views/ideaPanel.js
```diff
@@ -125,6 +155,40 @@ var IdeaPanel = AssemblPanel.extend({
...
+    }
+    else{
+      domObject.css('top', imgHeight);
+    }
+  },
+  /*
+    Manages the spacing at the top of the ideaPanel, depending on the panel having an
+    attachment or not.
+   */
+  checkContentHeight: function(){
```
swergas: Have you tried obtaining the same results in terms of image positioning/sizing/cropping using CSS only (no javascript)? I feel it would be much more robust and clean than using javascript.

assembl/static/js/app/views/attachments.js
```diff
@@ -261,6 +270,54 @@ var AttachmentFileEditableView = AttachmentEditableView.extend({
...
+    //to save the attachment
+    this.model.save();
+  },  
+
+  /*
+    There is a limit of 1 attachment, so this *should* only be called once
+   */
+  onChange: function(e){
+    var domObject = $(".content-ideapanel");
+    this.$el.find(".embedded-image-preview").load(function() {
```
swergas: Why not factorize this code with the code in checkContentHeight()? (or use pure CSS, see my other comment)

assembl/static/js/app/views/attachments.js
```diff
@@ -183,12 +189,13 @@ var AttachmentEditableView = AbstractAttachmentView.extend({
   _renderAttachmentPurpose: function(){
     var purposesHtml = [],
         that = this;
-    _.each(Attachments.attachmentPurposeTypes, function(attachmentPurposeDef) {
-      purposesHtml.push('<li><a class="js_attachmentPurposeDropdownListItem" data-id="' + attachmentPurposeDef.id + '" data-toggle="tooltip" title="" data-placement="left" data-original-title="' + attachmentPurposeDef.id + '">' + attachmentPurposeDef.label + '</a></li>');
-    });
-
+    if(this.model.get('@type') !== 'IdeaAttachment'){
```
swergas: I'm not very comfortable with this. To recap: To me the view of the image attached to the idea panel and its UX are so different from the "regular" attachment view which is used in a message (and which may be used in other places of Assembl where we could upload a list of files), that it needs its own view (which could inherit from a common parent). I guess this line of code which displays things differently depending on the type of the object which uses the view should be avoided in general, and this is an epiphenomenon of the problem. Others, any opinion?

assembl/static/js/app/views/attachments.js
```diff
@@ -261,6 +270,54 @@ var AttachmentFileEditableView = AttachmentEditableView.extend({
...
+    this.$el.find(".embedded-image-preview").load(function() {
+      var contentPanelPosition = $(window).height() / 3;
+      var imgHeight = $(this).height();
+      if(imgHeight > contentPanelPosition){
+        domObject.css('top', contentPanelPosition);
+      }else{
+        domObject.css('top', imgHeight);
+      }
+    });
+    $('[data-id="DO_NOT_USE"]').hide();
```
swergas: Same comment as https://github.com/assembl/assembl/pull/41/files#r77958007

PR #42, opened 2016-09-30
Deleted post count

paolina92:
This pull request is related to: 

- Idea deletion User Story (Pauline) [https://bluenove.atlassian.net/browse/DEVAS-23](https://bluenove.atlassian.net/browse/DEVAS-23)
- Bug fixing on deleted post count (Maparent)  [https://bluenove.atlassian.net/browse/DEVAS-27](https://bluenove.atlassian.net/browse/DEVAS-27)

### DIFF
--- a/assembl/static/js/app/views/message.js
+++ b/assembl/static/js/app/views/message.js
@@ -1534,9 +1534,28 @@ var MessageView = Marionette.LayoutView.extend({
     this.listenToOnce(this.messageModerationOptionsView, 'moderationOptionsSaveAndClose', this.onModerationOptionsSaveAndClose);
     this.listenToOnce(this.messageModerationOptionsView, 'moderationOptionsClose', this.onModerationOptionsClose);
   },
-
+  setNumPostAfterMessageDeletion:function(){
+    var that = this;
+    var currentIdeaId = "";
+    var ideaContentLinks = that.model.get('indirect_idea_content_links');
+    _.each(ideaContentLinks, function(ideaContentLink){
+      console.log(ideaContentLink);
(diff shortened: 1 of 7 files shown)


assembl/static/js/app/views/message.js
```diff
@@ -1534,9 +1534,17 @@ var MessageView = Marionette.LayoutView.extend({
     this.listenToOnce(this.messageModerationOptionsView, 'moderationOptionsSaveAndClose', this.onModerationOptionsSaveAndClose);
     this.listenToOnce(this.messageModerationOptionsView, 'moderationOptionsClose', this.onModerationOptionsClose);
   },
-
+  fetchIdeasCollection:function(){
```
ayyazdaniaryan: This method's contract can be better. () => Promise<ideaCollection>, for example. Having a side-effectful method like this would make us rely on setTimeouts with random time frames...

assembl/static/js/app/views/message.js
```diff
@@ -1557,7 +1565,7 @@ var MessageView = Marionette.LayoutView.extend({
           i18n.gettext('Message has been successfully deleted.'),
           { delay: 12000 }
         );
-
+        that.fetchIdeasCollection();
         // Refresh the messageList
         that.messageListView.render();
         setTimeout(function(){
```
ayyazdaniaryan: Ah...yes, here it is. A setTimeout of 500ms. How come? We should make the `fetchIdeasCollection` promisified. And in the `then` method, have the messageList re-render and `showMessageById`.

swergas:
And with the current state of the branch, the second message does not show.

PR #43, opened 2016-10-04
Added GPG encryption for database backups.

manthis:
I just implemented database encryption through GPG. The use of GPG is optional thanks to the use_gpg fabric task parameter.


manthis:
Needs QA

PR #44, opened 2016-10-05
Reviewed version of the fabfile

manthis:
The fabfile has been cleaned from virtuoso code, dead code and is now compliant to PEP8.


manthis:
Needs QA

PR #49, opened 2016-10-13
Deleted post count

swergas:
I create a pull request of @maparent branch so that we continue to match github flow and to provide a container for code review comments.

### DIFF
--- a/assembl/models/path_utils.py
+++ b/assembl/models/path_utils.py
@@ -221,6 +221,14 @@ class PostPathLocalCollection(object):
 
     def as_clause_base(self, db, include_breakpoints=False,
                        include_deleted=False):
+        """Express collection as a SQLAlchemy query clause.
+
+        :param bool include_breakpoints: Include posts where
+            a threadbreak happens
+        :param include_deleted: Include posts in deleted_publication_states.
+            True means only deleted posts, None means all posts,
+            False means only live posts or deleted posts with live descendants.
+        """
         assert self.reduced
         def base_query(labeled=False):
             post = with_polymorphic(
--- a/assembl/models/path_utils.py
+++ b/assembl/models/path_utils.py
@@ -526,7 +526,8 @@ class PostPathCounter(PostPathCombiner):
         return (post_count, viewed_count)
 
     def get_orphan_counts(self, include_deleted=False):
-        return self.get_counts_for_query(self.orphan_clause(self.user_id))
+        return self.get_counts_for_query(
+            self.orphan_clause(self.user_id, include_deleted=include_deleted))
 
     def end_visit(self, idea_id, level, result, child_results):
         if isinstance(idea_id, Idea):
--- a/assembl/models/path_utils.py
+++ b/assembl/models/path_utils.py
@@ -12,7 +12,7 @@ from sqlalchemy.sql.functions import count
 
 from .idea_content_link import (
     IdeaContentLink, IdeaContentPositiveLink, IdeaContentNegativeLink)
-from .post import Post, Content, SynthesisPost
+from .post import Post, Content, SynthesisPost, countable_publication_states
 from .annotation import Webpage
 from .idea import IdeaVisitor, Idea, IdeaLink, RootIdea
 from .discussion import Discussion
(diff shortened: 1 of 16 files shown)


assembl/models/post.py
```diff
@@ -57,6 +57,8 @@ class PublicationStates(DeclEnum):
     MODERATED_TEXT_NEVER_AVAILABLE = "MODERATED_TEXT_NEVER_AVAILABLE", ""
     DELETED_BY_USER = "DELETED_BY_USER", ""
     DELETED_BY_ADMIN = "DELETED_BY_ADMIN", ""
+    WIDGET_SCOPED = "WIDGET_SCOPED", ""
```
maparent: There is a TODO note in generic.py:
# TODO: Refactor hidden into PublicationStates.WIDGET_SCOPED
If the type is left out, the TODO is harder to understand; I think the TODO is meaningful. Of course the TODO would be better expressed as a task in jira.

assembl/models/path_utils.py
```diff
@@ -480,15 +516,16 @@ def get_counts(self, idea_id):
...
+            include_deleted=None)
         (post_count, viewed_count) = self.get_counts_for_query(q)
         (path_collection.count, path_collection.viewed_count) = (
             post_count, viewed_count)
         self.counts[idea_id] = post_count
         self.viewed_counts[idea_id] = viewed_count
         return (post_count, viewed_count)
 
-    def get_orphan_counts(self):
+    def get_orphan_counts(self, include_deleted=False):
```
maparent: well spotted, added use of the parameter.

assembl/models/post.py
```diff
@@ -348,6 +357,36 @@ def has_next_sibling(self):
             return self != self.parent.children[-1]
         return False
 
+    @property
+    def has_live_child(self):
+        for child in self.children:
+            if child.publication_state not in deleted_publication_states:
+                return True
+
+    def delete_post(self, cause):
```
maparent: Thank you, corrected.

assembl/models/path_utils.py
```diff
@@ -230,9 +233,13 @@ def base_query(labeled=False):
                 query = db.query(post.id.label("post_id"))
             else:
                 query = db.query(post.id)
-            query = query.join(
-                content, (content.id == post.id) &
-                         (content.tombstone_date == None))
+            query = query.join(content, content.id == post.id)
```
maparent: documented in post.delete_post, and PostPath.as_clause_base. Tell me if intelligible at this point.

assembl/models/post.py
maparent: Added comments in delete_post.

assembl/models/path_utils.py
```diff
@@ -306,17 +321,28 @@ def base_query(labeled=False):
...
+            if include_deleted:
+                post = with_polymorphic(
+                    Post, [], Post.__table__,
+                    aliased=False, flat=True)
+                q = q.join(
+                    post, (post.id == content.id) &
+                    post.publication_state.in_(deleted_publication_states))
+            else:
+                q = q.filter(content.tombstone_date == None)
+
```
ayyazdaniaryan: @maparent Why is the `with_polymorphic` done on `Post` here rather than `Content`? That is unclear to me.

assembl/models/path_utils.py
```diff
@@ -456,16 +492,24 @@ def copy_result(self, idea_id, parent_result, child_result):
         self.viewed_counts[idea_id] = parent_result.viewed_count
 
     def get_counts_for_query(self, q):
+        # HACKITY HACK
```
ayyazdaniaryan: Perhaps a docstring with `:params`, I'm guessing that the `q` input is a query?

assembl/tests/fixtures/posts.py
```diff
@@ -147,3 +147,55 @@ def fin():
...
+    test_session.flush()
+    p.set_parent(reply_deleted_post_4)
+    test_session.flush()
+
+    def fin():
+        print "finalizer reply_to_deleted_post_5"
+        test_session.delete(p)
+        test_session.flush()
+    request.addfinalizer(fin)
+    return p
```
ayyazdaniaryan: Can I just say ❤️ !

swergas:
I spotted a regression: in the current state of this branch, messages which have been deleted do not show anymore when the "Show also deleted messages" or the "Deleted messages" filter is active in the messageList.

assembl/tests/model_tests/test_path_utils.py
```diff
@@ -90,5 +90,55 @@ def as_post_nums(path):
...
+    assert not reply_post_1.is_tombstone
+    # ...until child is deleted
+    reply_post_2.delete_post(PublicationStates.DELETED_BY_ADMIN)
+    assert reply_post_2.publication_state == PublicationStates.DELETED_BY_ADMIN
+    assert reply_post_2.is_tombstone
+    assert reply_post_1.is_tombstone
+
+
 if Post.using_virtuoso:
     test_jack_layton_linked_discussion = pytest.mark.xfail(test_jack_layton_linked_discussion)
```
ayyazdaniaryan: Let's add some more API v1 tests to ensure that the regressions that we are seeing are covered?

Eg.
Test API v1 returning the whole structure of posts and checking for the correct message count, and correct messages to appear?

maparent:
As far as I can see, the deleted messages are shown or not according to the flag. Their content is never shown, as I thought was agreed. We need to discuss.

assembl/models/path_utils.py
maparent: This is outside the scope of the pull request. I'd be happy to document the whole thing, but what's Bluenove budget for that?

PR #50, opened 2016-10-22
fix SourceReader import was blocked when MailingList encountered empty HTML document

swergas:
fix SourceReader import was blocked when MailingList encountered empty HTML document

(I was trying to import emails from a mailing list into a discussion, using the mailing list feature)

### DIFF
--- a/assembl/models/mail.py
+++ b/assembl/models/mail.py
@@ -288,8 +288,8 @@ class AbstractMailbox(PostSource):
         doc = None
         try:
             doc = html.fromstring(message_body)
-        except: # We can get "ParserError: Document is empty", which has to be caught, otherwise it blocks the SourceReader
-            return message_body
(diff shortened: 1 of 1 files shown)


assembl/models/mail.py
```diff
@@ -288,8 +288,8 @@ def strip_full_message_quoting_html(message_body):
         doc = None
         try:
             doc = html.fromstring(message_body)
-        except: # We can get "ParserError: Document is empty", which has to be caught, otherwise it blocks the SourceReader
-            return message_body
+        except etree.ParserError: # If the parsed HTML document is empty, we get a "ParserError: Document is empty" exception. So the stripped message we return is an empty string (if we keep the exception it blocks the SourceReader)
+            return ""
```
benoitg: That will silently mask any and all parsing errors!  The processing loop should catch any unhandled exception, log it, skip the message (not import it empty), and continue.

PR #51, opened 2016-10-22
Message source manual start forces restart

swergas:
In the "Discussion parameters" page, "Server settings" section, the "manual start" link on a source was not starting a fetch if the source was in at least one of the error states. Now the fetch is really forced.

### DIFF
--- a/assembl/static/js/app/views/admin/generalSource.js
+++ b/assembl/static/js/app/views/admin/generalSource.js
@@ -110,8 +110,8 @@ var ReadSource = Marionette.ItemView.extend({
         url,
         {
           method: "POST",
-          contentType: "application/json",
-          data: payload
+          contentType: "application/json; charset=UTF-8",
+          data: JSON.stringify(payload)
         }
       ).then(function() {
         Growl.showBottomGrowl(Growl.GrowlReason.SUCCESS, i18n.gettext('Import has begun!'))
--- a/assembl/static/js/app/views/admin/generalSource.js
+++ b/assembl/static/js/app/views/admin/generalSource.js
@@ -8,6 +8,7 @@ var Marionette = require('../../shims/marionette.js'),
     i18n = require('../../utils/i18n.js'),
     Types = require('../../utils/types.js'),
     Ctx = require('../../common/context.js'),
+    Permissions = require('../../utils/permissions.js'),
     Growl = require('../../utils/growl.js'),
     Source = require('../../models/sources.js'),
     CollectionManager = require('../../common/collectionManager.js'),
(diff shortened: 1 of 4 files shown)


maparent:
J'approuve, merci!

assembl/static/js/app/views/admin/generalSource.js
```diff
@@ -110,8 +110,8 @@ var ReadSource = Marionette.ItemView.extend({
         url,
         {
           method: "POST",
-          contentType: "application/json",
-          data: payload
+          contentType: "application/json; charset=UTF-8",
```
ayyazdaniaryan: @swergas Why set the `charset=UTF-8`?

assembl/views/api2/content_source.py
```diff
@@ -22,11 +22,17 @@
...
     csource = ctx._instance
-    force_restart = request.params.get('force_restart', False)
-    reimport = request.params.get('reimport', False)
-    reprocess = request.params.get('reprocess', False)
-    upper_bound = request.params.get('upper_limit', None)
-    lower_bound = request.params.get('lower_limit', None)
+
+    def request_parameter(key, default_value=None):
+        # when client POSTs data as JSON, in Pyramid data does not arrive in request.POST but in request.json_body
+        return request.json_body.get(key, default_value)
```
ayyazdaniaryan: I recall that the payload of an application/json comes in json_body, but everything in the `request.GET` and `request.POST` body come in the `request.params`.

ayyazdaniaryan: Ooooh, I see. You are following the pattern they suggest here @swergas: http://docs.pylonsproject.org/projects/pyramid/en/latest/narr/webob.html#request-json-body

Okay, works for me.

assembl/views/api2/content_source.py
```diff
@@ -55,6 +61,9 @@ def fetch_posts(request):
             raise HTTPUnauthorized("Only discussion administrator\
                                    can "+'and'.join(requested))
 
+    if force_restart:
```
ayyazdaniaryan: Is the `fetch_post` method responsible for calling to reset all errors in the `ContentSource`? It's a bit spaghetti, non? Could be further down the call chain, check for the force_restart and have the `ContentSource` take care of its own `reset_errors()`. A thought, I'm not picky.

ayyazdaniaryan: @maparent This will likely require your eyes. We looked into moving the check to `force_restart` and doing the `reset_errors` downstream. We see that it *could* be done in SourceReader class, but we're unsure where would be an ideal place to do a check like this. Can you assist on where would be the best place to put a check like this?

assembl/views/api2/content_source.py
maparent: Yes... trying to remember WHY this was sent as JSON, and I got nothing. Maybe I was trying to standardize POST=>body, GET=>query parameters. But even that does not justify Json.
Beginning to think that just maybe we could use query parameters here, or even a standard post body, and simplify everything.

assembl/views/api2/content_source.py
maparent: What you're proposing makes a lot of sense to me. No objection at all to moving it downstream.

assembl/views/api2/content_source.py
maparent: Ah, I remembered... it's actually a constraint of it being an ajax call on the JS side. So yes, it has to be done this way.

PR #52, opened 2016-10-26
Openstack

maparent:
Many elements were postponed, so I propose merging work in progress now.


assembl/tasks/email_discussion_creator.py
```diff
@@ -0,0 +1,83 @@
...
+        sender_name = discussion.topic
+        sender_name = UnicodeDammit(sender_name).unicode_markup
+        # sanitize
+        sender_name = re.sub(
+            ur"[^-\w\s'\u2019\u2032\u00b4\.\(\)]", '', sender_name, 0, re.UNICODE)
+        sender = '"%s" <%s>' % (sender_name, sender_email)
+        sender_name = Header(sender_name, 'utf-8').encode()
+        if len(sender) > 255:
+            sender = sender_email
+        subject = localizer.translate(_("Your consultation was created"))
```
ayyazdaniaryan: Is this copy verified? Would have thought it would be closer to `[PlatformName] Congratulations! Your consultation was approved/created/whatever` ? Where `PlatformName` 99% of the time is Assembl.

assembl/tasks/piwik.py
```diff
@@ -0,0 +1,321 @@
...
+        missing_variables.append("piwik_api_token")
+    if len(missing_variables):
+        raise RuntimeError("missing configuration variables: " + ", ".join(missing_variables))
+
+    # TODO: Should this process first check that discussion.web_analytics_piwik_id_site is empty and do something different if it's not? (for example: return an error, so that we empeach discussion statistics to be scattered on different Piwik sites, which is difficult to merge)
+
+    try:
+        # Check wether a Piwik user with a `user_email` email exists
+        try:
+            user_already_exists = piwik_UsersManager_getUserByEmail(piwik_url, piwik_api_token, user_email)
```
ayyazdaniaryan: Ooof. The naming of this function breaks the backend coding style that we've picked. Python naming convention does not include camelCase. Maybe change to `get_user_by_email`?

assembl/tasks/piwik.py
```diff
@@ -0,0 +1,321 @@
...
+        user_password = ""
+        user_login = user_email
+
+        if not user_already_exists:
+            # Create a Piwik user with `user_email` as login and as email
+            user_password = string_generator(size=10)
+            user_created = piwik_UsersManager_addUser(piwik_url, piwik_api_token, user_email, user_password, user_email)
+            if not user_created:
+                # Try to find if creation failed because of rare/edge case of a Piwik user already existing with the user_email as login but not as email
+                print("##### user not created, trying to find why")
```
ayyazdaniaryan:
```
import logging

log  = logging.getLogger('assembl')
log.warn("user not created, trying to find why")
```

ayyazdaniaryan: We want a trace of this Post Mortem if something goes wrong...

assembl/tasks/piwik.py
```diff
@@ -0,0 +1,321 @@
...
+                user_with_email_as_login_exists = piwik_UsersManager_userExists(piwik_url, piwik_api_token, user_email)
+                if user_with_email_as_login_exists:
+                    # We will use this strange Piwik user
+                    user_login = user_email
+                    print("##### we are in the rare case of a Piwik user already existing with the user_email as login but not as email")
```
ayyazdaniaryan: Logger

assembl/tasks/piwik.py
```diff
@@ -0,0 +1,321 @@
...
+        if user_created:
+            res["user_password"] = user_password
+        if site_id:
+            res["site_id"] = site_id
+        return res
+    except requests.ConnectionError as e:
+        raise RuntimeError("call to Piwik returned an error")
+
+
+def piwik_UsersManager_userExists(piwik_url, piwik_api_token, userLogin):
```
ayyazdaniaryan: Naming convention? Dunno, not that picky at this point...

assembl/tasks/piwik.py
```diff
@@ -0,0 +1,321 @@
...
+
+def piwik_SitesManager_getSitesIdFromSiteUrl(piwik_url, piwik_api_token, url):
+    params = {
+        "module": "API",
+        "format": "JSON",
+        "token_auth": piwik_api_token
+    }
+    params["method"] = "SitesManager.getSitesIdFromSiteUrl"
+    params["url"] = url
+    result = requests.get(piwik_url, params=params, timeout=15)
```
ayyazdaniaryan: Good job on adding timeout!

ayyazdaniaryan: Is that in seconds or milliseconds?

assembl/tasks/piwik.py
```diff
@@ -0,0 +1,321 @@
...
+    Warning: A new Piwik site is created everytime this method is called, even if another Piwik website already exists with the same name and URLs"""
+    params = {
+        "module": "API",
+        "format": "JSON",
+        "token_auth": piwik_api_token
+    }
+    params["method"] = "SitesManager.addSite"
+    params["siteName"] = siteName
+    params["urls"] = urls
+    if ( ecommerce ):
```
ayyazdaniaryan: You can default the values like `ecommerce`, and default them to `None`.

ayyazdaniaryan: Then here, instead of long if statements:

```params[key] if key```.

Comment (ayyazdaniaryan) [inline assembl/tasks/piwik.py]: I know what @maparent would do:

```
def piwik_add_site(url, token, siteName, urls, **kwargs):
   ....
   extras = ('ecommerce', 'siteSearch', 'searchKeywordParameters', ...)
   for key in extras:
       params[key] = kwargs[key] if kwargs[key]
   ....

assembl/templates/includes/javascript.jinja2
```diff
@@ -23,3 +23,9 @@
 </script>
 
 <script type="text/javascript">var start_application = false;</script>
+
+<script type="text/javascript">
+  {% if analytics_url %}
+    var analyticsUrl = "{{ analytics_url }}";
```
ayyazdaniaryan: Meaningless in v2 (I wrote this)

assembl/templates/notifications/html_mail.jinja2
```diff
@@ -5,15 +5,15 @@ The HTML conforms to ink's CSS.  http://zurb.com/ink/docs.php
 
 It expects variables:
 subscription
+discussion
```
ayyazdaniaryan: Blessings!!!

assembl/templates/notifications/html_new_discussion.jinja2
```diff
@@ -0,0 +1,29 @@
...
+admin_org_name
+admin_org_url
+notification
+frontendUrls
+ink_css
+#}
+
+{% extends "notifications/html_mail.jinja2" %}
+
+{% block notification_content %}
```
ayyazdaniaryan: Copy is satisfactory for @FrankEscoubes and Fred?

assembl/templates/notifications/txt_new_discussion.jinja2
```diff
@@ -0,0 +1,23 @@
...
+documentation_url
+admin_email
+admin_name
+admin_org_name
+admin_org_url
+notification
+frontendUrls
+#}
+
+{{gettext("""Congratulations, your Assembl consultation is ready, and you can find it here:
```
ayyazdaniaryan: Copy? @FrankEscoubes and Fred?

assembl/view_def/etalab.json
```diff
@@ -0,0 +1,15 @@
+{
```
ayyazdaniaryan: Why does etalab have it's own view_def?

assembl/views/api2/discussion.py
```diff
@@ -822,3 +827,78 @@ def test_results(request):
              permission=P_READ)
 def test_sentry(request):
     raise RuntimeError("Let's test sentry")
+
+
+@etalab_discussions.post(permission=P_SYSADMIN)
```
ayyazdaniaryan: Never knew that view_config and corince can be mixed like this..

ayyazdaniaryan: Any side-effects?

assembl/views/api2/discussion.py
```diff
@@ -822,3 +827,78 @@ def test_results(request):
...
+    permissions = get_permissions(user_id, None)
+    is_etalab_request = (request.matched_route and request.matched_route.name == 'etalab_discussions')
+    if is_etalab_request:
+        # The Etalab specification says that the API call representing the instance creation request must contain the following fields:
+        # - requestIdentifier
+        # - name: the title of the discussion (discussion.topic)
+        # - slug
+        # - adminName
+        # - adminEmail
+        default_view = 'etalab'
```
ayyazdaniaryan: Ahh...gotcha.

assembl/views/api2/discussion.py
```diff
@@ -822,3 +827,78 @@ def test_results(request):
...
+            account = EmailAccount(profile=user, email=creator_email, verified=True)
+            db.add(user)
+            db.flush()
+        json['creator'] = user.uri()
+    else:
+        user = None
+    try:
+        instances = ctx.create_object(typename, json, user_id)
+        discussion = instances[0]
+        # Hackish. Discussion API? Generic post-init method?
```
ayyazdaniaryan: No, not too hackish. But if so, can wrap the logic below in a method like the piwik website creator, etc if you think it's that hackish. I don't.

ayyazdaniaryan: You'll have to find a way to carry data to that called method, of course...

assembl/views/api2/discussion.py
```diff
@@ -822,3 +827,78 @@ def test_results(request):
...
+        for instance in instances:
+            db.add(instance)
+        db.flush()
+        view = request.GET.get('view', None) or default_view
+        uri = "/".join((API_ETALAB_DISCUSSIONS_PREFIX, str(first.id))) if is_etalab_request else None
+        return CreationResponse(first, user_id, permissions, view, uri=uri)
+
+
+def includeme(config):
+    # Make sure that the cornice view is registered
```
ayyazdaniaryan: Nothing here though? Should remove includeme since the cornice views are added by a `config.scan()` in the `assembl.views` module?

assembl/views/traversal.py
```diff
@@ -243,10 +243,13 @@ def create_query(self, id_only=True, tombstones=False):
...
 
     def get_class(self, typename=None):
+        """Returns the collection class, or subclass designated by typename"""
+        cls = self._class
         if typename is not None:
-            return get_named_class(typename)
-        else:
-            return self.collection.collection_class
+            other_cls = get_named_class(typename)
+            if other_cls and issubclass(other_cls, cls):
```
ayyazdaniaryan: Why must the collection be a subclass of class?

doc/vmm.rst
```diff
@@ -0,0 +1,146 @@
+Setting up Dovecot and VMM
```
ayyazdaniaryan: Thank you for the docs!!!

doc/vmm.rst
```diff
@@ -0,0 +1,146 @@
...
+put in ``/usr/share/doc/vmm/examples/pgsql`` by ``apt-get``. Here are further
+instructions for specific steps.
+
+Installation prerequisites
+--------------------------
+
+.. code:: sh
+
+    fab env_dev install_dovecot_vmm
+    # currently on branch openstack
```
ayyazdaniaryan: Remove the comment

doc/vmm.rst
```diff
@@ -0,0 +1,146 @@
...
+
+    cp /usr/share/doc/vmm/examples/postfix/* /etc/postfix
+
+and adjust passwords by hand in those files.
+
+In ``/etc/vmm/vmm.cfg``:
+
+.. code:: patch
+
+    135c136
```
ayyazdaniaryan: This is confusing. What is 135c136?

fabfile.py
```diff
@@ -831,9 +833,10 @@ def install_basetools():
         path_pip = run('which pip')
         assert path_pip == '/usr/local/bin/pip',\
             "Make sure homebrew is in the bash path, got " + path_pip
-        sudo('pip install virtualenv')
+        run('pip install virtualenv')
+        run('pip install psycopg2')
     else:
-        sudo('apt-get install -y python-virtualenv python-pip')
+        sudo('apt-get install -y python-virtualenv python-pip python-psycopg2')
```
ayyazdaniaryan: This line is a blessing.

fabfile.py
```diff
@@ -1494,6 +1535,35 @@ def uninstall_piwik():
...
+        sudo("debconf-set-selections <<< 'postfix postfix/relayhost string %s'" % (external_smtp_host,))
+    else:
+        sudo("debconf-set-selections <<< 'postfix postfix/main_mailer_type string \"Internet site\"'")
+    sudo("DEBIAN_FRONTEND=noninteractive apt-get -y install postfix")
+
+
+@task
+def install_dovecot_vmm():
+    """Install dovecot and vmm for IMAP. Assumes postfix is installed. Configuration TODO."""
+    assert not env.mac
```
ayyazdaniaryan: lol, sorry MacOS fanboys...wait that's me now!

assembl/auth/util.py
```diff
@@ -144,6 +146,59 @@ def forget(self, request):
             request)
 
 
+class TokenSessionAuthenticationPolicy(SessionAuthenticationPolicy):
+    """ A session authentication policy that accepts tokens for identity instead of
+    the beaker session's login."""
+
+    def user_from_token(self, request):
+        token = request.headers.get('X-Api-Key', None)
```
maparent: Sure. Making it into a class var for now.

assembl/lib/discussion_creation.py
```diff
@@ -0,0 +1,46 @@
...
+"""
+
+from zope import interface
+from pyramid.path import DottedNameResolver
+from pyramid.config import aslist
+
+resolver = DottedNameResolver(__package__)
+
+
+class IDiscussionCreationCallback(interface.Interface):
```
maparent: That's what zope.interface is about yes :)

assembl/lib/discussion_creation.py
```diff
@@ -0,0 +1,46 @@
...
+
+def setup_discussion_callbacks(registry):
+    """Register all :py:class:`IDiscussionCreationCallback` implementations
+    according to ``local.ini``"""
+    settings = registry.settings
+    class_names = aslist(settings.get('discussion_callbacks', ''))
+    for class_name in class_names:
+        cls = resolver.resolve(class_name)
+        registry.registerUtility(
+            cls(), IDiscussionCreationCallback, cls.__name__)
```
maparent: Correct.

assembl/models/discussion.py
```diff
@@ -338,6 +343,14 @@ def get_base_url(self, require_secure=None):
         """
         return get_global_base_url(require_secure)
 
+    def get_discussion_urls(self):
+        discussion_url_http = self.get_base_url(False) + "/" + self.slug
```
maparent: It's the first parameter, I can just fill it.

assembl/models/discussion.py
```diff
@@ -497,12 +522,8 @@ def __init__(self, cls):
                 super(AllUsersCollection, self).__init__(cls, User)
 
             def decorate_query(self, query, owner_alias, last_alias, parent_instance, ctx):
-                try:
-                    return query.join(owner_alias)
-                except InvalidRequestError:
-                    # No real outerjoin in sqlalchemy. Use a dummy condition.
-                    return query.outerjoin(
-                        owner_alias, owner_alias.id != None)
+                return query.outerjoin(
```
maparent: If there are no users, we should still get the discussion as part of the join.

assembl/models/notification.py
```diff
@@ -897,33 +898,58 @@ def render_to_email_text_part(self):
...
-        locale = self.first_matching_subscription.user.get_preferred_locale()
+    def get_jinja_env(self):
+        threadlocals = self.threadlocals
+        if getattr(threadlocals, 'jinja_env', None) is None:
+            threadlocals.jinja_env = self.make_unlocalized_jinja_env()
+        self.setup_localizer(
+            threadlocals.jinja_env, self.first_matching_subscription.user)
+        return threadlocals.jinja_env
+
+    @classmethod
```
maparent: So they can be called without a notification instance, as is done in email_discussion_creator. It's twisting the intent, to be honest; Maybe I should send it through the notification mechanism. But Notifications require a NotificationSubscription, which I felt was inappropriate in this case. Changing that would be major, but open to discussion.

assembl/scripts/ini_combine.py
```diff
@@ -0,0 +1,30 @@
...
+if __name__ == '__main__':
+    ap = ArgumentParser(description='Combine two ini files.')
+    ap.add_argument('source', type=FileType('r'),
+                    help='The source file')
+    ap.add_argument('overlay', type=FileType('r'),
+                    help='The overlay file')
+    ap.add_argument('result', type=FileType('w'),
+                    help='The resulting file')
+    args = ap.parse_args()
+    source = Parser()
```
maparent: Lots of second thought about this one. Still think it's a good basic idea, but I'm not sure I see myself using it quite yet.
Using the ArgumentParser means that it's _not_ able to be used to modify a target in-place, btw; but we can use '-' as argument for source, overlay and/or result, meaning stdin/stdout. (not source and overlay both, of course.) (Adding that to a comment...)
But now, I'm thinking more in terms of a templatized production.ini, it's a different direction.

assembl/scripts/ini_combine.py
```diff
@@ -0,0 +1,30 @@
...
+    source.readfp(args.source)
+    overlay = Parser()
+    overlay.readfp(args.overlay)
+    for section in overlay.sections():
+        if section != 'DEFAULT' and not source.has_section(section):
+            source.add_section(section)
+        for option in overlay.options(section):
+            value = overlay.get(section, option)
+            # Special case: {random66} => random string of 66*(4/3) characters
+            if value.startswith('{random') and value.endswith("}"):
```
maparent: Not the key, the overlay value. It allows us to randomize the salt as part of this templatization process.

assembl/static/js/app/views/infobar.js
```diff
@@ -41,6 +41,26 @@ var PiwikIframeModal = Backbone.Modal.extend({
...
   },
+
+  getStatsUrl: function(){
+    // var url = "//piwik.coeus.ca/index.php?module=CoreAdminHome&action=optOut&language=fr";
+    var url = analyticsUrl;
+    if (! (url[url.length-1] === '/')) {
+      url = url + "/";
+    }
+    url = url + "index.php?module=CoreAdminHome&action=optOut&language=";
+    var locale = Ctx.getLocale() || "fr";  // Either get the locale, or the default is French
```
maparent: Hmmm... with all the cascades in the backend, which end in server settings, I really cannot see a scenario where Ctx.getLocale() would be null. I'd be more comfortable failing all of assembl if it were (I'm sure that would happen anyway.)
Hardcoding a default in the templates is a no-no obviously.

assembl/tasks/create_vmm_source.py
```diff
@@ -0,0 +1,44 @@
...
+    """
+    interface.implements(IDiscussionCreationCallback)
+
+    def discussionCreated(self, discussion):
+        from assembl.models import IMAPMailbox
+        password = b64encode(urandom(12))
+        email = "@".join((discussion.slug, config.get("imap_domain")))
+        mailbox = IMAPMailbox(
+            name=discussion.slug + " imap",
+            host='localhost',
```
maparent: In VMM scenario, it has to be localhost.

assembl/tasks/create_vmm_source.py
```diff
@@ -0,0 +1,44 @@
...
+            username=email,
+            password=password,
+            admin_sender=email,
+            discussion=discussion,
+            folder='inbox',
```
maparent: Idem.

assembl/tasks/create_vmm_source.py
```diff
@@ -0,0 +1,44 @@
...
+            port=143,
+            use_ssl=False)
+        discussion.db.add(mailbox)
+        with TemporaryFile() as stderr:
+            rcode = call(['sudo', 'vmm', 'ua', email, password],
```
maparent: This is run every time a discussion is created.
Calling fabric from assembl would be maximally inconsistent.

assembl/tasks/email_discussion_creator.py
```diff
@@ -0,0 +1,83 @@
...
+            'documentation_url': config.get('documentation_url'),
+            'admin_email': config.get('admin_email'),
+            'admin_name': config.get('admin_name'),
+            'admin_org_name': config.get('admin_org_name'),
+            'admin_org_url': config.get('admin_org_url'),
+        }
+        html_template = jinja_env.get_template(
+            'notifications/html_new_discussion.jinja2')
+        html_body = html_template.render(**template_data)
+        html_body = Premailer(html_body, disable_leftover_css=True).transform()
```
maparent: Copy-pasted from notification code, I don't know.

assembl/tasks/email_discussion_creator.py
```diff
@@ -0,0 +1,83 @@
...
+        text_body = text_template.render(**template_data)
+        sender_email = config.get('assembl.admin_email')
+        mailer = get_mailer(request)
+        localizer = request.localizer
+        sender_email = config.get('assembl.admin_email')
+        sender_name = discussion.topic
+        sender_name = UnicodeDammit(sender_name).unicode_markup
+        # sanitize
+        sender_name = re.sub(
+            ur"[^-\w\s'\u2019\u2032\u00b4\.\(\)]", '', sender_name, 0, re.UNICODE)
```
maparent: copy-pasted from send_change_password_email. Essentially allowing word chars and a few name-relevant punctuations, including dot, hyphen and various unicode apostrophes.
This is a great candidate for refactoring in a function, I'll do that.

assembl/tasks/email_discussion_creator.py
maparent: Not verified, but no objections from the etalab team either.

assembl/tasks/piwik.py
maparent: Not bad! As a one-liner:
params.extend({k: v for (k,v) in kwargs.iteritems() if k in extras})

assembl/views/api2/discussion.py
maparent: You can always apply multiple view_config decorators. Cornice just creates dyanamic view_config decorators. No side effect.

assembl/views/api2/discussion.py
maparent: I think it does not belong in the IDiscussionCreation mechanism; it's a bit too fundamental to discussion structure for that. Other than that... maybe in discussion.__init__? I'll think more about it.

assembl/views/api2/discussion.py
maparent: It's about the timing of includes vs scan. I want to make sure that the /instances cornice view is registered before the home view is registered, so it takes precedence. There's a comment to that effect here and in views/__init__.py:includeme

assembl/views/traversal.py
maparent: I'm in /data/SomeClass, so there's a class constraint, but my request also constrains the typename further. If the typename is not a subclass of SomeClass, I should get an empty result. (contradictory class constraints)

doc/vmm.rst
maparent: done

doc/vmm.rst
maparent: this is a diff patch, as if you had done a diff, and telling you the line number. Should I make it not-diff-format?

fabfile.py
maparent: I did not try vmm on mac, but it's not in homebrew.

assembl/models/discussion.py
ayyazdaniaryan: Good to know

assembl/models/notification.py
ayyazdaniaryan: Yes, I see. The notification system was intended to be subscribed to, and then it sends out emails. Okay, I see why you did it. Hmmm, tough one. We're bending assumptions a bit. Want to discuss this with @swergas next week? I think for this PR this change is okay. But this shows a rigidity in the notification system that I wasn't expecting. I say, TBD later.

assembl/tasks/create_vmm_source.py
ayyazdaniaryan: Okay, fair enough

assembl/tasks/create_vmm_source.py
ayyazdaniaryan: Oh. I thought that it's a bash script call. Since Fabric is, as advertised, 'just python', it wouldn't be like running the call on a daemon, etc.

I see your point though. We have never used Fabric in code before. Always commands run by human.

assembl/tasks/email_discussion_creator.py
ayyazdaniaryan: Okay, good enough

assembl/tasks/piwik.py
ayyazdaniaryan: Now you're just showing off hehe

doc/INSTALL.rst
```diff
@@ -234,7 +234,10 @@ update with the following:
 Setting up a production dedicated instance
 ------------------------------------------
 
-Start as a user with sudo access (NOT as root!)
+In what follows, we will assume that the nginx web server runs in group `www-data`, and that the assembl server runs under a separate user, called `assembl_user`. In the sample files in `doc`, we use `ubuntu` instead of `assembl_user`, as this is the most convenient for cloud setup.
+Note: It is also possible to use the same user, and to put assembl in `/var/www`. It may be simpler for a single-server setup.
+
+Start as a user with sudo access
```
swergas: Please re-add "(NOT as root!)" or something similar

assembl/tasks/piwik.py
maparent: That's a good reason.

assembl/tasks/email_discussion_creator.py
```diff
@@ -0,0 +1,79 @@
...
+            'notifications/html_new_discussion.jinja2')
+        html_body = html_template.render(**template_data)
+        html_body = Premailer(html_body, disable_leftover_css=True).transform()
+        text_template = jinja_env.get_template(
+            'notifications/txt_new_discussion.jinja2')
+        text_body = text_template.render(**template_data)
+        sender_email = config.get('assembl.admin_email')
+        mailer = get_mailer(request)
+        localizer = request.localizer
+        sender_email = config.get('assembl.admin_email')
```
maparent: thanks, correcting.

doc/INSTALL.rst
maparent: Intentionally removed. That was a wrong assumption. Root is fine.

PR #58, opened 2016-11-27
Shibboleth

maparent:
Basic saml functionality using python-social-auth; expand authorization_server_backend to enable auto-login using the login route using any social backend, including shibboleth.
(Use non-contextual login or /login_forceproviders if you want to do an email login.)

### DIFF
--- a/assembl/auth/__init__.py
+++ b/assembl/auth/__init__.py
@@ -98,3 +98,7 @@ class CrudPermissions(object):
             return (self.delete, self.delete_owned)
         else:
             raise ValueError()
+
+
+def includeme(config):
+    config.include('.social_auth')
--- a/assembl/auth/social_auth.py
+++ b/assembl/auth/social_auth.py
@@ -139,6 +139,21 @@ def auto_subscribe(backend, social, user, *args, **kwargs):
 
 class AssemblStrategy(PyramidStrategy):
 
+    def request_is_secure(self):
+        return self.request.scheme == 'https'
+
+    def request_path(self):
+        return self.request.path
+
+    def request_port(self):
+        return self.request.host_port
+
+    def request_get(self):
+        return self.request.GET
+
+    def request_post(self):
+        return self.request.POST
+
     def get_preferences(self):
         discussion = discussion_from_request(self.request)
         if discussion:
--- a/assembl/auth/social_auth.py
+++ b/assembl/auth/social_auth.py
@@ -11,11 +11,14 @@ from pyramid.security import (
     forget,
     Everyone,
     authenticated_userid)
+from pyramid.config import aslist
+import simplejson as json
 
 from social.apps.pyramid_app.utils import backends
 from social.strategies.pyramid_strategy import PyramidStrategy
-from social.utils import to_setting_name
+from social.utils import to_setting_name, setting_name, SETTING_PREFIX
 from social.exceptions import AuthException
+from social.backends.utils import load_backends
 
 from assembl.models import (
     User, Preferences, AbstractAgentAccount, IdentityProvider)
@@ -233,3 +236,53 @@ class AssemblStrategy(PyramidStrategy):
             # Update the user record with any changed info from the auth service.
             'assembl.auth.social_auth.user_details'
         )
+
+
+def get_active_auth_strategies(settings):
+    """Give the list of available social auth providers.
+    Includes multiple instances if a provider can have multiple servers.
+    This currently includes SAML, and eventually wordpress.
+    TODO: Should replace the login_providers config variable"""
+    all_backends = load_backends(settings.get('SOCIAL_AUTH_AUTHENTICATION_BACKENDS'))
+    for backend_name in all_backends:
+        def get_setting(name):
+            return (settings.get(setting_name(SETTING_PREFIX, backend_name, name), None)
+                    or settings.get(setting_name(backend_name, name), None))
+        if backend_name == 'wordpress-oauth2':
+            # TODO: This special case needs to be treated the same as saml asap.
+            # Also: maybe check preferences
+            yield backend_name
+        elif backend_name == 'saml':
+            # special case: Multiple IDPs
+            idps = get_setting('ENABLED_IDPS') or {}
+            for idp in idps.keys():
+                yield 'saml:' + idp
+        elif get_setting('key'):
+            yield backend_name
+
+
+def includeme(config):
+    """Pre-parse certain settings for python_social_auth, then load it."""
+    settings = config.get_settings()
+    settings['login_providers'] = aslist(settings.get('login_providers', ''))
+    settings['trusted_login_providers'] = aslist(settings.get('trusted_login_providers', ''))
+    if not any(settings['login_providers']):
+        sys.stderr.write('no login providers configured, double check '
+                         'your ini file and add a few')
+    for name in ('SOCIAL_AUTH_AUTHENTICATION_BACKENDS',
+                 'SOCIAL_AUTH_USER_FIELDS',
+                 'SOCIAL_AUTH_PROTECTED_USER_FIELDS',
+                 'SOCIAL_AUTH_FIELDS_STORED_IN_SESSION'):
+        settings[name] = aslist(settings.get(name, ''))
+    for name in ('SOCIAL_AUTH_SAML_ORG_INFO',
+                 'SOCIAL_AUTH_SAML_TECHNICAL_CONTACT',
+                 'SOCIAL_AUTH_SAML_SUPPORT_CONTACT',
+                 'SOCIAL_AUTH_SAML_ENABLED_IDPS'):
+        val = settings.get(name, '')
+        if val:
+            settings[name] = json.loads(val)
+    for k in settings.iterkeys():
+        if k.endswith("_SCOPE") and k.startswith("SOCIAL_AUTH_"):
+            settings[k] = aslist(settings.get(k, ''))
+    config.add_request_method(
+        'assembl.auth.social_auth.get_user', 'user', reify=True)
--- a/assembl/auth/social_auth.py
+++ b/assembl/auth/social_auth.py
@@ -52,6 +52,10 @@ def user_details(
     if social:
         social.interpret_social_auth_details(details)
         social.set_extra_data(response)
+        # Hackish, for backends like SAML that have multiple IDPs.
+        idp_name = response.get("idp_name", None)
+        if idp_name and not social.provider_domain:
+            social.provider_domain = idp_name
 
 
 def associate_by_email(backend, details, provider=None, user=None, *args, **kwargs):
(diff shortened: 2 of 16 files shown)


assembl/auth/social_auth.py
```diff
@@ -218,3 +236,53 @@ def get_pipeline(self):
...
+    for name in ('SOCIAL_AUTH_SAML_ORG_INFO',
+                 'SOCIAL_AUTH_SAML_TECHNICAL_CONTACT',
+                 'SOCIAL_AUTH_SAML_SUPPORT_CONTACT',
+                 'SOCIAL_AUTH_SAML_ENABLED_IDPS'):
+        val = settings.get(name, '')
+        if val:
+            settings[name] = json.loads(val)
+    for k in settings.iterkeys():
+        if k.endswith("_SCOPE") and k.startswith("SOCIAL_AUTH_"):
+            settings[k] = aslist(settings.get(k, ''))
```
ayyazdaniaryan: I don't get it. You're setting the values in the settings object for those that have a scope with a json value. So what happens afterwards? Does `config.get_settings()` return a reference to the main settings object that lives in the master thread for Pyramid? Only way that your changes mean anything when the process is running. Else, you're setting a dict and then ignoring it... Right or wrong?

assembl/models/preferences.py
```diff
@@ -775,3 +775,15 @@ def get_preference_data_list(self):
...
+def includeme(config):
+    """Initialize some preference values"""
+    from ..auth.social_auth import get_active_auth_strategies
+    # Note: This is temporary code, will probably create a new scalar type.
+    settings = config.get_settings()
+    # TODO: Give linguistic names to social auth providers.
+    active_strategies = {
+        k: k for k in get_active_auth_strategies(settings)}
+    active_strategies[''] = _("No special authentication")
+    Preferences.preference_data['authorization_server_backend']['scalar_values'] = active_strategies
```
ayyazdaniaryan: Hmm, interesting. You're loading `Preferences` into memory from SQLA. Then you're altering it during pyramid startup. Cool. So we can change the ini and restart the process and the global class `Preferences` will have the changes read from the ini file....

assembl/views/api/auth.py
```diff
@@ -307,3 +311,16 @@ def get_user_has_permission(request):
     if not discussion:
         raise HTTPNotFound("Discussion %d does not exist" % (discussion_id,))
     return a_users_with_permission(discussion_id, permission)
+
+
+@view_config(route_name='saml_metadata')
+def saml_metadata_view(request):
+    complete_url = request.route_url('social.complete', backend="saml")
+    saml_backend = load_backend(
+        load_strategy(request),
```
ayyazdaniaryan: Nvm, now I understand what a `Backend` is to python social auth, and what the `Strategy` object does...read [here](https://github.com/omab/python-social-auth/tree/master/social/backends)

assembl/views/api/auth.py
```diff
@@ -307,3 +311,16 @@ def get_user_has_permission(request):
...
+        "saml",
+        redirect_uri=complete_url,
+    )
+    metadata, errors = saml_backend.generate_metadata_xml()
+    if not errors:
```
ayyazdaniaryan: If Error though?? Does `generate_metadata_xml` throw an exception?

assembl/views/auth/views.py
```diff
@@ -165,8 +195,12 @@ def login_view(request):
     if request.scheme == "http"\
             and asbool(config.get("accept_secure_connection")):
         raise HTTPFound("https://" + request.host + request.path_qs)
-    return get_login_context(
-        request, request.matched_route.name == 'login_forceproviders')
+    force_providers = request.matched_route.name.endswith('_forceproviders')
+    if request.matched_route.name == 'contextual_login':
+        contextual_login = get_social_autologin(request)
+        if contextual_login:
+            raise HTTPFound(contextual_login)
```
ayyazdaniaryan: If there is a `contextual_login, redirect to that request? Otherwise, no exception thrown, continue as usual?

fabfile.py
```diff
@@ -846,7 +846,7 @@ def install_builddeps():
         # may require a sudo
         if not run('brew link libevent', quiet=True):
             sudo('brew link libevent')
-        run('brew install zeromq libtool libmemcached gawk')
+        run('brew install zeromq libtool libmemcached gawk libxmlsec1')
```
ayyazdaniaryan: Quee? Used by python social auth to speed up the XML parsing of the SAML metadata? I'm confused as to how this made it in here...

production.ini
```diff
@@ -124,7 +125,6 @@ SOCIAL_AUTH_AUTHENTICATION_BACKENDS = assembl.auth.wordpress.WordPressServerOAut
     social.backends.dropbox.DropboxOAuth
     social.backends.eveonline.EVEOnlineOAuth2
     social.backends.evernote.EvernoteSandboxOAuth
-    social.backends.fitbit.FitbitOAuth
```
ayyazdaniaryan: Ahhh finally, hope it doesn't throw that stupid error anymore.

testing.ini.example
```diff
@@ -45,6 +45,7 @@ db_schema = public
 # For virtuoso:
 #db_schema = assembl_test
 db_database = assembl_test
+# db_host = localhost
```
ayyazdaniaryan: I think this is one change that came from your cherry-picking of changesets. Not intended for this PR, but will be merged once the rest of SAML branch is merged?

requirements.txt
```diff
@@ -26,7 +26,9 @@ cov-core==1.15.0
 coverage==4.1
 cryptography==1.4
 DateTime==4.1.1
+defusedxml==0.4.1
```
ayyazdaniaryan: A brief explanation as to how this made it in please?

requirements.txt
```diff
@@ -43,6 +45,7 @@ future==0.15.2
 gensim==0.12.3
 google-api-python-client==1.5.1
 greenlet==0.4.10
+hiredis==0.2.0
```
ayyazdaniaryan: And the rest of these changes :)

assembl/auth/__init__.py
```diff
@@ -98,3 +98,7 @@ def crud_permissions(self, operation):
             return (self.delete, self.delete_owned)
         else:
             raise ValueError()
+
+
+def includeme(config):
```
maparent: http://docs.pylonsproject.org/projects/pyramid/en/latest/api/config.html#pyramid.config.Configurator.include

assembl/auth/social_auth.py
```diff
@@ -218,3 +236,53 @@ def get_pipeline(self):
             # Update the user record with any changed info from the auth service.
             'assembl.auth.social_auth.user_details'
         )
+
+
+def get_active_auth_strategies(settings):
+    """Give the list of available social auth providers.
```
maparent: You say generator, I say list, let's not call the whole thing off.

assembl/auth/social_auth.py
maparent: The python-social-auth process expects those preferences to be available as lists/dicts, not as strings. So I'm setting things up for that process.

assembl/auth/social_auth.py
```diff
@@ -218,3 +236,53 @@ def get_pipeline(self):
...
+                 'SOCIAL_AUTH_SAML_SUPPORT_CONTACT',
+                 'SOCIAL_AUTH_SAML_ENABLED_IDPS'):
+        val = settings.get(name, '')
+        if val:
+            settings[name] = json.loads(val)
+    for k in settings.iterkeys():
+        if k.endswith("_SCOPE") and k.startswith("SOCIAL_AUTH_"):
+            settings[k] = aslist(settings.get(k, ''))
+    config.add_request_method(
+        'assembl.auth.social_auth.get_user', 'user', reify=True)
```
maparent: Ouch, that was just moved. The reify was from February. It seems that reify caches the result.
http://pyramid.readthedocs.io/en/latest/api/request.html#pyramid.request.Request.set_property

assembl/models/preferences.py
maparent: Not exactly. This construction from local.ini is the basis. What is read from sqla is overlaid on that. In most cases, that particular value (the preference_data) will not be in the sqla (the preferences are sparse) hence what was built from local.ini will be preserved. But in theory, if we make a change to preferences, we should make a migration to preference_data that could be in the Preference jsons in sqla. I have not done it because it's never been stored yet afaik... a bit fragile, but maximally flexible.

assembl/views/api/auth.py
maparent: Good point. Added an exception.

fabfile.py
maparent: requirement of saml backend, yes. It's not just speeding up, the XML is cryptographically signed.

production.ini
maparent: no kidding.

testing.ini.example
maparent: could be. I say let's leave it in.

requirements.txt
maparent: part of those same saml requirements.

requirements.txt
maparent: Ah no real reason why this is here, but I found it while working on other stuff. Free gift :) Optimizes redis access.

assembl/auth/__init__.py
maparent: http://pyramid.readthedocs.io/en/latest/api/config.html#pyramid.config.Configurator.include

assembl/auth/social_auth.py
maparent: That was code from february that was just moved, you got me digging!
Basically, reify will cache the value.
http://pyramid.readthedocs.io/en/latest/api/request.html#pyramid.request.Request.set_property

assembl/auth/social_auth.py
maparent: pyramid-social-auth expects lists/json in those preferences, not strings. I'm preparing the field for it.

assembl/models/preferences.py
maparent: Not exactly. These base values are constructed with ini, then overlaid with whatever is in the Preference sqla objects. But the overlay is sparse, and in known cases the preference_data is absent from the preference objects in assembl. Otherwise, we should actually migrate the preference_data json in Preferences objects in the DB whenever we alter the preferences. I have not done it because I know preference_data is never defined, but that's actually wrong. So this is fragile, but OTH it does allow to change the preferences structure through data (which was given to me as a requirement.)

maparent:
sorry, I seemed to lose comments, then rewrote them, then they were there. Github is being weird.

ayyazdaniaryan:
I don't see other problems here from reviewing. Looks good.

PR #59, opened 2016-11-30
Add a different route set for the administration views

maparent:


assembl/templates/views/navBarRight.tmpl
```diff
@@ -20,6 +21,7 @@
     </li>
     {% endif %}
     {% endif %}
+    {% endif %}
```
ayyazdaniaryan: In this line:
```{% if request.current_route_path().split('/')[-1] not in ('login', 'register')  %}```
Maybe we should aim to put these kind of logic into methods? Hard to debug, read, etc. Eh, nitpicky..

File: assembl/views/admin/views.py [hunk 2]
```diff
@@ -18,12 +19,79 @@
...
+def base_admin_view(request):
+    """The Base admin view, for frontend urls"""
+    user_id = authenticated_userid(request) or Everyone
+    context = get_default_context(request)
+
+    session = Discussion.default_db
+    preferences = Preferences.get_default_preferences(session)
+    user = User.get(user_id)
+
+    if '_LOCALE_' in request.cookies:
```
ayyazdaniaryan: Very unDRY. I'll make this a method, as it's being used a few view_config's now..

assembl/views/admin/views.py
```diff
@@ -18,12 +19,79 @@
...
+                       LanguagePreferenceOrder.Parameter)
+    else:
+        locale = locale_negotiator(request)
+        process_locale(locale, user, session,
+                       LanguagePreferenceOrder.OS_Default)
+
+    target_locale = Locale.get_or_create(
+        strip_country(locale), session)
+    locale_labels = json.dumps(
+        DummyGoogleTranslationService.target_locale_labels_cls(target_locale))
```
ayyazdaniaryan: Admin view always get's the `DummyGoogle...Service`? Or should we parameterize with preferences for proper translation service?

assembl/views/admin/views.py
```diff
@@ -18,12 +19,79 @@
...
+    context['translation_locale_names_json'] = locale_labels
+
+    role_names = [x for (x,) in session.query(Role.name).all()]
+    permission_names = [x for (x,) in session.query(Permission.name).all()]
+    context['role_names'] = json.dumps(role_names)
+    context['permission_names'] = json.dumps(permission_names)
+    context['discussion'] = PseudoDiscussion()
```
ayyazdaniaryan: Ah okay...I see why `PseudoDiscussion` is used. Thought so. So this discussion is instantiated everytime the view is called. Why not make everything on it a classmethod or classvariable, so it's a singleton object? Not like this makes any difference on performance at this point...Ahhh again, being nitpicky

assembl/views/admin/views.py
```diff
@@ -18,12 +19,79 @@
...
+
+    response = render_to_response('../../templates/adminIndex.jinja2', context,
+                                  request=request)
+    # Prevent caching the home, especially for proper login/logout
+    response.cache_control.max_age = 0
```
ayyazdaniaryan: Again, unDRY. Abstract an `uncached_rendered_response` method...

ayyazdaniaryan: Quick code though...so I understand

assembl/lib/frontend_urls.py
```diff
@@ -46,7 +45,18 @@ def __init__(self, discussion):
...
-    The list of frontend routes.
+    The list of frontend discussion routes.
+
+    Important:  This should match with :js:class:`Router`
+    Used by :py:func:`assembl.views.backbone.views.home_view`, these routes
+    will all give the same view and further routing will happen
+    in the frontend."""
+
+    frontend_admin_routes = {
+        'admin_global_preferences': '/global_preferences',
```
maparent: Correct.

assembl/templates/adminIndex.jinja2
```diff
@@ -0,0 +1,66 @@
+{% extends 'base.jinja2' %}
+
+{% block page_title %}{% trans %}Administration{% endtrans %}{% endblock %}
+
+{% block meta_tags %}
+{% endblock meta_tags %}
+
+{% block content %}
+<input type="hidden" name="discussion-id" id="discussion-id" value="" />
```
maparent: Following pattern, the same JS code will look for that data in the same place.

assembl/templates/adminIndex.jinja2
```diff
@@ -0,0 +1,66 @@
+{% extends 'base.jinja2' %}
+
+{% block page_title %}{% trans %}Administration{% endtrans %}{% endblock %}
```
maparent: Arf... trans is the proper jinja2 mechanism, gettext forces us to send gettext as a jinja param for no good reason. That pattern won somehow, and I don't see myself refactoring it.

assembl/views/admin/views.py
maparent: No translation service needed in that context, except to get the list of locales.

assembl/views/admin/views.py
maparent: basically because it breaks symmetry with the discussion, which is an instance and not a class. I understand why you draw the equivalence between class and singleton instance, but I don't see it that way.

assembl/templates/adminIndex.jinja2
```diff
@@ -0,0 +1,66 @@
...
+{% block main_js %}
+    <script>
+        var CKEDITOR_BASEPATH = '/static/js/node_modules/ckeditor/';
+    </script>
+    <script src="{{ STATIC_URL }}/js/build/infrastructure.min.js"></script>
+    <script src="{{ STATIC_URL }}/js/build/app.js"></script>
+{% endblock %}
+
+{% block facebook_sdk_load %}
+{% endblock%}
```
maparent: Good point. Done.

assembl/templates/views/navBarRight.tmpl
maparent: Good point, and certainly worth a comment somewhere, but has nothing to do with my PR, as it's not new code. Fixing everything within 10 lines of any change fulfills the definition of scope creep. (Same with  code that was moved in a PR, btw.) (Sorry, vaguely a rant, but you're right we need a process for such comments. PR may not be it.)

assembl/views/admin/views.py
maparent: Excellent point. Done.

assembl/templates/views/navBarRight.tmpl
ayyazdaniaryan: Yeah, agreed, not in the scope of this PR. Just saw the code, actually. It's not that big of a deal, honestly. I just know that new devs will have a really fun time hunting template bugs if one emerges. This is not priority. And indeed is scope creep.

assembl/views/admin/views.py
ayyazdaniaryan: Okay, I appreciate the answer. And I accept it :)

assembl/views/admin/views.py
ayyazdaniaryan: Ooh, didn't know there is much of a difference. Okay, I'll note this somewhere. When working on new APIs, I'll be aware of this factor.

assembl/templates/adminIndex.jinja2
```diff
@@ -1,25 +1,10 @@
-{% extends 'base.jinja2' %}
+{% extends 'index.jinja2' %}
```
ayyazdaniaryan: Noice

assembl/templates/discussion_list.jinja2
```diff
@@ -38,6 +38,9 @@
     {% if permissions_admin_url %}
       <h4 class="h4"><a href='{{ permissions_admin_url }}'>{{ gettext("Change global permissions") }}</a></h4>
     {% endif %}
+    {% if preferences_admin_url %}
+      <h4 class="h4"><a href='{{ preferences_admin_url }}'>{{ gettext("Change global preferences") }}</a></h4>
```
ayyazdaniaryan: Oh yeahh! Good call.

assembl/views/__init__.py
```diff
@@ -230,6 +240,66 @@ def process_export_list(ls):
...
+        }
+        if (locale.id, source_of_evidence) in lang_pref_signatures:
+            return
+    lang = UserLanguagePreference(
+        user=user, source_of_evidence=source_of_evidence.value, locale=locale)
+    session.add(lang)
+    session.flush()
+
+
+def get_locale_from_request(request, session=None, user=None):
```
ayyazdaniaryan: Yeah, looks good.

assembl/views/__init__.py
```diff
@@ -230,6 +240,66 @@ def process_export_list(ls):
         )
 
 
+def process_locale(
```
ayyazdaniaryan: Moved this from the old `backbone.views` module, did we?

assembl/views/admin/views.py
```diff
@@ -58,22 +52,7 @@ def base_admin_view(request):
...
-        process_locale(locale, user, session,
-                       LanguagePreferenceOrder.Parameter)
-    else:
-        locale = locale_negotiator(request)
-        process_locale(locale, user, session,
-                       LanguagePreferenceOrder.OS_Default)
-
-    target_locale = Locale.get_or_create(
-        strip_country(locale), session)
+    target_locale = get_locale_from_request(request, session, user)
```
ayyazdaniaryan: LGTM

assembl/static/js/app/utils/tourManager.js
```diff
@@ -27,6 +27,9 @@ var TourManager = Marionette.Object.extend({
       },
 
   initialize: function() {
+    if (Ctx.isAdminApp()) {
+      return;
+    }
```
swergas: Would it be possible to not load the tours on admin app, instead of instanciating the TourManager and hardcoding a break inside it? (creates a coupling problem?)

maparent: Good idea. Pull request merged, but worth doing separately. (Might do it on develop.)

PR #61, opened 2016-12-04
New frontend structure

ayyazdaniaryan:
Suggested new structure for developing the new front-end


assembl/__init__.py
```diff
@@ -91,8 +92,9 @@ def main(global_config, **settings):
         from .lib.migration import bootstrap_db_data
         bootstrap_db_data(session, False)
 
+    config.add_static_view('react', 'react', cache_max_age=3600)
```
swergas: I'm really not sure the word "react" should appear in the URL...

assembl/lib/frontend_urls.py
```diff
@@ -65,7 +65,7 @@ def __init__(self, discussion):
 
     @classmethod
     def register_frontend_routes(cls, config):
-        from assembl.views.backbone.views import home_view
+        from assembl.views.discussion.views import home_view
```
swergas: Why make the word "react" appear clearly in the architecture and not the word "backbone"? We should probably choose none or both?

assembl/views/__init__.py
```diff
@@ -38,7 +38,8 @@
 
 
 default_context = {
-    'STATIC_URL': '/static'
+    'STATIC_URL': '/static',
+    'REACT_URL': '/react'
```
swergas: Same remark: I'm really not sure the word "react" should appear in the URL...

assembl/__init__.py
paolina92: Suggestions to replace "react":
- static_v2
- frontend
- assembl (but the occurence will be present 3 times)

assembl/lib/frontend_urls.py
paolina92: In my opinion "discussion" is ok

assembl/views/__init__.py
paolina92: see my suggestions in the my other comment

assembl/lib/frontend_urls.py
maparent: I think so too. this module contains both backbone and react views, hence the rename. (plus, backbone vs home was confusing)

assembl/__init__.py
ayyazdaniaryan: @swergas @paolina92 `static2` sound okay?

ayyazdaniaryan: It was just a suggestion to use `react`. Open for a change.

paolina92: yes static2 okay!

ayyazdaniaryan:
I changed the folder name to static2, the main route to access react is still `server:port/v2/discussion_slug`.

PR #62, opened 2016-12-07
Push the logic to ensure linked list integrity upon delete/reorder to backend

maparent:
...and database
Made reordering into a RPC ajax call. Not RESTish, but this guarantees integrity.


swergas:
My code review: OK
But not tested

PR #64, opened 2016-12-13
New security

maparent:
Have LangStrings ask their owning object for CrudPermissions. Create a composite query to determine the owning object, and use events to prepopulate it.

### DIFF
--- a/assembl/models/langstrings.py
+++ b/assembl/models/langstrings.py
@@ -4,7 +4,8 @@ from datetime import datetime
 
 from sqlalchemy import (
     Column, ForeignKey, Integer, Boolean, String, SmallInteger,
-    UnicodeText, UniqueConstraint, event, inspect, Sequence, events)
+    UnicodeText, UniqueConstraint, event, inspect, Sequence, events,
+    literal)
 from sqlalchemy.sql.expression import case
 from sqlalchemy.ext.declarative import declared_attr
 from sqlalchemy.orm import (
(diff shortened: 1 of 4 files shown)


assembl/models/langstrings.py
```diff
@@ -464,6 +464,49 @@ def EMPTY(cls, db=None):
...
+                    getattr(target, reln).owner_object = target
+        event.listen(owner_class, "load", load_owner_object, propagate=True)
+        event.listens_for(owner_class, "refresh", load_owner_object, propagate=True)
+        def set_owner_object(target, value, old_value, initiator):
+            if old_value is not None:
+                old_value.owner_object = None
+            value.owner_object = target
+        for reln in relns:
+            cls._owning_relns.append((owner_class, reln))
+            event.listen(getattr(owner_class, reln), "set", set_owner_object, propagate=True)
```
ayyazdaniaryan: So even on a relationship object being `set` on the object, the class recalculates it's ownership? Or does the relationship of the object also get a crud ownership?

assembl/models/langstrings.py
```diff
@@ -464,6 +464,49 @@ def EMPTY(cls, db=None):
     def reset_cache(cls):
         pass
 
+    # Which object owns this?
+    owner_object = None
+    _owning_relns = []
```
ayyazdaniaryan: This becomes a `self._owning_relns` here in Python on the object yes? Not a class variable, if I understand correctly?

assembl/models/langstrings.py
```diff
@@ -464,6 +464,49 @@ def EMPTY(cls, db=None):
...
+
+    def get_owner_object(self):
+        if self.owner_object is None and inspect(self).persistent:
+            self.owner_object = self.owner_object_from_query()
+        return self.owner_object
+
+    def owner_object_from_query(self):
+        queries = []
+        for owning_class, reln_name in self._owning_relns:
+            backref_name = owning_class.__mapper__.relationships[reln_name].backref[0]
```
ayyazdaniaryan: the `[0]`, not hackish? Always assume there is a backref? It's from the `lazy='dynamic'`, if I understood correctly.

assembl/models/langstrings.py
maparent: Yes, we recalculate ownership when the langstring is set on the name (eg.) Not sure I understand your second question, but the relationship itself has no crud.

assembl/models/langstrings.py
maparent: The sqla backref structure is an unnamed tuple, not much I can do about the [0].
But yes I assume there's a backref.
That would fail early at worst.

assembl/models/langstrings.py
ayyazdaniaryan: Ah okay, cool, thanks.

assembl/models/langstrings.py
ayyazdaniaryan: AH! That's what this is. Okay, carry on, understood on my part

assembl/models/langstrings.py
ayyazdaniaryan: Good enough, as long as it's not a silent fail

PR #65, opened 2016-12-15
Discrepancy between smtplib.SMTP and pyramid_mailer configurations.

maparent:
Converge on pyramid_mailer.

### DIFF
--- a/assembl/models/notification.py
+++ b/assembl/models/notification.py
@@ -30,6 +30,7 @@ from  sqlalchemy.orm.exc import DetachedInstanceError
 from zope import interface
 from pyramid.httpexceptions import HTTPUnauthorized, HTTPBadRequest
 from pyramid.i18n import TranslationStringFactory, make_localizer
+from pyramid_mailer.message import Message
 from celery import current_task
 from jinja2 import Environment, PackageLoader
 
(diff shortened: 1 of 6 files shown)


assembl/models/notification.py
```diff
@@ -967,40 +968,43 @@ def get_to_email_address(self):
...
-        return msg.as_string()
+        headers['List-Subscribe'] = frontendUrls.getUserNotificationSubscriptionsConfigurationUrl()
+        headers['List-Unsubscribe'] = frontendUrls.getUserNotificationSubscriptionsConfigurationUrl()
+
+        subject = Header(self.get_notification_subject(), 'utf-8')
+
+        sender = Header(self.event_source_object().creator.name, 'utf-8')
+        sender.append(" <" + self.get_from_email_address() + ">", 'ascii')
+        recipient = self.get_to_email_address()
+        message = Message(
```
ayyazdaniaryan: Ah okay, you changed the structure from `msg` object to `headers` to comply with pyramid mailer?

assembl/models/notification.py
```diff
@@ -967,40 +968,43 @@ def get_to_email_address(self):
...
+            subject=subject.encode(),
+            sender=sender.encode(),
```
ayyazdaniaryan: the encode method encodes what format to what? Would there be any issues with the unicode strings of multi-language texts?

assembl/views/api2/notification.py
```diff
@@ -85,7 +85,7 @@ def mail_text_preview(request):
              ctx_instance_class=Notification, permission=P_READ,
              accept="text/html", name="mail")
 def mail(request):
-    return Response(request.context._instance.render_to_email(),
+    return Response(request.context._instance.render_to_message().to_message().as_string(),
```
ayyazdaniaryan: `as_string` returning plain text or html?

assembl/models/notification.py
maparent: precisely for encoding. to quoted-printable.

PR #68, opened 2016-12-21
Localization system implementation

paolina92:
- Creation of a Context Class and add getBrowserLanguage method inside
- Creation of a Langstring Object to manage localization string
- Creation of the root reducer
- Synchronisation of of translations with store


paolina92:
I added the frontend router on the same pull request / branch.

assembl/static2/tests/unit/containers/authentication.spec.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import ReactTestRenderer from 'react-test-renderer';
+import Authentication from '../../../js/app/containers/authentication';
+
+describe('This test concern Authentication component', () => {
+  it('Should test React renderer', () => {
+    const renderer = ReactTestRenderer.create(
+      <Authentication />
+      );
+    const result = renderer.toJSON();
```
ayyazdaniaryan: Hmm, ooh, so the `ReactTestRenderer` renders the entire virtual DOM and flushes the state onto it? So it's not actually binding the elements in the react DOM to the real DOM? So, what runs these tests? A node.js test runner? If so, then these are unit tests only, testing just the logic, and not the browser support and compatibility, right?

assembl/static2/tests/unit/utils/context.spec.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import Context from '../../../js/app/utils/context';
+
+describe('This test concern Context Class', () => {
```
ayyazdaniaryan: Me gusta

assembl/static2/js/app/index.jsx
```diff
@@ -1,11 +1,27 @@
 import React from 'react';
```
ayyazdaniaryan: Is there a reason why all the js files are now renamed to jsx?

assembl/static2/js/app/containers/authentication.jsx
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
+import { Translate } from 'react-redux-i18n';
+
+const Authentication = () => (
+  <p><Translate value="authentication.panelTitle" /></p>
```
paolina92: Yes that's a failure of this system, that's why I suggest to use namespaces. I used this system in my last projects and it's not so bad. However, this system is simplest than Assembl v2 ;)

assembl/static2/js/app/routes.jsx
```diff
@@ -0,0 +1,17 @@
...
+import Authentication from './containers/authentication';
+import Home from './containers/home';
+import Ideas from './containers/ideas';
+import Synthesis from './containers/synthesis';
+import Debate from './containers/debate';
+
+export default (
+  <Route>
+    <Route path="/:slug/authentication" component={Authentication} />
+    <Route path="/:slug/home" component={Home} />
```
paolina92: As you wish but is it not most natural when the last parameter changes?

assembl/static2/js/app/utils/langString.js
```diff
@@ -0,0 +1,64 @@
+const LangString = {
```
paolina92: I change for "Translations". Is it okay?

assembl/static2/tests/unit/containers/authentication.spec.js
paolina92: exact

assembl/static2/js/app/index.jsx
paolina92: Yes, it's better to use jsx extension when we have to use a react component. If not, we have to use js extension. Right?

assembl/static2/js/app/utils/context.js
```diff
@@ -0,0 +1,11 @@
+import LangString from './langString';
+
+class Context {
```
paolina92: Hmmm. Good question. I'm not sure that "context" is the good name. I need to have some global functions in the project. So I renamed it to "globalFunctions.js"? Is it okay? I need your feedback please.
However, the locale is already in the store with all translations.
The store.getState() method return this:
i18n{
    locale:"fr",
    translations:{
        de:{...},
        en:{...},
        fr:{...}
    }
}

assembl/static2/tests/unit/containers/authentication.spec.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import ReactTestRenderer from 'react-test-renderer';
+import Authentication from '../../../js/app/containers/authentication';
```
paolina92: I tried to use "~" but I broke my teeth... I need help to configure it

PR #69, opened 2017-01-02
Statistics ui

maparent:
First attempt at creating a UI for statistics. Needs UI work. Limited intervals. May fail on IE.


swergas:
This looks great :+1:

PR #71, opened 2017-01-14
Sentiment mapping

paolina92:


assembl/static/js/app/views/message.js
```diff
@@ -1048,7 +1010,122 @@ var MessageView = Marionette.LayoutView.extend({
...
+      var currentClass = $(event.target).attr("class");
+      if(currentClass.indexOf('active') <= -1){
+        var that = this;
+        var currentUser = Ctx.getCurrentUser();
+        var clickedSentiment = $(event.target).attr("data-id");
+        var currentPost = this.model.get("@id");
+        var data = {"target": currentPost, "user": currentUser.id, "@type": clickedSentiment};
+        var payload = JSON.stringify(data);
+        $.ajax(
+          "/data/Discussion/" + Ctx.getDiscussionId() + "/posts/" + this.model.getNumericId() + "/sentiments", {
```
swergas: Also, just like the like feature, clicking on an emotion icon should trigger a call to Piwik (trackEvent). And what does the like feature become? Is it completely replaced? Then we should migrate all likes which have been saved in the database to positive emotions, and remove all the code and model structures related to the like feature. Data and computations about likes in the get_participant_time_series_analytics API2 route should be replaced by their equivalents in emotions.

assembl/views/api/post.py
```diff
@@ -233,29 +228,20 @@ def get_posts(request):
         orphans = Idea._get_orphan_posts_statement(
             discussion_id, True, include_deleted=deleted).subquery("orphans")
         posts = posts.join(orphans, PostClass.id == orphans.c.post_id)
-        ideaContentLinkQuery = ideaContentLinkQuery.join(
-            orphans, PostClass.id == orphans.c.post_id)
```
swergas: I'm not sure I understand why all this code in this file is deleted?

assembl/views/api/post.py
```diff
@@ -396,15 +367,27 @@ def get_posts(request):
...
+        # and the benefit is not that high.
+        sentiment_counts = discussion.db.query(
+                PostClass.id, SentimentOfPost.type, count(SentimentOfPost.id)
+            ).join(SentimentOfPost
+            ).filter(PostClass.id.in_(posts.with_entities(PostClass.id).subquery()),
+                     SentimentOfPost.tombstone_condition()
+            ).group_by(PostClass.id, SentimentOfPost.type)
+        sentiment_counts_by_post_id = defaultdict(dict)
+        for (post_id, sentiment_type, sentiment_count) in sentiment_counts:
+            sentiment_counts_by_post_id[post_id][sentiment_type[10:]] = sentiment_count
```
swergas: Again, why 10?

assembl/static/js/app/views/message.js
maparent: I could not use a save method, because that might cause PUT/PATCH instead of a POST. I have created a side-effect on POST that tombstones the previous sentiment, not so practical to do this on PUT/PATCH. This is an exception to REST.

assembl/views/api/post.py
maparent: Simplification while I was here: Instead of keeping two queries synchronized, which is error prone, I only create one and repurpose it using
ideaContentLinkQuery = posts.with_entities(...

assembl/views/api/post.py
maparent: added a constant.

assembl/models/generic.py
```diff
@@ -509,6 +511,26 @@ def special_quad_patterns(cls, alias_maker, discussion_id):
...
 
+    @property
+    def sentiment_counts(self):
+        from .action import SentimentOfPost
+        r = self.db.query(
+                SentimentOfPost.type, count(SentimentOfPost.id)
+            ).filter(SentimentOfPost.post_id == self.id,
+                     SentimentOfPost.tombstone_condition()
+            ).group_by(SentimentOfPost.type)
+        return {k[10:]: v for (k, v) in r}
```
maparent: added a constant

assembl/static/js/app/views/message.js
maparent: Like feature: Ask Frank what's happening with the like feature. I suspect some discussions will use one feature, some the other? I did not kill it yet, that's another project.
Piwik: Great idea, you should do it. But you're right we should think about whether we want to keep both like features side-by-side or unify their meaning.

maparent: On migration: it would be easy to do so. But I'm not sure that a like means the same when it's the only option as when it's one of three options. Worth thinking about.
There's a popularity sort for the two-column interface, we need to think if it will use the new lIke or the old one. There was no thought given to feature interaction.

maparent: get_participant_time_series_analytics: Unless we decide to kill the old like feature, I think we need both. OR maybe one per distinct sentiment, but I think that's not the point of analytics. (validating that with you)

swergas: OK I asked Frank, his answer is:
* yes, sentiments completely replace the current like feature, which should disappear from the code. 
* yes, analytics must contain data for each type of sentiments
* Deadline for first production delivery is today or tomorrow, so these improvements can come in a subsequent pull request

maparent: OK. And do we refactor the old likes so they behave as new likes?
Note on the stats: that will be 8 new sets of columns in the stats (sentiment given+received.) It will make the tables unwieldy. But I understand it may be desired data.

PR #72, opened 2017-01-17
Add docker

dnephin:
Hello, it's been a while. I've taken another approach to implementing #3. This replaces #4.

With [dobi](http://dnephin.github.io/dobi/install.html) and [docker](https://docs.docker.com/engine/installation/) installed, you can setup a new development environment from a fresh checkout using:

```
dobi init devenv
# create an admin user
docker exec -ti assembl_assembl_1 assembl-add-user -m admin@example.com -n "Admin User" --username admin --password password development.ini
```

`dobi build-wheels` will build wheels for all the python dependencies so that reinstalling them is much faster, and so that dev dependencies aren't required in the runtime container.

`dobi list` shows a list of all the described tasks.


swergas:
Hello, thanks a lot for continuing the work you began 2 years ago! We will read your PR carefully :) I'm sorry in advance if this takes some time, as our team still has to level up about docker internally. We'll dive into docker's rabbit hole and come back to you!

development.ini
```diff
@@ -68,7 +68,7 @@ db_schema = public
 # For virtuoso:
 # db_schema = assembl
 db_database = assembl
-# db_host = localhost
+db_host = database
```
swergas: It looks that this change would break the current installation procedure of Assembl. It would worry me if Assembl did not allow a docker-less installation anymore, and forced the use of docker (and other tools of its ecosystem) instead. Could we find a way to support both? For example, at some point @maparent was working on the development of a .ini overriding/cascading mechanism. Maybe this could help? (a specific docker ini file would override development.ini)

dobi.yaml
```diff
@@ -0,0 +1,93 @@
+
+meta:
```
swergas: There seems to be an overlap between the use/functionnality of fabric, dobi and docker-compose. Would it be a good idea to replace the use of dobi by a fabric task which would use docker-compose?

dobi.yaml
```diff
@@ -0,0 +1,93 @@
...
+#
+# Build local wheels so that dev dependencies don't need to be installed in the 
+# runtime image.
+#
+image=wheeler:
+    image: assembl-wheeler
+    context: docker/
+    dockerfile: Dockerfile.wheel
+
+job=build-wheels:
```
swergas: I do not have yet enough knowledge about wheels to understand the whole wheels part of the PR (and what it improves compared to the current status).

development.ini
dnephin: Yes, I think that change is necessary

dobi.yaml
dnephin: [Python wheels](http://pythonwheels.com/) has all the relevant info summarized pretty well.

The advantage in this PR is that you don't need to install dev dependencies in the webapp container. 

This makes it much faster to rebuild for two reasons. One is fewer ubuntu packages to install, and the second is installing from a wheel is significantly faster than installing from source.

dobi.yaml
dnephin: I don't think there is a lot of overlap, each has a different focus:

* `fabric` focuses on deployment and host installs
* `docker-compose` focuses on isolated development environments
* `dobi` focuses on portable, and automated build operations

There is some overlap in fabric and dobi right now, but fabric requires that you're running an SSH server and you want to install all the dependencies on your host. This install can fail pretty easily if you're running an unexpected version of ubuntu, or will definitely fail for any non-ubuntu distro (I run arch).

With the `dobi` + `docker` setup you can be fairly confident that the install is going to work on any platform (linux, osx, or windows). It also simplifies the install instructions. Instead of pages of readme, you can summarize the entire process in two lines (as shown in the OP). This could even be reduced to a single line if we create a task for "create admin user".

The "deployment" and "admin" tasks from `fabric` can still be used from the prompt created by `dobi shell`

maparent: Automated build is a normal requirement for docker, but not so normal for development, where we often need to run a single phase, and not from scratch. So dobi would not replace fabric. OTH, fabric allows us to combine atomic tasks into a global task, so it can fulfill that specific dobi feature. What you say dobi gives us is better portability, and that is clearly valuable. However, is it portable because it always installs docker? We want to be able to run assembl "naked" on posix platforms, if possible.
I am curious about the steps of our fabfile install that fail on your arch setup; I never tried arch. If dobi can abstract those steps without docker, we're clearly interested.
Otherwise, as I told you earlier, I'm not against having dobi in the docker build chain; but I would personally want this chain to be optional.

development.ini
maparent: @swergas we'll have to see if we can devote time to this internally, I would be in favour.

dobi.yaml
dnephin: The fabfile uses `apt-get install` which is the package manager for debian/ubuntu, but will never work for any redhat variant, arch, gentoo, etc.

Using docker you can run a `ubuntu` container on any platform that supports docker (any linux, osx, or windows).

maparent: Ah, redhat! Thank you.
And using docker on windows is certainly the right answer for windows; it's certainly too far from posix. But on gentoo etc., maybe we should do this, but maybe again we should detect and use yum instead. Ideally both.
But more to the point: what I hear is confirmation that dobi does not abstract installation of packages on the base platform, but helps to install packages within the container. If working on a redhat-based container is a goal, and it is a worthy goal, then I would contend that working naked on a redhat-based distro is also a goal. Doing this at the fabfile level becomes independently useful. (If we decide we can stick to an ubuntu container, then dobi abstracting package installation within the container is irrelevant, no? But I don't want to think like that.)
What may be true is that installing some packages at the dobi level before the fabfile runs might be more efficient, and that this part of the fabfile should end up being a noop. But that's a different matter.

dnephin: > But more to the point: what I hear is confirmation that dobi does not abstract installation of packages on the base platform, but helps to install packages within the container.

Yes, `dobi` is a tool for working with docker containers.

> If working on a redhat-based container is a goal,

It's not. The container is always `ubuntu` (or `alpine`). But it can run on any host (not just ubuntu).

> Doing this at the fabfile level becomes independently useful.

Sure, you could attempt to support and maintain the install on every host... or you could do it just once in a container! Why do it more than once?

> If we decide we can stick to an ubuntu container, then dobi abstracting package installation within the container is irrelevant, no? 

No, because you can run that ubuntu container on any host (windows, osx, ubuntu, arch, redhat, etc).

It means that *everyone* gets exactly the same thing, instead of having to maintain multiple copies of things, and then trying to figure out why things happen to not work for platform x when they work on y.

maparent: OK, I must be missing something. Suppose I'm running on arch, the fabfile fails, I decide to go docker as you suggest. I want to build the docker image myself (instead of using an official image.) That image is ubuntu-based, so the fabfile using apt-get works within the image. dobi does not contribute to installing anything to that scenario, am I not correct?

maparent: I think now that I got on a wrong track. I thought you said that dobi solved an installation abstraction issue because you mentioned overlap; what it does is simplify a docker install. Of course that gains us platform independence at some level. But the overlap you mentioned between dobi and fabric does not mean that one should replace the other, because the overlap is in very different layers. So we keep fabric for naked (dockerless) or within-container installs, and we use dobi for the docker install option (which is the only option on some platforms.) I'm happy with this scenario now that I am clear on it.

dnephin: I see what you're saying. I think the difference is that I would never run the fabfile in the first place.

If I'm new to a project and I just want to try it out,  I don't want to install a bunch of things on my host that I may never use again, or that I might not use for a while.  I don't want to have to install updates for them, and worry about them running services. And I don't want to type my root password for it, without examining the file first (which is huge! so would take a long time). And that's assuming I'm running `ubuntu` and the `fabfile` "might" work.

So instead I want to install things in Docker. That way I can just `docker rmi` when I'm done. Even if I don't remove them, I can easily contribute things to the project without worrying about breaking it for OSX, because everyone would be running the same image. If I install on my host, I have to guess at how to make it work for OSX, redhat, etc.

>  I want to build the docker image myself 

Right, I think I would always want to build it myself, because I want to be able to test out my changes.

So I could `docker build`, and the `Dockerfile` could just run `fab ...` except that is very slow. Any change to the `fabfile` would invalidate the entire cache, which would incur another 20-30min rebuild of the image.

Using `dobi` (and having separate steps for things outside of the `fabfile`) means that most of the time the `docker build` cache will be used, and I'll be able to  pickup changes in just a few seconds, instead of 20-30min.

(`fabric` also requires an ssh server. Running an SSH server in a container is kind of an anti-pattern, but it's still possible, and there are hacks around it as well, which I think I had in my earlier PR).

dnephin: I guess another advantage is that the postgresql database, redis, memcache, and elasticsearch also run in containers, and not on the host.

This means that I can just run `docker-compose down` to stop everything when I'm done. On the host I'll have to remember to shut down the services.

maparent: Agreed on the compose point.
Interesting point about the docker cache, and yes that makes perfect sense, thank you!
SSH... yes, that's an issue that has started to annoy me as well. I might actually solve it separately, it's no big deal. If you don't want to wait for me, I suggest a separate PR.

paolina92:

PR #73, opened 2017-01-19
look for mergeable subscriptions in User.get_notification_subscriptions

maparent:
and proper merging of notification subscriptions (there and in UserTemplate)


assembl/lib/sqla.py
```diff
@@ -848,6 +853,60 @@ def translate_to_json(v):
 
     dummy_context = DummyContext()
 
+    def locked_object_creation(
```
ayyazdaniaryan: This method works great for the use-case that it was made for. However, perhaps a comment or note should be placed on its limitation, @maparent . That way it won't be used a general purpose locking mechanism under all conditions by unknowing devs. Agree?

assembl/models/auth.py
```diff
@@ -1505,61 +1550,46 @@ def get_notification_subscriptions_and_changed(self):
...
-        raise RuntimeError("Could not create the template's subscriptions")
+            return [
+                cls(
+                    discussion_id=discussion_id,
+                    user_id=my_id,
+                    creation_origin=NotificationCreationOrigin.DISCUSSION_DEFAULT,
+                    status=(NotificationSubscriptionStatus.ACTIVE
+                            if subscribed[cls.__mapper__.polymorphic_identity.name]
+                            else NotificationSubscriptionStatus.INACTIVE_DFT))
+                for cls in missing
```
ayyazdaniaryan: Clever use of generator execution to recall the query on `missing`.

load_testing/requirements.txt
```diff
@@ -1,35 +1,35 @@
...
-ws4py==0.3.4
+waitress==1.0.2
+WebOb==1.7.1
+WebTest==2.0.25
+ws4py==0.3.5
 WSGIProxy==0.2.2
 WSGIProxy2==0.4.2
 yg.lockfile==2.1.1
-zc.lockfile==1.1.0
+zc.lockfile==1.2.1
```
ayyazdaniaryan: This part was necessary for this PR? Could have been it's own PR, but I suppose the git gymnastics are too much. Okay, approved. One thing I would suggest is that Assembl developers update the docs, showing what tools/services they use for managing their PyPi and NPM dependencies. What do you think?

load_testing/scenarios.py
```diff
@@ -36,3 +37,26 @@ def test(url):
         test('/data/Discussion/{disc_id}/')
         test('/data/Discussion/{disc_id}/idea_links')
         test('/api/v1/discussion/{disc_id}/posts?{posts}&order=reverse_chronological&view=default')
+
+    def notification_subscriptions(self):
```
ayyazdaniaryan: Note to developers, the conditions to run this test vary widely from the typical use-case of running `pytest`.

ayyazdaniaryan: Hence why it is in not following the standard python testing templates.

ayyazdaniaryan: From what I know, have to run this test against an actual `uswgi` server, not pserve, in order to get the full benefits of multiprocessing. Install instructions [here](https://uwsgi-docs.readthedocs.io/en/latest/Install.html). On mac, it's `brew install uwsgi`. Then start the service in the assembl project like this:

```
uwsgi --die-on-term --ini local.ini --ini-paste-logged local.ini
```

PR #76, opened 2017-02-01
Add analytics to sentiments on message.

ayyazdaniaryan:
Please review the method in which events were added to the analytics client, and when they are called.

Also consider this question: Should analytics event be thrown at same time as sentiment ajax call to backend? Or should ajax call successfully complete, and then analytics event be thrown? Discuss.

### DIFF
--- a/assembl/static/js/app/views/message.js
+++ b/assembl/static/js/app/views/message.js
@@ -39,6 +39,15 @@ var MIN_TEXT_TO_TOOLTIP = 5,
     IDEA_CLASSIFICATION_LENGTH = 3;
 
 
+// Enum of conversion between sentiment to statistics
+var SENTIMENT_TO_STASTICS = {
+  LikeSentimentOfPost: 'like',
+  DisagreeSentimentOfPost: 'disagree',
+  DontUnderstandSentimentOfPost: 'idu',
+  MoreInfoSentimentOfPost: 'more',
+  DESELECT: 'deselect'
+};
+
 /**
  * @class app.views.message.IdeaClassificationNameListView
  * Classification view that is shown in the underneath each message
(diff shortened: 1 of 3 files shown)


maparent:
I personally think that piwik should register the intent of the user, not the system health, and eventual ajax call failure is not a piwik concern. Ajax success/failure is a raven concern.

maparent:
No comment on the code, lgtm but I don't know piwik, I'd like Quentin to review.
I wonder if we need separate signals for each sentiment; that's handled by my stats. At least I'd like to know that there's a way to get a consolidated "all sentiments" at the piwik level, because I suspect it's the level of analysis I would use there.

assembl/static/js/app/views/message.js
```diff
@@ -1127,6 +1150,40 @@ var MessageView = Marionette.LayoutView.extend({
...
+        analytics.trackEvent(analytics.events.SELECT_SENTIMENT_MORE);
+        break;
+      case 'deselect':
+        analytics.trackEvent(analytics.events.DESELECT_SENTIMENT);
+        break;
+      default:
+        console.warn("Improper sentiment was passed for analytics");
+    }
+  },
+
```
ayyazdaniaryan: Done. Could you check again the new format?

assembl/static/js/app/views/message.js
```diff
@@ -1056,12 +1066,14 @@ var MessageView = Marionette.LayoutView.extend({
       $('.emoticon').addClass('emoticon-connected');
     }
   },
-  onClickEmoticon:function(event){
+
+  onClickEmoticon: function(event){
+    console.log("Message Model:", this.model);
```
ayyazdaniaryan: Thanks! I forgot that one. Removed.

ayyazdaniaryan:
@maparent Yes, regarding registering intent of use of sentiments makes perfect sense. It's the approach I took as well. Also, regarding a general sentiment level event vs specific sentiment tracking, I can only say that this is what the requirements were. PM can use piwik for reporting purposes.

ayyazdaniaryan:
@paolina92 Branch is ready for merge?

PR #78, opened 2017-02-02
Fix routes

ayyazdaniaryan:
@paolina92 Here are the fixes to get routes for styleguide and profiles. There is now an issue with React router, which I'm not as familiar with.

### DIFF
--- a/assembl/static2/js/app/routes.jsx
+++ b/assembl/static2/js/app/routes.jsx
@@ -25,8 +25,8 @@ export default (
       <Route path=":slug/terms" component={Terms} />
     </Route>
     <Route path="/v2/">
-      <Route path=":slug/profile" component={Profile} />
-      <Route path=":slug/styleguide" component={Styleguide} />
+      <Route path="profile" component={Profile} />
+      <Route path="styleguide" component={Styleguide} />
     </Route>
     <Route path="*" component={NotFound} />
   </Router>
--- a/assembl/static2/js/app/routes.jsx
+++ b/assembl/static2/js/app/routes.jsx
@@ -16,6 +16,10 @@ export default (
   <Router>
     <Route path="/v2/" component={App}>
       <IndexRoute component={Home} />
+
+      <Route path="profile" component={Profile} />
+      <Route path="styleguide" component={Styleguide} />
+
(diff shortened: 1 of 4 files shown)


assembl/static2/js/app/routes.jsx
```diff
@@ -24,10 +28,6 @@ export default (
       <Route path=":slug/community" component={Community} />
       <Route path=":slug/terms" component={Terms} />
     </Route>
-    <Route path="/v2/">
-      <Route path=":slug/profile" component={Profile} />
-      <Route path=":slug/styleguide" component={Styleguide} />
-    </Route>
```
paolina92: Please let the style guide route outside of the App component:

<Router>
    <Route path="/v2/" component={App}>
        ........
     </Route>
     <Route path="/v2/">
        <Route path="styleguide" component={Styleguide} />
    </Route>
</Router>

paolina92: All components below App need a discussion ID. May be I have to rename "App" component because this is confusing.

paolina92: Please don't forget to run after your developments: `npm run elint`

assembl/views/discussion/views.py
```diff
@@ -175,25 +175,28 @@ def home_view(request):
     return response
 
 
+@view_config(route_name='new_home', request_method='GET', http_cache=60,
+             renderer='assembl:templates/index_react.jinja2')
+@view_config(route_name='new_styleguide', request_method='GET',
+             renderer='assembl:templates/index_react.jinja2')
+@view_config(route_name='new_profile', request_method='GET',
+             renderer='assembl:templates/index_react.jinja2')
```
paolina92: It's too bad to create routes in the backend and in the frontend. If I need to create ./v2/whatever, I have to do it in the backend also.
To resume, there is 2 cases: I need a discussion ID or I don't need a discussion ID. Is it possible to find a solution with this approach?

vincentfretin: Yeah, I think we just need two fallback routes that match all urls /{discussion}/v2/*reminder and /*reminder which render templates/index_react.jinja2 and you can remove 
new_styleguide and new_profile here.

PR #79, opened 2017-02-06
Search feature (wip)

vincentfretin:
Index posts, synthesis, ideas, idea announcements, users in elasticsearch, implement a search and livesearch.


assembl/static2/js/app/containers/search.jsx
```diff
@@ -0,0 +1,266 @@
...
+// TODO pagination component doesn't seem to work
+// TODO translations fr/en
+// TODO issue with sidebar range filter, it sometime disappears
+// (when there is only 1 result), and the selected range is not kept (maybe because of the date field...)
+
+const PostHit = (props) => {
+  return (
+    <div className={props.bemBlocks.item().mix(props.bemBlocks.container('item'))}>
+      <div className={props.bemBlocks.item('title')}>
+        <a
```
paolina92: Utiliser plutôt `Link` si possible

assembl/static2/js/app/containers/search.jsx
```diff
@@ -0,0 +1,271 @@
...
+          href={props.result._source.url}
+          dangerouslySetInnerHTML={{ __html: get(props.result, 'highlight.subject', props.result._source.subject) }}
+        />
+      </div>
+      <div className={props.bemBlocks.item('content')}>
+        <p dangerouslySetInnerHTML={{ __html: truncate(get(props.result, 'highlight.introduction', props.result._source.introduction)) }} />
+        <p dangerouslySetInnerHTML={{ __html: truncate(get(props.result, 'highlight.conclusion', props.result._source.conclusion)) }} />
+      </div>
+      <div className={props.bemBlocks.item('date')}>
+        { `Publié le ${props.result._source.creation_date}` }
```
paolina92: Pour les dates, tu peux utiliser le `Localize`de `react-redux-i18n`. Il y a un exemple parmi les derniers commit sur develop.

assembl/static2/js/app/containers/search.jsx
```diff
@@ -0,0 +1,271 @@
...
+   ActionBar, ActionBarRow
+} from 'searchkit';
+import get from 'lodash/get';
+
+import 'searchkit/theming/theme.scss';
+import '../../../css/views/search.scss';
+
+import GlobalFunctions from '../utils/globalFunctions';
+
+const truncate = (text) => {
```
paolina92: Peut-être un nom de méthode plus explicite ? On a du mal à comprendre ce que ça fait.

assembl/static2/js/app/containers/search.jsx
```diff
@@ -0,0 +1,266 @@
...
+   Layout, LayoutBody, LayoutResults,
+   SideBar,
+   TopBar,
+   ActionBar, ActionBarRow
+} from 'searchkit';
+import get from 'lodash/get';
+
+import 'searchkit/theming/theme.scss';
+import '../../../css/views/search.scss';
+
```
vincentfretin: Oui mais là on veut justement que ce soit local pour que le bundle searchv1.js ne contiennent que ces styles et pas l'ensemble de bootstrap 3 !

paolina92: ok j'avais mal compris en fait. C'est clair maintenant.

assembl/static2/js/app/containers/search.jsx
vincentfretin: Yes, that's the plan. react-router Link pour la v2 et un composant Link qui fait juste un rendu d'un <a href> pour la v1.

assembl/static2/js/app/containers/search.jsx
```diff
@@ -0,0 +1,266 @@
...
+  'subject_und', // post
+  'subject_fr', // post
+  'subject_en', // post
+  'body_und', // post
+  'body_fr', // post
+  'body_en' // post
+];
+
+export default class Search extends React.Component {
+
```
vincentfretin: done. Au début je pensais que j'allais utiliser redux, mais au final non.

assembl/static2/js/app/containers/search.jsx
vincentfretin: Ba je peux pas plus explicite, c'est pour tronquer du texte. D'ailleurs je vais plutôt utiliser la fonction de lodash https://lodash.com/docs/4.17.4#truncate

vincentfretin: Ah c'est vrai qu'elle ne fait pas seulement que tronquer ma fonction. :) Je vais voir ça.

assembl/static2/js/app/containers/search.jsx
vincentfretin: C'est fait dans https://github.com/assembl/assembl/pull/79/commits/b67fa211fa25f222b951659ff4e71667f763d5c0
Il n'y a plus qu'à créer les routes pour la v2.

assembl/views/search/views.py
```diff
@@ -0,0 +1,62 @@
...
+    creator_ids = set([hit['_source']['creator_id']
+                       for hit in result['hits']['hits']
+                       if hit['_source'].get('creator_id', None) is not None])
+    session = get_session_maker()
+    creators = session.query(models.AgentProfile.id, models.AgentProfile.name
+        ).filter(models.AgentProfile.id.in_(creator_ids)).all()
+    creators_by_id = dict(creators)
+    # u'query': {u'bool': {u'filter': [{u'term': {u'discussion_id': u'23'}}]}}
+    filters = [fil for fil in query['query']['bool']['filter']]
+    discussion_id = [f.values()[0].values()[0]
```
swergas: Are you sure this is enough to prevent elasticsearch from leaking private discussion data? I'm talking about people who could use this route and craft elasticsearch queries. We need to be really sure of this. I don't know Elasticsearch query language well enough to be sure that this code is secure enough. But this code transfers security concern from our current system to ES query language, which makes me feel unconfortable as the user can craft send any complex ES query (for example, can a user find a way to override our discussion_id filter, or to make multiple queries into one, where only one would have this filter?). I guess the other way would be that our backend does the ES search and then parses each hit to check if user is allowed to read this hit, but this is far from ideal. Maybe we should look a bit more into ES security modules (ACLs, etc for example via X-Pack/Shield?).

assembl/static2/css/views/search.scss
```diff
@@ -0,0 +1,346 @@
...
+    width: 20px;
+    height: 20px;
+    margin: 0 0 0 -14px;
+    padding:  2px 0 0 6px;
+    border-radius: 50%;
+    background-color: $highlightColor;
+}
+
+
+.sk-tag-filter {
```
swergas: Please add a comment to here to say that this style customizes the rendering of the searchkit external module and that these CSS classes names are chosen by this library, because it's not self-explanatory when you are looking at it outside of the context of this commit/PR.

assembl/lib/sqla.py
```diff
@@ -1690,12 +1691,14 @@ def orm_update_listener(mapper, connection, target):
         return
     session = object_session(target)
     if session.is_modified(target, include_collections=False):
+        reindex_content(target, 'update')
```
swergas: Should we really (re)index an object on every creation/update/deletion? If the indexation code fails, does it make the database operation fail? If yes, isn't it a problem? I would say that on indexation failure, the operation should still work, only the indexation of this object should be retried later. Would it be a good idea to debounce reindexation, in order to avoid congestion problems? Or does elasticsearch handles this by itself? A type of example case I'm thinking about is some very frequent user action in the UI that would update an object in the database each time: for this kind of actions we don't want to trigger (re)indexation as often as the action. These actions could include things like user likes a message (cf https://github.com/assembl/assembl/pull/79/commits/41f5582b3f8c7539747ab88fa904b3e0c9e128d4 ), user marks a message as read (even more when we will add a feature which automatically marks messages as read when they are displayed for a certain amount of time on screen.

assembl/templates/index.jinja2
```diff
@@ -78,6 +78,8 @@
       {% set userId = 'system.Everyone' %}
     {% endif %}
 
+    <script src="{{REACT_URL}}/build/searchv1.js"></script>
+    <!--<script src="{{ STATIC_URL }}2/build/searchv1.js"></script>-->
```
swergas: Please remove the commented script if it adds no value

assembl/static2/js/app/components/search.jsx
```diff
@@ -0,0 +1,647 @@
...
+      return `posts/${id}`;
+    }
+  };
+}
+
+const PublishedInfo = (props) => {
+  const { date, userId, userName, className } = props;
+  return (
+    <div className={className}>
+      <Translate value="search.published_the" />{' '}<Localize value={date} dateFormat="date.format" />
```
swergas: "published_the" and "Published the" is not good English I guess. Maybe replace with something like "publication_date" or "published_on".

assembl/static2/js/app/searchv1.jsx
```diff
@@ -0,0 +1,76 @@
...
+
+
+const myCreateStore = () => {
+  const store = createStore(RootReducer, applyMiddleware(Thunk));
+  const assemblLocale = window.assembl_locale.split('_')[0];
+  const userLocale = GlobalFunctions.getLocale(assemblLocale);
+  syncTranslationWithStore(store);
+  store.dispatch(loadTranslations(Translations));
+  store.dispatch(setLocale(userLocale));
+  return store;
```
swergas: I don't understand this commit https://github.com/assembl/assembl/pull/79/commits/e7d617d298dde0a1c72535d7ca13cab5ba633f0c , maybe that's just because I'm not familiar with react-redux-i18n.

assembl/views/search/views.py
```diff
@@ -0,0 +1,62 @@
...
+    if not in_tests():
+        changes._join()
+
+
+
+def includeme(config):
+    config.add_route('search', '/_search')
+    # join ElasticChanges datamanager to the transaction at the beginning
+    # of the request to avoid joining when the status is Committing
+    config.add_subscriber(join_transaction, NewRequest)
```
swergas: I don't understand what is happening here. Could you add more comments?

assembl/static2/js/app/searchv1.jsx
```diff
@@ -0,0 +1,76 @@
...
+    super(props);
+    this.state = { isExpert: false };
+  }
+  componentWillMount() {
+    const discussionId = GlobalFunctions.getDiscussionId();
+    const connectedUserId = GlobalFunctions.getConnectedUserId();
+    this.discussionId = discussionId;
+    this.connectedUserId = connectedUserId;
+    if (connectedUserId) {
+      fetchPermissionsForUser(discussionId, connectedUserId).then((permissions) => {
```
swergas: I think permissions of the current user are/should be already accessible without doing one more Ajax call. Either the backend sends it in the HTML page (like on Assembl v1), or received in the current user object when first fetched elsewhere in the application (and so available in the application state?). Anyways, https://github.com/assembl/assembl/pull/79/commits/de432650e23ae17466b78583beff1b2977458f6e is a very interesting commit :)

assembl/indexing/changes.py
```diff
@@ -0,0 +1,153 @@
+import logging
+import threading
+import os
+
+from transaction.interfaces import ISavepointDataManager, IDataManagerSavepoint
```
vincentfretin: The file changes.py is actually specific to assembl project because of the imports.
I prefer to keep all indexing related code in the indexing package if you don't mind.
Some files in lib folder contain actually domain-specific code, like lib/frontend_urls.py, so there is not really a clear separation here.

You can read http://transaction.readthedocs.io/en/latest/ to know more about data (resource) managers participating in a transaction. We have here two data managers in the transaction, sqlachelmy data manager from zope.sqlachelmy (see venv/lib/python2.7/site-packages/zope/sqlalchemy/datamanager.py), and here elasticsearch I wrote.
We use an elasticsearch data manager here to queue all changes and only put the changes to elasticsearch if postgres successfully committed the transaction.

When a commit occurs, the data managers are first sorted by their sortKey, sqlalchemy is first (key is "~sqlalchemy..."), elasticsearch is second (key is "ZZZZZ...").
Here is the workflow:
- sqlalchemy tpc_begin() does session.flush()
- elasticsearch (changes.py) tpc_begin() does nothing
- sqlalchemy commit() expires objects
- elasticsearch commit() does nothing
- sqlalchemy tpc_vote() commits to postgres. If an exception occurs, it will call tpc_abort on all resource managers, so clearing pending changes for elasticsearch data manager. 
- elasticsearch tpc_vote() does nothing
- sqlalchemy tpc_finish() does nothing
- elasticsearch tpc_finish() write to elasticsearch. If an exception occurs, tpc_abort will be called on all resources managers, but here the changes are already committed to postgres, so it actually doesn't really do anything. But you get a 500 http code.

For the savepoint stuff, you can actually at any moment in your code use `savepoint = transaction.savepoint()` and if there is an exception, rollback with `savepoint.rollback()` to return the transaction to this saved state. For postgres, this is implemented with nested transaction if supported.

I'll put those comments in the changes.py file.

assembl/views/search/views.py
vincentfretin: I tested some crafted query, for example having a discussion_id in the hasChild filter that is different from the one in the query, I didn't get any result. Indeed, the intersection of the main query and the hasChild filter is an empty set. And if you don't specify a discussion_id in a hasChild filter, all you will get is a wrong score (number of posts of an user).
The only way you can do a query and have results is by specifying a discussion_id in a bool query. Any query that don't have that will get you an HTTPUnauthorized exception.

After some thoughts, there is maybe a crafted query based on hasChild or hasParent with inner_hits where we can leak info. I'll get back to you on that. The solution would be to remove any inner_hits key from the query before doing the search.

As I understand the security model, you have global roles and only local roles on a discussion. Here we are indexing only PUBLISHED posts, and unindex them if their state change. Even if the posts are published, you may not have the permission to read them because you don't have the read permission on the discussion. The code here takes care of verifying you have the read permission on the discussion, and raise HTTPUnauthorized otherwise.
For the moment, the READ permission on a post is the same as the READ permission of the discussion, so checking the READ permission on the discussion is enough. But this may change in the future. Checking the READ permission for each result may be a good idea. I already retrieve the object for an idea or an agent profile to get the counters, so verifying additional permissions shouldn't be much overhead. We only have 5 results here.

The security modules proposed by X-Pack is a paid service, you need to buy a license. As far I understand it's to protect the elasticsearch rest api if you expose it externally, and the ACLs is I suppose to give the permissions of some endpoints and HTTP verbs to specific users.
We don't need this. We don't expose elasticsearch externally. We only expose it via the pyramid view and you can only do GET localhost:9200/_search here because we do the query via the elasticsearch python binding es.search.

assembl/views/search/views.py
vincentfretin: Read the first comment to have more info about data managers participating to the transaction.
Here I join the elasticsearch data manager to the transaction at the beginning of the request.
If I don't do that, the elasticsearch data manager may actually join the transaction when the transaction is committing, and this fail with an error. The after_insert sqlalchemy event that I use to index an object (and so the elasticsearch data manager join the transaction) is triggered via the session.flush() in the tcp_begin of sqlalchemy data manager. You can't add data managers to a transaction which is in a Committing state.

assembl/static2/css/views/search.scss
vincentfretin: Ok, I'll add a comment.

assembl/lib/sqla.py
vincentfretin: If the indexation fails, the postgres transaction is already committed, so there is no loss of data here.

If an object is modified several times during the transaction, only the last modification (including all previous changes) is kept. There is only a single request to elasticsearch at the end of the transaction.
We are using the elasticsearch bulk rest api here to index several documents with one request. It's just a PUT request and it returns immediately. The indexing in elasticsearch is then done asynchronously.
If by any bad luck, the elasticsearch is not responding, the postgres database and elasticsearch will be out of sync. We can always reindex completely the elasticsearch index which take 1 or 2 minutes to sync it again with the postgres database.

assembl/templates/index.jinja2
vincentfretin: Ok, will do.

assembl/static2/js/app/components/search.jsx
vincentfretin: Ok, published_on it is.

assembl/static2/js/app/searchv1.jsx
vincentfretin: I take the language of the v1 interface (the window.assembl_locale global variable), then I load the translations for this language like we do in v2 (see store.js)
In v2, we manage the selected locale in the store. In v1, it's a global variable assembl_locale.

assembl/static2/js/app/searchv1.jsx
vincentfretin: This is indeed the case for the v2, where the information is already in the redux store.
For v1, I don't have the info in the store, this is why I do an extra ajax call. But indeed I just saw there is ```<script type="text/json" id="permissions-json">["read", "add_post", "edit_post", "add_extract", "edit_extract", "add_idea", "edit_idea", "edit_synthesis", "send_synthesis", "admin_discussion", "sysadmin", "edit_my_extract", "self_register", "self_register_req", "vote", "read_public_cif", "export_post", "moderate_post", "discussion_stats", "delete_my_post", "delete_post"]</script>``` on the page, so I guess I can use that instead.

assembl/static2/js/app/components/search.jsx
swergas: check

assembl/templates/index.jinja2
swergas: check

assembl/static2/css/views/search.scss
swergas: check

assembl/static2/js/app/searchv1.jsx
swergas: check

assembl/indexing/changes.py
swergas: check

assembl/views/search/views.py
swergas: ok for your fix https://github.com/assembl/assembl/pull/84/commits/a5126bb1356e086a72e5fdfadd05aade441f64a5

assembl/views/search/views.py
swergas: OK I commited your comment https://github.com/assembl/assembl/commit/cd5f7acd9c4d3ba56d060d77b53e1cbae22231ba

assembl/lib/sqla.py
swergas: OK I added your comment in https://github.com/assembl/assembl/commit/f0be89cae3277fbe51364cbc59cd6b0295f26943

assembl/static2/js/app/searchv1.jsx
swergas: check

swergas:
Code Review phase of this PR is now complete. This PR is followed by https://github.com/assembl/assembl/pull/84 .

PR #80, opened 2017-02-08
Improve UI of share feature

swergas:


swergas:
Known drawback of this new UI: as it now appears in a modal, if another popin was already shown in the UI (for example after having clicked on the name of a user to browse his messages), then the share popin replaces the previous popin (popin modals do not stack in our current system).

assembl/static/js/app/common/context.js
```diff
@@ -613,6 +613,14 @@ Context.prototype = {
...
     }
 
+    var modalClass = "";
+    if ( evt && evt.currentTarget && $(evt.currentTarget).attr("data-modal-class") ){
+      modalClass = $(evt.currentTarget).attr("data-modal-class");
+    }
+    if ( _.isObject(options) && "modal_class" in options ){
+      modalClass = options.modal_class;
+    }
+
```
paolina92: Mysterious code...

assembl/static/widget/share/index.html
```diff
@@ -22,29 +19,261 @@
 	box-sizing: border-box;
 	width: 100%;
 	max-width: 1000px;
+	padding: 3px;
```
paolina92: Is it necessary to put css in the index? Is it possible to create a specific css file?

PR #83, opened 2017-03-03
Timeline objects and back-office

maparent:


assembl/templates/views/adminTimelineEvent.tmpl
```diff
@@ -0,0 +1,42 @@
+<div class="subpanel-body bx">
+  <div>
+    <div class="admin-msg-elm">
+      <div class="control-group">
+        <label for="start_date" class="control-label">{{ gettext('Start date') }}</label>
+        <div class="controls">
+          <input class="js_start_date input-medium" id="start_date" name="start_date" type="datetime-local" value="<%= event.getStartNoTZ() %>"></input>
```
ayyazdaniaryan: OOOOhhhhhh, using an HTML5 datetime picker? And there is no post-processing on the server? Yeesh...I haven't used it before, but just tested it on chrome, and it only returns the local date and time. No timezone. So, we'll in effect have wacky hours in our database based on location, no? This one might need some processing, either on front-end or backend. Possibly have `moment.js` convert them to UTC, or append the timezone, so that the times have some relevance to each other?

ayyazdaniaryan: Also, ONLY supported on Chrome and Edge (shudders) http://caniuse.com/#feat=input-datetime . Known fact?

assembl/views/api2/timeline.py
```diff
@@ -0,0 +1,52 @@
...
+    previous_event.previous_event_id = None
+    instance.previous_event_id = None
+    if next_event:
+        next_event.previous_event_id = None
+    db.flush()
+    instance.previous_event_id = pre_previous_id
+    previous_event.previous_event_id = instance.id
+    if next_event:
+        next_event.previous_event_id = previous_event.id
+    return HTTPOk()
```
ayyazdaniaryan: Or, return the timeline structure? Eh, either works, no standards for it.

assembl/alembic/versions/0888e0f1a92d_timeline.py
```diff
@@ -0,0 +1,102 @@
...
+            for d_id, ev_id, title, description in data:
+                title = m.LangString.create(title, locales[d_id])
+                description = m.LangString.create(description, locales[d_id])
+                db.add(title)
+                db.add(description)
+                db.flush()
+                db.execute(
+                    'UPDATE timeline_event SET title_id=%d, description_id=%d WHERE id = %d' % (
+                    title.id, description.id, ev_id))
+            mark_changed()
```
maparent: with an execute, sqla doesn't know if I'm doing a query or update, so it does not know whether it should end with a commit or rollback. mark_changed forces commit.

assembl/models/timeline.py
```diff
@@ -44,12 +45,26 @@ class TimelineEvent(DiscussionBoundBase):
...
         info={'rdf': QuadMapPatternS(None, DCTERMS.description)})
 
+    title = relationship(LangString,
+        lazy="joined", single_parent=True,
+        primaryjoin=title_id == LangString.id,
+        backref=backref("title_of_timeline_event", lazy="dynamic"),
+        cascade="all, delete-orphan")
+
+    description = relationship(LangString,
+        lazy="joined", single_parent=True,
```
maparent: joined: I want the description to load with the timeline.
dynamic: I want to update the langstring without having to traverse back to the timeline unless necessary.

assembl/views/api2/timeline.py
```diff
@@ -0,0 +1,52 @@
...
+from ..traversal import InstanceContext, CollectionContext
+from . import (
+    FORM_HEADER, JSON_HEADER, instance_put_json, instance_put_form,
+    collection_add_json, collection_add_with_params, instance_del)
+
+
+@view_config(
+    context=InstanceContext, request_method='DELETE', renderer='json',
+    ctx_instance_class=TimelineEvent)
+def event_del(request):
```
maparent: correct.

assembl/views/api2/timeline.py
maparent: Would also make sense, I guess. This is a RPC call, not a restful call; allows atomicity.

assembl/static/js/app/views/admin/adminTimelineEvents.js
```diff
@@ -0,0 +1,244 @@
...
+        title = new LangString.Model(),
+        titles = {},
+        description = new LangString.Model(),
+        descriptions = {},
+        preferences = Ctx.getPreferences();
+    if (eventCollection.length > 0) {
+      lastEventId = eventCollection.models[eventCollection.length-1].id;
+    }
+    _.map(preferences.preferred_locales, function(loc) {
+      titles[loc] = '';
```
maparent: no, no! it's setting the known locales to empty string values. und is for empty locale values.

assembl/static/js/app/views/admin/adminTimelineEvents.js
```diff
@@ -0,0 +1,244 @@
...
+      descriptions[loc] = '';
+    });
+    title.initFromDict(titles);
```
maparent: in models.langstring.js:445

assembl/static/js/app/views/admin/adminTimelineEvents.js
```diff
@@ -0,0 +1,244 @@
...
+    description.initFromDict(titles);
+    event = new TimelineEvent.Model({
+      title: title,
+      description: description,
+      previous_event: lastEventId,
+    });
+    eventCollection.add(event);
+    event.save();
```
maparent: Your solution works for creation, but I think subsequent edition is easier with my approach; otherwise I have to know how much the frontend model has diverged from the backend model. So I used that approach for everything because I could.

assembl/static/js/app/views/admin/adminTimelineEvents.js
```diff
@@ -0,0 +1,244 @@
...
+        owner_relative_url: this.model.url() + '/description',
+      }));
+  },
+  reorderColumnDown: function(ev) {
+    var index = this.getIndex(),
+        nextModel = this.model.collection.at(index + 1);
+    Promise.resolve($.ajax(nextModel.url() + "/reorder_up", {
+        method: "POST"})).then(function(data) {
+        nextModel.collection.fetch({
+          success: function() {
```
maparent: Good point, I'll do this later today.

assembl/static/js/app/views/admin/adminTimelineEvents.js
```diff
@@ -0,0 +1,244 @@
...
+  },
+  deleteColumn: function(ev) {
+    var nextModel = null,
+        prevColumn = this.model.get('previous_event'),
+        index = this.getIndex();
+    if (index + 1 < this.model.collection.length) {
+      nextModel = this.model.collection.at(index+1);
+    }
+    this.model.destroy({
+      success: function() {
```
maparent: I'm making sure it's a single backend call. Two calls can fail independently and leave the model in an inconsistent state.

assembl/static/js/app/utils/types.js
```diff
@@ -28,6 +28,10 @@ var Types = {
   TABLE_OF_CONTENTS: 'TableOfIdeas',
   USER: 'User',
   PARTNER_ORGANIZATION: 'PartnerOrganization',
+  TIMELINE_EVENT: 'TimelineEvent',
```
maparent: I need to add comments. I forgot a bit myself, to be honest.

assembl/templates/views/adminTimelineEvent.tmpl
maparent: No timezone: I do some postprocessing to add/remove the "Z", you may have missed it.
Partial support: well, it falls back to textfield behaviour, which is not that bad, and then it'll validate on fronted. I agree this is less than satisfactory, it was coded fast. Using moment.js would be a nice improvement.

assembl/static/js/app/views/admin/adminTimelineEvents.js
maparent: Note: the fallback logic for failure is to ask to remove the page... which was good enough for me in an admin view. The growl has to be followed by a model reload anyhow. Are you sure it's necessary?

assembl/static/js/app/models/timeline.js
```diff
@@ -0,0 +1,117 @@
...
+    } else if (e2.id == e1p) {
+      return 1;
+    } else if (e1p === null) {
+      return -1;
+    } else if (e2p === null) {
+      return 1;
+    } else {
+      e1 = this.get(e1p);
+      if (e1 !== undefined) {
+        return this.comparator(e1, e2);
```
maparent: When is recursion not the right answer to a problem? :->

assembl/alembic/versions/0888e0f1a92d_timeline.py
ayyazdaniaryan: Gotcha, good reminder. There are two transactions to be aware of, the zope one (for pyramid) and the one for alembic...

assembl/static/js/app/models/timeline.js
ayyazdaniaryan: :)

assembl/static/js/app/utils/types.js
ayyazdaniaryan: That would be lovely, thank you.

assembl/static/js/app/views/admin/adminTimelineEvents.js
ayyazdaniaryan: Ah gotchaaa. Make sense!

assembl/static/js/app/views/admin/adminTimelineEvents.js
ayyazdaniaryan: Yes, I agree that we'd have to do the housecleaning on post-creation changes. So, perhaps the chosen method is the way to go. The code had to be written anyway.

assembl/static/js/app/views/admin/adminTimelineEvents.js
ayyazdaniaryan: It's not that important to have a growl, if the AJAX call fails anyway on timeout/server crash. At least there is some indication that the action did not follow through.

assembl/static/js/app/views/admin/adminTimelineEvents.js
ayyazdaniaryan: It's clear. And indeed going the socket route might not be very OO, with guard code sprinkled everywhere. Makes delicious spaghetti though...

assembl/templates/views/adminTimelineEvent.tmpl
ayyazdaniaryan: Okay, so our agreement on this was to take the time to comment what time indicates, and be more explicit about the timezone stored. Furthermore, comment in code made for front-end to be aware that the incoming datetime from backend is in UTC.

assembl/views/api2/timeline.py
ayyazdaniaryan: 👍

assembl/views/api2/timeline.py
ayyazdaniaryan: I'm more than okay with it. Just compared it to other manual end-points. We don't have much consistency across our APIs that are not V2 (like these `view_config`s). That's a discussion for another day.

PR #84, opened 2017-03-10
Code Review of the search feature branch

vincentfretin:
Some changes after the code review of https://github.com/assembl/assembl/pull/79

### DIFF
--- a/assembl/views/search/views.py
+++ b/assembl/views/search/views.py
@@ -54,6 +54,12 @@ def search_endpoint(context, request):
     for hit in result['hits']['hits']:
         source = hit['_source']
         creator_id = source.get('creator_id', None)
+        # Remove inner_hits key to not leak posts from private discussion.
+        # You can easily craft a query to get the participants of a public
+        # discussion and do a has_child filter with inner_hits on a private discussion.
+        if 'inner_hits' in hit:
+            del hit['inner_hits']
+
         if creator_id is not None:
             source['creator_name'] = creators_by_id.get(creator_id)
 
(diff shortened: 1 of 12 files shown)


vincentfretin:
FYI, between #79 and this PR, there were the following commits directly on the develop branch:
https://github.com/assembl/assembl/commit/523bcd61a414674d6dcab7d13be9b3d83ab985b9
https://github.com/assembl/assembl/commit/0631ce3771603cb91e7fa6ce0c7916d95c1be015
https://github.com/assembl/assembl/commit/38c789f2477bb72f2b7084fc136e5c2beff816dc

assembl/views/search/views.py
```diff
@@ -59,10 +59,27 @@ def search_endpoint(context, request):
...
-            source['num_posts'] = models.AgentProfile.get_instance(source['id']).count_posts_in_discussion(discussion_id)
+            agent_profile = models.AgentProfile.get_instance(source['id'])
+            if not agent_profile.user_can(user_id, CrudPermissions.READ, permissions):
+                raise HTTPUnauthorized
+
+            source['num_posts'] = agent_profile.count_posts_in_discussion(discussion_id)
+        # Don't do an extra request to verify the CrudPermissions.READ permission
+        # for post or synthesis.
+        # It's currently the same 'read' permission as the discussion.
+        # elif hit['_type'] in ('synthesis', 'post'):
```
swergas: If you do the check for an Idea and for a User (to verify that the user has the permission to see it), why not do it also for a Post?

vincentfretin: For idea and user, I already get the instance to get counters, so checking the permission doesn't do an extra sql query. For a post, I don't need to get the instance. So here I don't check the permission because the CrudPermissions.READ permission for post and discussion is actually the same, it's the "read" permission that we already checked at the beginning.

PR #86, opened 2017-03-27
First part of the Phase 1

paolina92:
The scope of the feature is defined here:

https://bluenove.atlassian.net/browse/DEVAS-321


assembl/static2/js/app/client.js
```diff
@@ -0,0 +1,7 @@
+import ApolloClient, { createNetworkInterface } from 'apollo-client';
+
+const client = new ApolloClient({
+  networkInterface: createNetworkInterface({ uri: 'http://localhost:3000/graphql' })
```
vincentfretin: I think there is a backend task to do here, we should have in the page the url of the graphql endpoint, which is the domain/slug/graphql.

assembl/static2/js/app/components/common/navbar.jsx
```diff
@@ -21,7 +20,9 @@ class NavBar extends React.Component {
     });
   }
   componentWillReceiveProps() {
-    this.setState({ isMenuHidden: true });
+    this.state = {
+      isMenuHidden: true
+    };
```
vincentfretin: You should only use 'this.state ='  in the constructor, everywhere else use setState.

paolina92: Except here, I'm pretty sure that's what I do

PR #87, opened 2017-03-29
Phase1

ayyazdaniaryan:


assembl/graphql/schema.py
```diff
@@ -1,81 +1,439 @@
...
+        return subject
+
+    def resolve_body(self, args, context, info):
+        body = resolve_langstring(self.get_body(), args.get('lang'))
+        return body
 
+    def resolve_sentiment_counts(self, args, context, info):
+        sentiment_counts = self.sentiment_counts
+        return SentimentCounts(
+            dont_understand=sentiment_counts['dont_understand'],
```
ayyazdaniaryan: Is this a safe way to get the keys? Cannot usr `getattr` or maybe even `get`?

assembl/graphql/schema.py
```diff
@@ -1,81 +1,439 @@
...
+    def resolve_posts(self, args, context, info):
+        # TODO
+        return []
+
+
+class Thematic(SecureObjectType, SQLAlchemyObjectType):
+    class Meta:
+        model = models.Thematic
+        interfaces = (Node, )
+        only_fields = ('id', 'identifier')
```
ayyazdaniaryan: Difference between `id` and `identifier`?

assembl/graphql/schema.py
```diff
@@ -1,81 +1,439 @@
...
+    posts = SQLAlchemyConnectionField(PostConnection)#, idea_id=graphene.ID())
+    # ideas = SQLAlchemyConnectionField(Idea)
+    thematics = graphene.List(Thematic, identifier=graphene.String())
+    # agent_profiles = SQLAlchemyConnectionField(AgentProfile)
+
+    def resolve_ideas(self, args, context, info):
+        # TODO do we need to do a special query to only retrieve ideas
+        # associated to the RootIdea?
+        connection_type = info.return_type.graphene_type  # this is IdeaConnection
+        model = connection_type._meta.node._meta.model  # this is models.Idea
```
ayyazdaniaryan: Is this really the cleanest way to get the Model type?

assembl/graphql/schema.py
```diff
@@ -1,81 +1,439 @@
...
+        connection_type = info.return_type.graphene_type  # this is PostConnection
+        model = connection_type._meta.node._meta.model  # this is models.PostUnion
+        query = get_query(model, context)
+        discussion_id = context.matchdict['discussion_id']
+        query = query.filter(model.discussion_id == discussion_id
+            ).filter(model.hidden == False
+            ).filter(model.tombstone_condition()
+            ).filter(model.publication_state == models.PublicationStates.PUBLISHED)
+
+        # TODO filter posts for a specific idea
```
ayyazdaniaryan: What are you trying to do here exactly? Does post take an argument?

assembl/graphql/schema.py
```diff
@@ -1,81 +1,439 @@
...
+class CreateThematic(graphene.Mutation):
+    class Input:
+        # Careful, having required=True on a graphene.List only means
+        # it can't be None, having an empty [] is perfectly valid.
+        title_entries = graphene.List(LangStringEntryInput, required=True)
+        description_entries = graphene.List(LangStringEntryInput)
+        identifier = graphene.String(required=True)
+        video = graphene.Argument(VideoInput)
+        questions = graphene.List(QuestionInput)
+        # TODO upload img example http://docs.pylonsproject.org/projects/pyramid-cookbook/en/latest/forms/file_uploads.html
```
ayyazdaniaryan: We've done something like this from the front-end for attachment upload. Using `multipart/form-data` format. But it's in Backbone.

assembl/graphql/schema.py
```diff
@@ -1,81 +1,439 @@
...
+    @staticmethod
+    def mutate(root, args, context, info):
+        cls = models.Thematic
+        discussion_id = context.matchdict['discussion_id']
+        user_id = context.authenticated_userid or Everyone
+
+        permissions = get_permissions(user_id, discussion_id)
+        allowed = cls.user_can_cls(user_id, CrudPermissions.CREATE, permissions)
+        if not allowed or (allowed == IF_OWNED and user_id == Everyone):
+            raise HTTPUnauthorized()
```
ayyazdaniaryan: 👍

assembl/graphql/schema.py
```diff
@@ -1,81 +1,439 @@
...
+
+        identifier = args.get('identifier')
+        with cls.default_db.no_autoflush:
```
ayyazdaniaryan: Why on `no_autoflush`? Because of the exception thrown?

assembl/graphql/schema.py
```diff
@@ -1,81 +1,439 @@
...
+                    kwargs['video_description'] = video_description
+
+                kwargs['video_html_code'] = video['html_code']
+
+            saobj = cls(
+                discussion_id=discussion_id,
+                title=title_langstring,
+                identifier=identifier,
+                **kwargs)
+            db = saobj.db
```
ayyazdaniaryan: You have `default_db` at your disposal already

assembl/graphql/schema.py
ayyazdaniaryan: You haven't created the SQLA object yet, why care about the session being flushed on error?

assembl/graphql/schema.py
```diff
@@ -1,81 +1,439 @@
...
+                        question_input['title_entries'])
+                    saobj.children.append(
+                        models.Question(
+                            title=title_ls,
+                            discussion_id=discussion_id
+                        )
+                    )
+                db.flush()
+
+        return CreateThematic(thematic=saobj)
```
ayyazdaniaryan: Wtf...I followed you until this line. Is this how Graphene `mutate` method works?!

assembl/tests/views_tests/test_graphql.py
```diff
@@ -0,0 +1,273 @@
...
+import json
+
+import pytest
+from graphql_relay.node.node import to_global_id
+
+from assembl import models
+from assembl.graphql.schema import Schema as schema
+
+
+@pytest.fixture(scope="function")
```
vincentfretin: This fixture makes sense only in this module. It is not meant to be shared in other modules. I'd prefer it to be colocated with the tests. If later we have several test modules where it can be useful, yeah it makes sense to move it now. I'll do it.

assembl/alembic/versions/14e5db4fde8e_add_tables_needed_for_phase_1.py
```diff
@@ -0,0 +1,50 @@
...
+            sa.Column('title_id', sa.Integer,
+                sa.ForeignKey('langstring.id'), nullable=False),
+        )
+
+
+def downgrade(pyramid_env):
+    with context.begin_transaction():
+        op.drop_table('thematic')
+        op.drop_table('question')
+        op.drop_table('proposition')
```
vincentfretin: oups, I'll remove this line.

assembl/alembic/versions/b0871ff12f95_thematic_identifier.py
```diff
@@ -0,0 +1,26 @@
...
+down_revision = '14e5db4fde8e'
+
+from alembic import context, op
+import sqlalchemy as sa
+
+
+def upgrade(pyramid_env):
+    with context.begin_transaction():
+        op.add_column(
+            'thematic',
```
vincentfretin: Ok, will do that then!

assembl/tests/views_tests/test_graphql.py
```diff
@@ -0,0 +1,273 @@
...
+
+def test_get_thematics_noresult(graphql_request):
+    res = schema.execute('query { thematics(identifier:"survey") { id, title, description, numPosts, numContributors, questions { title }, video {title, description, htmlCode} } }', context_value=graphql_request)
+    assert json.loads(json.dumps(res.data)) == {u'thematics': []}
+
+
+def test_get_thematics_no_video(discussion, graphql_request, test_session):
+    title = u"Comprendre les dynamiques et les enjeux"
+    title = models.LangString.create(title, locale_code="fr")
+    thematic = models.Thematic(
```
vincentfretin: Right, I now use this code twice. Indeed it make sense to refactor it into a fixture.

assembl/tests/views_tests/test_graphql.py
```diff
@@ -0,0 +1,273 @@
...
+        thematic {
+            title,
+            identifier
+            video {title, description, htmlCode}
+        }
+    }
+}
+""", context_value=graphql_request)
+    assert json.loads(json.dumps(res.data)) == {
+        u'createThematic': {
```
vincentfretin: No, it doesn't matter if it's ascii, if there is accent, I don't know. The keys and values in res.data are unicode, so I use unicode in the expected result too.

assembl/graphql/schema.py
```diff
@@ -1,81 +1,439 @@
...
+        return langstring
+
+    return None
+
+
+class AgentProfile(SecureObjectType, SQLAlchemyObjectType):
     class Meta:
         model = models.AgentProfile
-        interfaces = (URINode, )
+        interfaces = (Node, )
```
vincentfretin: Yeah, it's the graphql way to have opaque id. The client doesn't need to know what the id means. The URINode you did actually added extra complexity, and I think it was not really accurate. You can have several types in GraphQL that share the same sqlalchemy type.

assembl/graphql/schema.py
vincentfretin: It's safe, the self.sentiment_counts property initialize to 0 the 4 sentiments.

assembl/graphql/schema.py
vincentfretin: identifier is used as an identifier for the phase, it's the same behavior that for the timeline.

assembl/graphql/schema.py
vincentfretin: Yeah if you want to retrieve what you wrote in the Meta class of the Idea objecttype, it's the only way. I wrote this code to maybe generalize it at one point so I can use it in other resolve functions.

assembl/graphql/schema.py
vincentfretin: Yeah we could have a idea_id argument to resolve_posts to only retrieve the posts from this idea. We don't need the ideas and posts for now. So we will discuss what we want in futures sprints. I disabled the resolve_ideas for now too. We can leave the resolve_posts as it is for now.

assembl/graphql/schema.py
vincentfretin: Great, I'll look at it.

assembl/graphql/schema.py
vincentfretin: An implicit flush can be triggered when you add objects to relations and may throw an error because there is no id yet associated to the parent.
I didn't test it without the no_autoflush to see if it was actually ok or not. I just did the same as we have now in the api2.

assembl/graphql/schema.py
vincentfretin: Lol. What is bothering you? In a mutation, you can return a result, whatever you want. Here we return the created thematic. The client can say what it wants to be returned, specific fields of the thematic.

assembl/alembic/versions/b0871ff12f95_thematic_identifier.py
ayyazdaniaryan: I know this is being nitpicky. Thanks

assembl/tests/views_tests/test_graphql.py
ayyazdaniaryan: Should we assume to have more graphql tests that are not based on thematic and questions? If so, then perhaps pulling the fixture out would be useful.

assembl/alembic/versions/b0871ff12f95_thematic_identifier.py
vincentfretin: It's done in https://github.com/assembl/assembl/pull/87/commits/c8b1c038db79ac28a312a7f489bc5b7ac0b12478

assembl/graphql/schema.py
ayyazdaniaryan: But can you have several GraphQL types with an ID that is identical to their sqlalchemy type? Meaning, the specific model? Anyways, I understand that using `URINode` added more complexity without much benefit. This is only used by Graphene to track its cache, and Apollo on front-end for the magic it does (pagination, etc). I don't mind it being `Node`

assembl/graphql/schema.py
ayyazdaniaryan: Right, makes sense now.

assembl/graphql/schema.py
vincentfretin: I'm not sure I understand your question here. Graphene opaque global id is actually just base64 encoded graphene_type:sqlalchemy_id

assembl/graphql/schema.py
ayyazdaniaryan: AHhhh okay. I lost you because you're returning the `Mutation` class with the object identified. It does make sense looking back up to what GraphQL does.

assembl/graphql/schema.py
ayyazdaniaryan: Right. Fair enough. I don't think there's any reason to change this, and yeah, API v2 uses it a lot as it is forming objects and relations via json. I guess you're not doing something that different.

assembl/graphql/schema.py
ayyazdaniaryan: This is what you want to look at: https://github.com/homeslicesolutions/backbone-model-file-upload

assembl/graphql/schema.py
ayyazdaniaryan: Cool. Sounds good!

assembl/graphql/schema.py
ayyazdaniaryan: Fair enough. It looks very...javascripty. But, it's understandable.

assembl/graphql/schema.py
ayyazdaniaryan: Okay, fair enough

assembl/tests/views_tests/test_graphql.py
vincentfretin: Actually it's not the same code, one is a thematic without video, and the other one with a video, I can't factor it.

PR #93, opened 2017-04-21
Require the "right" sso (OR right email) for self-registration in autologin case.

maparent:
Also repair next-view on autologin.

### DIFF
--- a/assembl/models/discussion.py
+++ b/assembl/models/discussion.py
@@ -363,18 +363,28 @@ class Discussion(DiscussionBoundBase, NamedClassMixin):
 
     def check_authorized_email(self, user):
         # Check if the user has a verified email from a required domain
+        from .social_auth import SocialAuthAccount
         require_email_domain = self.preferences['require_email_domain']
-        if not require_email_domain:
+        autologin_backend = self.preferences['authorization_server_backend']
+        if not (require_email_domain or autologin_backend):
             return True
         for account in user.accounts:
             if not account.verified:
                 continue
-            email = account.email
-            if not email or '@' not in email:
-                continue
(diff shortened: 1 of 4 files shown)


assembl/models/social_auth.py
```diff
@@ -135,6 +135,15 @@ class SocialAuthAccount(
     def provider(self):
         return self.identity_provider.provider_type
 
+    @property
+    def provider_with_idp(self):
+        provider = self.provider
+        if provider == 'saml':
+            idp_name = self.uid.split(':')[0]
```
ayyazdaniaryan: A comment on a typical shape `self.uid` would go far here.

assembl/models/discussion.py
```diff
@@ -359,18 +359,28 @@ def get_discussion_urls(self):
 
     def check_authorized_email(self, user):
         # Check if the user has a verified email from a required domain
+        from .social_auth import SocialAuthAccount
         require_email_domain = self.preferences['require_email_domain']
-        if not require_email_domain:
+        autologin_backend = self.preferences['authorization_server_backend']
```
ayyazdaniaryan: Is there anywhere in the discussion preferences or in a documentation page where we can describe the shape and content of this pref? If it was an ini thing, there would be a clear trail of what this is supposed to entail, for example...

assembl/models/social_auth.py
ayyazdaniaryan: When does this also get used?

maparent: The IDP name is used all over the place, but this new method is only used in Discussion.check_authorized_email

PR #95, opened 2017-04-28
Utils refactoring

paolina92:
Quick refactoring


assembl/static2/js/app/utils/utilityManager.js
```diff
@@ -0,0 +1,37 @@
...
+  setComponent(comp) {
+    this.component = comp;
+  }
+};
+
+export const modalManager = new UtilityManager();
+
+export const alertManager = new UtilityManager();
+
+export const displayAlert = (style, msg, topPosition = false, time = 6000) => {
```
ayyazdaniaryan: I'm okay with this, _for now_, but I wanted to write methods in the future such as `displayError`, `displaySuccess`, etc etc. This will start to make the file feel polluted.

assembl/static2/package.json
```diff
@@ -57,6 +57,7 @@
     "isomorphic-fetch": "^2.2.1",
     "lodash": "^4.17.4",
     "moment": "^2.17.1",
+    "prop-types": "^15.5.8",
```
ayyazdaniaryan: In prep for react upgrade?

assembl/static2/js/app/utils/utilityManager.js
```diff
@@ -0,0 +1,37 @@
...
+    showAlert: true,
+    topPosition: topPosition
+  });
+  setTimeout(() => {
+    this.alertComponent.setState({
+      showAlert: false
+    });
+  }, time);
+};
+export const displayModal = (title, body, footer, footerTxt, button, showModal) => {
```
ayyazdaniaryan: Also, let's comment a bit more. Or put examples in the styleguide? From reading this method, it's unclear what the type of each parameter should be.

ayyazdaniaryan: Would have to read the component's definition, etc...

assembl/static2/package.json
paolina92: yes

assembl/static2/js/app/utils/utilityManager.js
paolina92: You're right. I should update the style guide and add a comment.

PR #96, opened 2017-05-01
json_error_view

maparent:
The error view will actually have the json type. Change parameter order.

### DIFF
--- a/assembl/static2/js/app/components/login/sendPwdForm.jsx
+++ b/assembl/static2/js/app/components/login/sendPwdForm.jsx
@@ -16,8 +16,9 @@ class SendPwdForm extends React.Component {
   }
   componentWillReceiveProps(nextProps) {
     const { auth } = nextProps;
-    if (auth.passwordChangeRequest && auth.passwordChangeRequest.success === false) {
-      const firstError = auth.passwordChangeRequest.data[0];
+    const resp = auth.passwordChangeRequest;
+    if (resp && resp.success === false) {
+      const firstError = resp.data[0];
       const msg = firstError.message;
       displayAlert('danger', msg, true);
     }
--- a/assembl/static2/js/app/components/login/sendPwdForm.jsx
+++ b/assembl/static2/js/app/components/login/sendPwdForm.jsx
@@ -17,10 +17,8 @@ class SendPwdForm extends React.Component {
   componentWillReceiveProps(nextProps) {
     const { auth } = nextProps;
     if (auth.passwordChangeRequest && auth.passwordChangeRequest.success === false) {
-      let msg;
-      if (auth.passwordChangeRequest.data === 404) {
-        msg = I18n.t('login.emailNotFound');
-      } else { msg = I18n.t('login.passwordChangeRequestError'); }
+      const firstError = auth.passwordChangeRequest.data[0];
+      const msg = firstError.message;
       displayAlert('danger', msg, true);
     }
   }
--- a/assembl/static2/js/app/utils/httpRequestHandler.js
+++ b/assembl/static2/js/app/utils/httpRequestHandler.js
@@ -43,22 +43,15 @@ export const xmlHttpRequest = (obj) => {
         }
         resolve(resp);
       } else {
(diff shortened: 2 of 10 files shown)


assembl/static2/js/app/components/login/sendPwdForm.jsx
```diff
@@ -17,10 +17,8 @@ class SendPwdForm extends React.Component {
   componentWillReceiveProps(nextProps) {
     const { auth } = nextProps;
     if (auth.passwordChangeRequest && auth.passwordChangeRequest.success === false) {
```
paolina92: `if (auth.passwordChangeRequest && !auth.passwordChangeRequest.success)`
Maybe you can create a dedicated method with an eloquent name and call it in the `componentWillReceiveProps`

ayyazdaniaryan: Sure. Will look into making it more clear.

assembl/static2/js/app/components/login/sendPwdForm.jsx
```diff
@@ -16,11 +16,10 @@ class SendPwdForm extends React.Component {
   }
   componentWillReceiveProps(nextProps) {
     const { auth } = nextProps;
-    if (auth.passwordChangeRequest && auth.passwordChangeRequest.success === false) {
-      let msg;
-      if (auth.passwordChangeRequest.data === 404) {
-        msg = I18n.t('login.emailNotFound');
-      } else { msg = I18n.t('login.passwordChangeRequestError'); }
+    const resp = auth.passwordChangeRequest;
+    if (resp && resp.success === false) {
```
swergas: Shouldn't there also be an "else if (!resp)" ?

assembl/static2/js/app/utils/httpRequestHandler.js
```diff
@@ -43,22 +43,15 @@ export const xmlHttpRequest = (obj) => {
...
-        try {
-          resp = JSON.parse(xhr.responseText);
-          reject(resp || xhr.statusText);
-        } catch (e) {
-          // A non-JSONError response, likely from Pyramid itself
-          // Short term solution until contract established
-          reject(xhr.status);
-        }
+        // Contract agreed upon. If API is to fail, must respond with
+        // JSONError type. Front-end respects this type of response only.
```
swergas: What if backend code has a failure which does not even results in the production of a JSONError? (but a 500 something)

ayyazdaniaryan: Then a 500 error will be thrown, with a content-type of `application/json` with the reason why. The error is supposed to span the 400-500 ranges. Correct me if I am wrong @maparent

assembl/static2/js/app/components/login/sendPwdForm.jsx
ayyazdaniaryan: Ahh, as in there is an absence of a `resp` object? If so, then none of the code below should be run. This is a possible case, as an action can be triggered. However, will a user ever see this case? 99% no, because these actions occur in different components and cannot be triggered in another other way.

assembl/static2/js/app/utils/httpRequestHandler.js
maparent: Not all our errors will be json at first, and I think that json errors will be most useful in the 400 range. 500 range is much less predictable by nature.

ayyazdaniaryan: Hmm, then perhaps status check should be for the 400 range, and in the 500s, the `JSONParse` be in a `try/catch` ? Maintains safety in the front-end whilst giving flexibility on 500 level errors?

maparent: I would not check status range at all. Use the mimetype to know if it's a jsonError, now that the mimetype is correct.

ayyazdaniaryan: Duuh, of course. I'll make that change.

PR #97, opened 2017-05-04
Production perf

ayyazdaniaryan:
Significantly improve our production bundle of react by using babel plugins that convert stateless react components into functional components, and also hoists components so that components are not redefined throughout the bundle.

Merging this *WILL* break your development environment. In order to bring your development environment into a stable state, do the following after merge:

```
assembl-ini-files local.ini
supervisorctl reload  #if supervisord is already running
```


assembl/static2/js/app/components/login/assemblLogin.jsx
```diff
@@ -8,11 +8,15 @@ import { get, getFullPath, getContextual } from '../../utils/routeMap';
...
+    const next = this.props.next ?
+      this.props.next
+      : (
+        slug ? get('home', { slug: slug }) : null
+      );
     const errorMessage = ('error_message' in this.props && this.props.error_message) ? this.props.error_message : null;
     return (
       <div>
-        <form className="login" method="POST" action={getFullPath('ctxOldLogin', { slug: slug })}>
+        <form className="login" method="POST" action={slug ? getFullPath('ctxOldLogin', { slug: slug }) : getFullPath('oldLogin')}>
```
swergas: I saw you, incorporating my develop commits into your feature PR :p

doc/developper_tasks.rst
```diff
@@ -67,4 +67,17 @@ Raw sql connection
 .. code:: sh
 
     psql -U assembl -h localhost assembl
-    
+
+Updating supervisord.conf file
+------------------------------
+
+If you require to update, improve or otherwise add more commands for supervisor to manage, you must affect change in the tempate
```
vincentfretin: tempate -> template

assembl/static2/webpack.config.js
```diff
@@ -2,57 +2,57 @@ var path = require('path');
...
-            NODE_ENV: JSON.stringify('production')
-          }
-        }),
-        new webpack.optimize.UglifyJsPlugin(),
+  entry: {
+    bundle: ['babel-polyfill', './js/app/index'],
+    searchv1: ['babel-polyfill', './js/app/searchv1']
+  },
+  output: {
+    path: path.join(__dirname, 'build'),
```
swergas: Wat is the real diff in this file? I cannot see the difference... is it just spacing?

vincentfretin:
Is the BABEL_ENV really used here? NODE_ENV is not automatically used?

If you want to go further, it is now recommanded to use
https://github.com/babel/babel-preset-env
instead of babel-preset-es2015

doc/developper_tasks.rst
ayyazdaniaryan: It's not my PR if it doesn't have a spelling mistake somewhere 😂

ayyazdaniaryan:
NODE_ENV is not propagated down to babel level. I checked. I like the idea that you're proposing, actually. I think it's a bigger task than the intentions of this PR, however. This seems to be the set of babel plugins needed to compile your JS down to the targeted browsers. @paolina92 has used the es6/react presets, which might bloat our babel process a big more, but do not affect production in any way. The bundle remains unchanged, as I understand it. But, I will keep this in mind, and we can make a decision to move towards targeted browser presets.

assembl/static2/package.json
```diff
@@ -9,8 +9,8 @@
   "scripts": {
     "test": "jest",
     "test:watch": "npm test -- --watch",
-    "dev": "webpack -d --watch",
-    "build": "webpack -p",
+    "dev": "BABEL_ENV=dev webpack -d --watch",
```
swergas: What is the purpose of defining BABEL_ENV ?

ayyazdaniaryan: The new `.babelrc` file is now segregated by env. It must be set somewhere

ayyazdaniaryan: It's done for production and manual development runs here. It's also done in `supervisord.conf` as well.

assembl/static2/webpack.config.js
ayyazdaniaryan: Yup. They were in a tab of 4 spaces. Our JS standard is 2 spaces per tab.

assembl/static2/js/app/components/login/assemblLogin.jsx
ayyazdaniaryan: Haha yeah. It was a very strange case of rebasing error or something. Anyways, merged your changes right back in there. On a merge, your changes should remain intact.

assembl/static2/package.json
vincentfretin: I thought -d for webpack meant NODE_ENV=development and -p means NODE_ENV=production. 
This is not the case at all
```  -d                                    shortcut for --debug --devtool sourcemap --output-pathinfo
  -p                                    shortcut for --optimize-minimize

https://babeljs.io/docs/usage/babelrc/ says 

> The env key will be taken from process.env.BABEL_ENV, when this is not available then it uses process.env.NODE_ENV if even that is not available then it defaults to "development".

So I think you can simplify the configuration here, don't use BABEL_ENV, don't do the change in supervisorctl.tmpl and just modify the build command in package to be:
NODE_ENV=production webpack -p
This is actually what I do in another project.
NODE_ENV=production webpack --config webpack.config.prod.js
 So it seems you already have

vincentfretin: You need to use development instead of dev in .babelrc.

vincentfretin:
This PR is quite old now. I just tested again transform-react-constant-elements and transform-react-inline-elements. This is really not worth it, you reduce the bundle size by 335 bytes, ridiculous. :)
I close it.

PR #100, opened 2017-05-11
install yarn

vincentfretin:
Install yarn package manager for ubuntu/debian and mac.

### DIFF
--- a/fabfile.py
+++ b/fabfile.py
@@ -823,6 +823,7 @@ def install_single_server():
     execute(skeleton_env, None)
     execute(install_java)
     execute(install_elasticsearch)
+    execute(install_yarn)
     execute(install_database)
     execute(install_assembl_server_deps)
     execute(check_and_create_database_user)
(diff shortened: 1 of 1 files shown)


fabfile.py
```diff
@@ -1544,6 +1545,19 @@ def install_java():
 
 
 @task
+def install_yarn():
+    """Install yarn"""
+    if not env.mac:
+        if not exists('/etc/apt/sources.list.d/yarn.list'):
```
ayyazdaniaryan: All looks well. One question. If one of our production servers are on Ubuntu 14.04, will this step also be required? 
https://nodejs.org/en/download/package-manager/#debian-and-ubuntu-based-linux-distributions

vincentfretin: you already install an updated node version via nodeenv if you don't have a recent node version on the system, see update_node task in fabfile.py

vincentfretin: on my machine I installed node from nodesource repository yeah.

swergas:
Looks good to me too

PR #102, opened 2017-05-16
Navigation between v1 and v2

swergas:
Corresponds to story DEVAS-396

### DIFF
--- a/assembl/static2/js/app/actions/debateActions.js
+++ b/assembl/static2/js/app/actions/debateActions.js
@@ -33,7 +33,7 @@ export const fetchDebateData = (debateId) => {
       dispatch(resolvedFetchDebateData(debateData));
     }).catch((error) => {
       const firstError = error[0];
-      if (firstError && "status" in firstError && firstError.status === 401) {
+      if (firstError && 'status' in firstError && firstError.status === 401) {
         dispatch(unauthorizedDebateData);
       } else {
         dispatch(failedFetchDebateData(error));
--- a/assembl/static2/js/app/actions/debateActions.js
+++ b/assembl/static2/js/app/actions/debateActions.js
@@ -33,7 +33,7 @@ export const fetchDebateData = (debateId) => {
       dispatch(resolvedFetchDebateData(debateData));
     }).catch((error) => {
       const firstError = error[0];
-      if (firstError.status === 401) {
+      if (firstError && "status" in firstError && firstError.status === 401) {
         dispatch(unauthorizedDebateData);
       } else {
         dispatch(failedFetchDebateData(error));
--- a/assembl/static2/js/app/components/debate/navigation/timelineSegment.jsx
+++ b/assembl/static2/js/app/components/debate/navigation/timelineSegment.jsx
@@ -16,16 +16,6 @@ class TimelineSegment extends React.Component {
     const { debateData } = this.props.debate;
     const slug = { slug: debateData.slug };
 
-    if ( phaseIdentifier == "thread" ){
-      const body = <div><Translate value="debate.redirectToThreadPhase" /></div>;
-      displayModal(null, body, true, null, null, true);
-      const redirectionURL = get('oldDebate', slug);
-      setTimeout(function(){
-        location.href = redirectionURL;
-      }, 6000);
-      return;
-    }
-
     if (isStepCompleted || isCurrentPhase) {
       browserHistory.push(`${get('debate', slug)}?phase=${phaseIdentifier}`);
     } else {
--- a/assembl/static2/js/app/components/debate/navigation/timelineSegment.jsx
+++ b/assembl/static2/js/app/components/debate/navigation/timelineSegment.jsx
@@ -15,6 +15,17 @@ class TimelineSegment extends React.Component {
     const { isStepCompleted, isCurrentPhase, phaseIdentifier } = this.props;
     const { debateData } = this.props.debate;
     const slug = { slug: debateData.slug };
+
+    if ( phaseIdentifier == "thread" ){
+      const body = <div><Translate value="debate.redirectToThreadPhase" /></div>;
+      displayModal(null, body, true, null, null, true);
+      const redirectionURL = get('oldDebate', slug);
+      setTimeout(function(){
+        location.href = redirectionURL;
+      }, 6000);
+      return;
+    }
+
     if (isStepCompleted || isCurrentPhase) {
       browserHistory.push(`${get('debate', slug)}?phase=${phaseIdentifier}`);
     } else {
--- a/assembl/static2/js/app/components/debate/navigation/timelineSegment.jsx
+++ b/assembl/static2/js/app/components/debate/navigation/timelineSegment.jsx
@@ -1,10 +1,7 @@
 import React from 'react';
 import { browserHistory } from 'react-router';
-import { Translate, Localize } from 'react-redux-i18n';
 import { connect } from 'react-redux';
 import { get } from '../../../utils/routeMap';
-import { displayModal } from '../../../utils/utilityManager';
-import { getStartDatePhase, getPhaseName } from '../../../utils/timeline';
 
 class TimelineSegment extends React.Component {
   constructor(props) {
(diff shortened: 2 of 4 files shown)


assembl/static2/js/app/pages/debate.jsx
```diff
@@ -18,6 +21,7 @@ class Debate extends React.Component {
     this.showThumbnails = this.showThumbnails.bind(this);
     this.hideThumbnails = this.hideThumbnails.bind(this);
     this.displayThumbnails = this.displayThumbnails.bind(this);
+    this.redirectToV1Threads = this.redirectToV1Threads.bind(this);
```
paolina92: You should rename this method to `redirectToV1` and remove the `Threads` term because it's for all phases different than the "survey" phase, not only `thread` but `twoColums` or `tokenVote` for instance.

assembl/static2/js/app/pages/debate.jsx
```diff
@@ -30,11 +34,41 @@ class Debate extends React.Component {
   displayThumbnails() {
     this.setState({ isThumbnailsHidden: !this.state.isThumbnailsHidden });
   }
+  redirectToV1Threads() {
```
paolina92: This method should not be here, but at a higher level. Certainly in `main.jsx` because we need it in the debate section but also in the landing page.

assembl/static2/js/app/pages/debate.jsx
```diff
@@ -30,11 +34,41 @@ class Debate extends React.Component {
...
   render() {
+    const { debateData } = this.props.debate;
+    const { identifier } = this.props; // identifier of current phase being accessed
     const { loading, thematics } = this.props.data;
-    const { identifier, isNavbarHidden } = this.props;
+    const { isNavbarHidden } = this.props;
     const isParentRoute = !this.props.params.phase || false;
     const themeId = this.props.params.themeId || null;
+
+    if (identifier && !isPhaseStarted(debateData.timeline, identifier)) {
```
paolina92: This block should stay were it was before and can't be in a render method.
You should learn about react lifecycle. That's why you used a setTimeout I suppose, because you don't use the good lifecycle method.
https://facebook.github.io/react/docs/react-component.html

assembl/static2/js/app/utils/translations.js
```diff
@@ -167,7 +167,8 @@ const Translations = {
         goTo: 'Rendez-vous sur la phase %{currentPhaseName}.'
       },
       themes: 'Thèmes',
-      notStarted: 'La phase %{phaseName} n\'a pas encore commencé. Merci de revenir à partir du '
+      notStarted: 'La phase %{phaseName} n\'a pas encore commencé. Merci de revenir à partir du ',
+      redirectToThreadPhase: 'Vous allez être redirigé vers la phase %{phaseName} dans quelques secondes. Assembl est en cours d\'amélioration, veuillez nous excuser pour le changement d\'interface.'
```
paolina92: Rename this key because it's not only for the `thread`.

assembl/static2/js/app/utils/translations.js
```diff
@@ -354,7 +355,8 @@ const Translations = {
         goTo: 'Go to %{currentPhaseName} phase.'
       },
       themes: 'Themes',
-      notStarted: 'The %{phaseName} phase is not started. Please come back from the '
+      notStarted: 'The %{phaseName} phase is not started. Please come back from the ',
+      redirectToThreadPhase: 'You are going to be redirected to the %{phaseName} phase in a few seconds. Assembl is being improved, please excuse us for the user interface change.'
```
paolina92: Rename this key because it's not only for the `thread`.

assembl/static2/js/app/pages/debate.jsx
ayyazdaniaryan: Why not utils if it is supposed to be used in multiple places?

assembl/static2/js/app/pages/debate.jsx
ayyazdaniaryan: Hmm.. `isPhaseStarted` is not a promise. Therefore, my suggestion is to take this entire `if/else` block, and put it in the `componentDidMount`. This method is called after render. And it seems that there is no `setState` called in your block, so no re-render will be done. Safe.

ayyazdaniaryan: So you can completely avoid the `setTimeout(func, 100)`.

assembl/static2/js/app/actions/debateActions.js
```diff
@@ -33,7 +33,7 @@ export const fetchDebateData = (debateId) => {
       dispatch(resolvedFetchDebateData(debateData));
     }).catch((error) => {
       const firstError = error[0];
-      if (firstError.status === 401) {
+      if (firstError && 'status' in firstError && firstError.status === 401) {
```
ayyazdaniaryan: I think, based on the contract we have jsonError in backend, this `error` object will always exist. Otherwise this block of code will not be hit. But anyways, it's fine to keep it, I think.

assembl/static2/js/app/components/debate/navigation/timelineSegment.jsx
```diff
@@ -15,14 +12,9 @@ class TimelineSegment extends React.Component {
...
+
     if (isStepCompleted || isCurrentPhase) {
       browserHistory.push(`${get('debate', slug)}?phase=${phaseIdentifier}`);
-    } else {
-      const { locale } = this.props.i18n;
-      const startDate = getStartDatePhase(debateData.timeline, phaseIdentifier);
-      const phaseName = getPhaseName(debateData.timeline, phaseIdentifier, locale).toLowerCase();
-      const body = <div><Translate value="debate.notStarted" phaseName={phaseName} /><Localize value={startDate} dateFormat="date.format" /></div>;
-      displayModal(null, body, true, null, null, true);
     }
```
swergas: The behaviour you expect is not part of the specification as of today, but I would be happy to discuss specification and make it more precise if you like :). And from my understanding, when I go to the `/{slug}/debate` page and current phase is a "thread" phase, I should indeed see the modal.

assembl/static2/js/app/pages/debate.jsx
swergas: I have no information about expected behavior of other phases than phase 1 and 2.

assembl/static2/js/app/pages/debate.jsx
swergas: I do also think it would fit better in another file, but I could not find any file more suited for this, and have not enough knowledge yet about the whole architecture of the React frontend to come up with a more suited one. I'll trust your judgement about the best place :)

assembl/static2/js/app/pages/debate.jsx
swergas: I think I already tried the componentDidMount way and obtained the same warnings/errors from React. I would be happy to do a quick pair programming session with one of you about this :)

assembl/static2/js/app/utils/translations.js
swergas: Please validate this information with design team.

assembl/static2/js/app/actions/debateActions.js
swergas: I commited this because I encountered the bug while I was developing this feature.

assembl/static2/js/app/components/debate/navigation/timelineSegment.jsx
paolina92: when I read acceptance test, it's very clear. The PO expect to have the popin without redirection to `/debate`. I'm not sure that what you did will be approved by the UX and the PO. The screenshot shows an example of the popin in order to keep the same UI.

assembl/static2/js/app/pages/debate.jsx
paolina92: No, in the timeline, you can add many phases, so the specification is: "Each phase that has no the survey identifier => redirection to version 1"

assembl/static2/js/app/pages/debate.jsx
paolina92: @ayyazdaniaryan I prefer in `main.jsx` than in `utils` because it's a UI behavior. Do you have a better argument?

assembl/static2/js/app/pages/debate.jsx
paolina92: okay but did you read the documentation? I send you the link.

assembl/static2/js/app/utils/translations.js
paolina92: It's validated. So I think that `redirectToV1` is a better key name.

assembl/static2/js/app/pages/debate.jsx
ayyazdaniaryan: @paolina92 I see your point. Very well, if you're not going to put it in`utils` since it's not UI interactions, then sure, `main.jsx` it is. You can pass it as a prop to all children, that is the react way, no?

assembl/static2/js/app/actions/debateActions.js
ayyazdaniaryan: Oh? Do you know which API from backend caused a crash? Because APIs that relate to V2 should respond with a `JSONError` now, which would ensure that this object would always exist.

ayyazdaniaryan: So if it failed, perhaps there is a backend problem to look into.

PR #106, opened 2017-06-06
Composite ini

maparent:
This branch had many goals:
1. Clean up the local.ini mess, so the fabfile takes all its informations from .rc files and generates the local.ini file from it (vs using a mix of fab environment, rc files and ini files.). 

There is now a single source of truth: the .rc file. The .rc files are a cascade, so it is easy to define configuration as a delta, and changes to the basis of the cascade will be propagated to all instances.

This involves a migration step, to be done by hand.

2. Use docker-compose to break up the assembl docker image.

That step is partially complete: We have broken up many components, notably the database.
We have yet to finish dockerization of mail, piwik and sentry (sentry is still work in progress.)


maparent:
Non, car alors il n'irait pas sur Webpack.

assembl/scripts/pshell.py
```diff
@@ -0,0 +1,44 @@
...
+
+def main(argv=sys.argv, quiet=False):
+    command = PShellCommandA(argv, quiet)
+    try:
+        return command.run()
+    except:
+        import pdb; pdb.post_mortem()
+
+
+class PShellCommandA(PShellCommand):
```
ayyazdaniaryan: Document intent please. It's to read a config file. But uncertain whether it takes an rc file, ini file, etc.

assembl/scripts/pypsql.py
```diff
@@ -14,6 +15,7 @@ def main():
     parser.add_argument('--ask_password', '-P', action="store_true",
                         help='ask for database password',)
     parser.add_argument('--database', '-d', help="database")
+    parser.add_argument('--autocommit', action="store_true", help="autocommit")
```
ayyazdaniaryan: Not necessarily in scope of this PR, but okay. What does this do, in tandem?

doc/developper_tasks.rst
```diff
@@ -12,11 +12,11 @@ Download a copy of a remote database to develop locally
 
 .. code:: sh
 
-    fab env_name_of_remote_env database_download
-    fab env_dev database_restore
+    fab -c configs/config_of_remote_instance.rc database_download
```
ayyazdaniaryan: `configs/{config_of_remote_instance}.rc` for clarity.

doc/docker.rst
```diff
@@ -1,13 +1,38 @@
...
+Go to ``localhost:9000`` to create a token for the admin user (username in ``configs/docker.rc``, generated password in ``configs/docker_random.ini``). Make sure that token can administer projects (and organizations?). Add ``sentry_api_token = (the token)`` to your ``my_docker.rc``.
+
+Then stop the docker containers (``Ctrl-C`` in the docker-compose window, or ``docker-compose -f docker/build/docker-compose-stage1.yaml down``), then:
+
+.. code:: sh
+
+    fab -c configs/my_docker.rc docker_compose
+    docker-compose -f docker/build/docker-compose.yaml up
+
+Your admin user for each assembl server will be ``admin@hostname``. Use the password recovery progress to set its initial password. (If you want to pshell instead, you can do it from ``docker exec -i -t build_assembl1_1 bash``.)
```
ayyazdaniaryan: Great info here.

docker/Dockerfile
```diff
@@ -1,55 +1,46 @@
-FROM    ubuntu:16.04
+#FROM    ubuntu:16.04
+FROM    debian:jessie
 
 ENV     DEBIAN_FRONTEND noninteractive
-ARG	GITREPO=https://github.com/assembl/assembl.git
-ARG	GITBRANCH=master
+ARG	GITREPO=assembl/assembl
+ARG	GITBRANCH=composite_ini
```
ayyazdaniaryan: TODO: Changeme on merge.

docker/Dockerfile
```diff
@@ -1,55 +1,46 @@
...
+            curl \
             uwsgi-plugin-python
-RUN     ssh-keygen -P '' -f ~/.ssh/id_rsa && \
-            cat ~/.ssh/id_rsa.pub >> ~/.ssh/authorized_keys
-RUN     cd /opt ; set -x ; git clone -b $GITBRANCH $GITREPO
-WORKDIR /opt/assembl
+RUN         useradd -m -U -G www-data assembl_user && \
+            ssh-keygen -P '' -f ~/.ssh/id_rsa && cat ~/.ssh/id_rsa.pub >> ~/.ssh/authorized_keys && \
+            sudo -u assembl_user -i sh -c "cd && mkdir .ssh && ssh-keygen -P '' -f .ssh/id_rsa && cat .ssh/id_rsa.pub >> .ssh/authorized_keys" && \
+            cat ~/.ssh/id_rsa.pub >> ~assembl_user/.ssh/authorized_keys
```
ayyazdaniaryan: Shouldn't it be `~/assembl_user/`?

docker/Dockerfile
```diff
@@ -1,55 +1,46 @@
...
-           sed -i 's/accept_secure_connection = true/accept_secure_connection = false/' local.ini && \
-           fab devenv install_single_server && \
-           cp docker/nginx_config_docker/default /etc/nginx/sites-enabled/ && \
-           /etc/init.d/ssh stop
-RUN     /etc/init.d/ssh start && \
-           ssh-keyscan localhost && \
-           fab devenv  build_virtualenv && \
-           fab devenv update_pip_requirements && \
+           curl -o fabfile.py https://raw.githubusercontent.com/$GITREPO/$GITBRANCH/fabfile.py && \
+           touch empty.rc && \
```
ayyazdaniaryan: lol

fabfile.py
```diff
@@ -10,65 +10,120 @@
...
+        # We want fname to be usable both on host and target.
+        # Use project-path-relative names to that effect.
+        if fname.startswith('~/'):
+            path = dirname(__file__)
+            if env.host_string != 'localhost':
+                path = env.get('projectpath', path)
+            fname = join(path, fname[2:])
+        else:
+            fname = join(dirname(rc_filename), fname)
+        service_config = combine_rc(fname, service_config)
```
ayyazdaniaryan: You couldn't help yourself, could you?

fabfile.py
```diff
@@ -10,65 +10,120 @@
...
+    if overlay is not None:
+        service_config.update(overlay)
+    service_config.pop('_extends', None)
+    service_config.pop('', None)
```
ayyazdaniaryan: Why is there an empty key?

fabfile.py
```diff
@@ -10,65 +10,120 @@
...
+    return service_config
 
 
-load_service_configs()
+def filter_global_names(rc_data):
+    """Some keys in rc files are prefixed with * or _ for ini conversion purposes"""
```
ayyazdaniaryan: Add: Returns the list of names in rc file that are not global nor private. Just keys. Unless the value of the key is `__delete_key__``

fabfile.py
```diff
@@ -10,65 +10,120 @@
...
+    return {k.lstrip('*').lstrip('_'): v for (k, v) in rc_data.iteritems()
+            if v != '__delete_key__'}
+
+
+def as_bool(b):
```
ayyazdaniaryan: This function already exists in Pyramid somewhere....

fabfile.py
```diff
@@ -10,65 +10,120 @@
...
+    if not rc_file:
+        abort("You must specify a .rc file")
+    if not os.path.exists(env.rcfile):
+        abort("This .rc file does not exist locally: " + rc_file)
+    env.update(filter_global_names(env))
+    env.update(filter_global_names(combine_rc(rc_file)))
+    sanitize_env()
+
+
+def task(func):
```
ayyazdaniaryan: Smart. Can you further elaborate the use of `@wraps`? What is its purpose when you are adding a wrapper yourself?

fabfile.py
```diff
@@ -10,65 +10,120 @@
...
+    """Some keys in rc files are prefixed with * or _ for ini conversion purposes"""
+    return {k.lstrip('*').lstrip('_'): v for (k, v) in rc_data.iteritems()
+            if v != '__delete_key__'}
+
+
+def as_bool(b):
+    return str(b).lower() in {"1", "true", "yes", "t", "on"}
+
+
+def sanitize_env():
```
ayyazdaniaryan: Is this reading from rc files now? Where in the lifecycle is this called?

fabfile.py
```diff
@@ -93,27 +148,127 @@ def listdir(path):
...
+            # create the local.ini in a temp file
+            venvcmd("python assembl/scripts/ini_files.py compose -o %s -r %s %s" % (
+                local_file_name, random_file_name, env.rcfile))
+            # send the random file if changed
+            if rt != os.path.getmtime(random_file_name):
+                put(random_file_name, random_ini_path)
+            # send the local file
+            put(local_file_name, local_ini_path)
+        finally:
+            os.unlink(random_file_name)
```
ayyazdaniaryan: Why?

fabfile.py
```diff
@@ -93,27 +148,127 @@ def listdir(path):
...
+                random_ini_path, random_ini_path, local_ini_path))
+        venvcmd("python assembl/scripts/ini_files.py migrate -o %s %s " % (
+            dest_path, env.rcfile))
+    else:
+        # OK, this is horrid because I need the local venv.
+        local_venv = env.get("local_venv", "./venv")
+        assert os.path.exists(local_venv + "/bin/python"),\
+            "No usable local venv"
+        # get placeholder filenames
+        with NamedTemporaryFile(delete=False) as f:
```
ayyazdaniaryan: Does `f` get namespaced?

fabfile.py
```diff
@@ -298,10 +441,10 @@ def as_venvcmd(cmd, chdir=False):
     return cmd
 
 
-def venvcmd(cmd, shell=True, user=None, pty=False, chdir=True):
+def venvcmd(cmd, chdir=True, user=None, pty=False, **kwargs):
```
ayyazdaniaryan: Doesn't break old API for our fab usage?

fabfile.py
```diff
@@ -847,12 +961,38 @@ def install_assembl_deps():
...
+        sudo("apt-get install python-certbot-nginx")
+
+
+@task
+def generate_certificate():
+    """Generate a certificate for https, and add renewal to crontab"""
+    hostname = env.public_hostname
+    if not exists('/etc/letsencrypt/live/%s/fullchain.pem' % (hostname)):
+        sudo("certbot certonly --webroot -w /var/www/html -d " + hostname)
+    sudo("echo '12 3 * * 3 letsencrypt renew' | uniq | crontab")
```
ayyazdaniaryan: Why pipe to `uniq`?

fabfile.py
```diff
@@ -847,12 +961,38 @@ def install_assembl_deps():
...
+        if 'jessie' in release_data:
+            sudo("echo 'deb http://ftp.debian.org/debian jessie-backports main' >> /etc/apt/sources.list")
+            sudo("apt-get update")
+        elif 'ubuntu' in release_data:
+            sudo("apt-get install software-properties-common")
+            sudo("add-apt-repository ppa:certbot/certbot")
+            sudo("apt-get update")
+        else:
+            raise NotImplementedError("Unknown distribution")
+        sudo("apt-get install python-certbot-nginx")
```
ayyazdaniaryan: In newer version of Ubuntu (16.01) and later, it's `apt-get install letsencrypt`. Still want to use the `certbot` if we update ubuntu versions, say to version `17`?

fabfile.py
```diff
@@ -1110,70 +1235,106 @@ def check_and_create_sentry_database_user():
...
+            "name": env.public_hostname,
+            "slug": slug})
+        assert r
+    key_url = "{base}projects/{organization}/{slug}/keys/".format(
+        base=base, organization=organization, slug=slug)
+    r = requests.get(key_url, headers=headers)
+    assert r
+    keys = r.json()
+    assert len(keys), "No key defined?"
+    default = [k for k in keys if k["label"] == "Default"]
```
ayyazdaniaryan: Fragile. This could change on new versions of sentry...

fabfile.py
```diff
@@ -1110,70 +1235,106 @@ def check_and_create_sentry_database_user():
...
-        execute(database_create_postgres)
-
+        key = keys[0]
+    # This should ideally go in the .rc file, but fab should not write rc files.
+    # putting it in the local random file for now.
+    parser = RawConfigParser()
+    parser.optionxform = str
+    if os.path.exists(env.random_file):
+        parser.read(env.random_file)
+    parser.set("DEFAULT", "sentry_key", key["public"])
```
ayyazdaniaryan: ```default_section = "DEFAULT"``` DRY principle

fabfile.py
```diff
@@ -1110,70 +1235,106 @@ def check_and_create_sentry_database_user():
...
+            password=env.db_password, host=env.db_host, user=env.db_user))
+        return not checkDatabase.failed
 
-def virtuoso_db_directory():
-    return join(env.projectpath, 'var/db')
 
+def check_if_db_tables_exist():
+    with settings(warn_only=True):
+        checkDatabase = venvcmd('assembl-pypsql -1 -u {user} -p {password} -n {host} -d {database} "{command}"'.format(
+            command="SELECT count(*) from permission", database=env.db_database,
```
ayyazdaniaryan: Any table other than `permission` might have done?

assembl/alembic/env.py
```diff
@@ -23,8 +24,9 @@
 set_config(config.file_config._sections['app:assembl'])
 pyramid_env = bootstrap(config.config_file_name)
 configure_zmq(pyramid_env['registry'].settings['changes.socket'], False)
-
 configure_engine(pyramid_env['registry'].settings, False)
+configure_indexing(pyramid_env['registry'].settings)
```
maparent: Are you sure it does nothing? Then let's remove the parameter, I say.

assembl/auth/make_saml.py
```diff
@@ -0,0 +1,109 @@
...
+from cryptography.hazmat.backends import default_backend
+from cryptography.hazmat.primitives import serialization
+from cryptography.hazmat.primitives.asymmetric import rsa
+from cryptography import x509
+from cryptography.x509.oid import NameOID
+from cryptography.hazmat.primitives import hashes
+try:
+    from cryptography.x509 import random_serial_number
+except ImportError:
+    # Old cryptography version
```
maparent: cryptography library version has not been updated in a while.

assembl/scripts/ini_files.py
```diff
@@ -144,72 +166,392 @@ def main():
...
+    return saml_info
+
+
+def diff_ini(first, second, diff=None, existing_only=False):
+    """Diff ini files
+
+    Generate a parser with any value in the second that is different in the first.
+    Returns a ConfigParser.
+    Takes interpolation into account. Does not include values that disappeared."""
+    from ConfigParser import _Chainmap
```
maparent: It's ConfigParser in python2, configparser in python 3. Which docs are your reading?

ayyazdaniaryan: Clearly, wrong docs....

fabfile.py
maparent: Yes, but fabfile cannot count on pyramid being installed, at least initially.

assembl/alembic/env.py
maparent: done.

assembl/auth/make_saml.py
```diff
@@ -0,0 +1,109 @@
...
+
+def private_key_from_cleaned_text(key_text):
+    key_text = '\n'.join([l.strip() for l in
+                          str(key_text).strip('\n').split('\n')])
+    key_text = "-----BEGIN RSA PRIVATE KEY-----\n\n%s\n-----END RSA PRIVATE KEY-----\n" % key_text
+    return serialization.load_pem_private_key(
+        key_text, password=None, backend=default_backend())
+
+
+def make_saml_cert(key, country='', state='', locality='', org='', cn='',
```
maparent: Correct.

assembl/auth/make_saml.py
```diff
@@ -0,0 +1,109 @@
...
+            x509.SubjectAlternativeName([
+                x509.DNSName(n) for n in altnames]),
+            critical=False)
+    crt = crt.sign(key, hashes.SHA256(), default_backend())
+    crt_text = crt.public_bytes(serialization.Encoding.PEM)
+    return (cleanup_x509_text(crt_text), crt)
+
+
+if __name__ == '__main__':
+    saml_info = {"country": "FR", "state": "Hauts-de-Seine",
```
maparent: __main__ is just for testing.

assembl/auth/make_saml.py
```diff
@@ -0,0 +1,109 @@
...
+                 "locality": "Levallois-Perret", "org": "Bluenove",
+                 "cn": "assembl-enterprise.bluenove.com",
```
maparent: again, this code will not be exercized IRL. I have built this information into bluenove.rc, in the private config repository, because that applies to all bluenove servers.

assembl/indexing/changes.py
```diff
@@ -193,4 +194,19 @@ def tpc_abort(self, transaction):
         self._clear()
 
 
-changes = ElasticChanges(transaction.manager)
+_changes = None
+
+
+def get_changes():
+    global _changes
```
maparent: It is invoked downstream, and it is initialized too late in process (dependent on config info) to be made a module variable.

assembl/scripts/ini_files.py
```diff
@@ -144,72 +166,392 @@ def main():
...
+    overlay = asParser(overlay)
+    for section in overlay.sections():
+        # Avoid including DEFAULTSECT
+        for key in overlay._sections[section]:
+            value = overlay.get(section, key)
+            if adding or config.has_option(section, key):
+                ensureSection(config, section)
+                config.set(section, key, value)
+    if overlay._defaults:
+        config._defaults.update(overlay._defaults)
```
maparent: Eh... the pythonic thing is to use private vars responsibly, being aware it is more prone to break. The responsible adult thing might be to add a test.... but since python2 is EOL, and python3 even seems to use the same var name, I'm not too worried.

assembl/scripts/ini_files.py
```diff
@@ -144,72 +166,392 @@ def main():
...
+
+def migrate_beaker_config(random_ini, overlay):
+    """Migrate old-style session... to beaker.session..."""
+    for section in random_ini.sections():
+        # Avoid including DEFAULTSECT
+        if not overlay.has_section(section):
+            continue
+        o_section = overlay._sections[section]
+        for key in random_ini._sections[section]:
+            if key.startswith("beaker."):
```
maparent: This is hackish, yes. But it's about removing legacy complexity in a migration, not the main code.

assembl/scripts/ini_files.py
```diff
@@ -144,72 +166,392 @@ def main():
...
+                alt_key = 'beaker.' + key
+                if alt_key in o_section and key not in o_section:
+                    overlay.set(section, key, overlay.get(section, alt_key))
+    return overlay
+
+
+def dump(ini_file):
+    """Dump the ini file, showing interpolations and errors."""
+    ini_file = asParser(ini_file, SafeConfigParser)
+    ini_file._defaults['here'] = os.getcwd()
```
maparent: comes from beaker, not sure how/where needs to change.

assembl/scripts/ini_files.py
```diff
@@ -144,72 +166,392 @@ def main():
...
+    for section in ini_file.sections():
+        print "\n[%s]\n" % section
+        for option in ini_file._sections[section]:
+            print option, '=',
+            try:
+                print ini_file.get(section, option)
+            except Exception as e:
+                print ini_file.get(section, option, True)
+                print "*"*15, "Could not interpolate"
+                print e
```
maparent: print goes to stdout (unless you mess with sys.stdout.)
logs are logs. This subcommand is meant to only be used on CLI, not through fabric.

assembl/scripts/ini_files.py
```diff
@@ -144,72 +166,392 @@ def main():
...
+                prefix = keyu[:-12]
+                if value == "{saml_key}":
+                    saml_key_text, saml_key = make_saml_key()
+                    base.set(section, key, saml_key_text)
+                    saml_keys[prefix] = saml_key
+                    changed = True
+                else:
+                    saml_keys[prefix] = value
+            elif value.startswith('{random') and value.endswith("}"):
+                value = b64encode(urandom(int(value[7:-1])))
```
maparent: Added the size check. I think regexp is overkill when startswith/endswith is enough.

assembl/scripts/ini_files.py
```diff
@@ -144,72 +166,392 @@ def main():
...
+        if val == '__delete_key__':
+            # Allow to remove a variable from rc
+            # so we can fall back to underlying ini
+            p.remove_option(section, key)
+        else:
+            p.set(section, key, val)
+    return p
+
+
+def iniconfig_to_rc(parser, dest=None, extends=None, target_dir=None):
```
maparent: well... do you want to?

assembl/scripts/ini_files.py
```diff
@@ -144,72 +166,392 @@ def main():
...
+            random_data.write(f)
+    else:
+        random_file = None
+    base = compose(rc_filename, random_file)
+    diff = diff_ini(base, expected_ini)
+    return iniconfig_to_rc(diff, extends=rc_filename, target_dir=target_dir)
+
+
+def main():
+    parser = ArgumentParser(prog='ini_files.py')
```
maparent: good idea. I was guarding against the default of prog[0], which could be assembl_ini_files

assembl/scripts/pypsql.py
maparent: actually used in fabfile. Very much suggested by an error message otherwise.

docker/Dockerfile
maparent: ~username means the home of that user.

docker/Dockerfile
maparent: eh.

fabfile.py
maparent: I could not find a better way, really.

fabfile.py
maparent: I forgot, honestly. empty lines maybe?

fabfile.py
maparent: here and there, look for as_bool in the fabfile

fabfile.py
maparent: why clean up? more precise question please

fabfile.py
maparent: it's a variable name like any other.

fabfile.py
maparent: Yes, but second argument hardly used and I checked usage.

fabfile.py
maparent: good point. Care to correct it?

fabfile.py
maparent: so the command is idempotent, otherwise we'd get that line repeated.

fabfile.py
maparent: Yes, any API can break.

fabfile.py
maparent: pretty much.

fabfile.py
ayyazdaniaryan: AH, `unlink` is same as remove. You're just deleting a file. Nothing to see here then...

fabfile.py
ayyazdaniaryan: Cool, just checking.

fabfile.py
ayyazdaniaryan: Sure, I'll take that on.

fabfile.py
maparent: wraps ensures that the wrapper keeps the same name and docstring as the original

fabfile.py
maparent: they pollute the namespace, no?

assembl/scripts/ini_files.py
maparent: Reason I won't bother: When we go python3, we'll go fab2 and a lot of this will have to be refactored.

fabfile.py
```diff
@@ -1528,18 +1519,16 @@ def get_vendor_config():
...
 
-@task
 def update_vendor_themes(frontend_version=1):
-    """Update optional themes in assembl/static/css/themes/vendor"""
     sanitize_env()
-    config = get_vendor_config()
-    config_section_name = 'theme2_repositories' if frontend_version == 2 else 'theme_repositories'
-    if config.has_section(config_section_name):
+    assert frontend_version in (1, 2)
+    theme_varname = "theme%d_repositories__git-urls" % frontend_version
```
swergas: I don't see yet this variable added in a config file (I'm reading code commit per commit)

configs/base_env.rc
```diff
@@ -0,0 +1,76 @@
...
+supervisor__autostart_celery_notify = true
+supervisor__autostart_celery_notify_beat = true
+supervisor__autostart_celery_translate = false
+supervisor__autostart_source_reader = true
+supervisor__autostart_changes_router = true
+supervisor__autostart_pserve = false
+supervisor__autostart_nodesass = false
+supervisor__autostart_gulp = false
+supervisor__autostart_webpack = false
+supervisor__autostart_uwsgi = false
```
swergas: Why not supervisor__autostart_uwsgi = true ?

fabfile.py
```diff
@@ -127,6 +129,43 @@ def listdir(path):
...
+    random_ini_path = os.path.join(env.projectpath, 'random.ini')
+    local_ini_path = os.path.join(env.projectpath, 'local.ini')
+    if exists(local_ini_path):
+        run('cp %s %s.%d' % (
+            local_ini_path, local_ini_path, int(time())))
+
+    if env.host_string == 'localhost':
+        # The easy case
+        venv("python assembl/scripts/ini_files.py compose -o local.ini " + env.rcfile)
+    else:
```
swergas: I don't understand the difference of treatments between localhost case and remote server case

fabfile.py
```diff
@@ -166,6 +166,58 @@ def create_local_ini():
 
 
 @task
+def migrate_local_ini():
```
swergas: Is this task used only for development of this .rc system? Or is it meant to be used in a regular pipeline?

docker/Dockerfile
swergas: yes!

configs/base_env.rc
maparent: Good question. All prod servers have autostart_uwsgi set to false, and it's not actually the best idea. SO changing it, but it may mean a bit more testing of the restart procedure (necessary in all cases.)

fabfile.py
maparent: It's not in a config file, it's an argument to the function. I kept the fabfile assumption that both always have to be updated.

fabfile.py
maparent: added comments.

fabfile.py
maparent: comment added.

docker/docker-compose.yaml
```diff
@@ -0,0 +1,67 @@
...
+  web:
+    image: nginx:1.12-alpine
+    ports:
+      - "8080:80"
+    volumes:
+     - ./nginx_config_docker/default:/etc/nginx/conf.d/assembl1.template
+     - ../var/docker/stack1/assembl1_run:/opt/assembl/var/run
+    environment:
+      ASSEMBL_HOST: assembl
+      PUBLIC_HOSTNAME: assemlbl1.local
```
swergas: ?

fabfile.py
```diff
@@ -1381,6 +1381,17 @@ def docker_startup():
...
+    copied = False
+    if not exists("/opt/assembl_static/static"):
+        run("cp -rp %s/assembl/static /opt/assembl_static/" % env.projectpath)
+        copied = True
+    if not exists("/opt/assembl_static/static2"):
+        run("cp -rp %s/assembl/static2 /opt/assembl_static/" % env.projectpath)
+        copied = True
+    if copied:
+        run("chmod a+r /opt/assembl_static/static")
+        run("find /opt/assembl_static -type d | xargs chmod a+x")
```
swergas: and you don't do the same for static2 ?

fabfile.py
```diff
@@ -1403,6 +1406,34 @@ def flushmemcache():
 
 
 @task
+def docker_compose():
```
swergas: Adding a comment about what this task does would be good.

configs/docker.rc
```diff
@@ -29,6 +29,12 @@ _uses_global_supervisor = false
 # What are the public hostnames of your assembl hosts? Can be multiple, separated by spaces
 docker_assembl_hosts = assembl1.local assembl2.local
 
+# This is where you give the path to your base docker.rc file,
+# presumably overriding this one.
+# If it is not part of git, you would probably store it in docker/build,
+# which is visible to the docker image as /opt/default_config
+docker_base_rc = ~/configs/docker.rc
```
swergas: Do you have an example use case of this.

assembl/scripts/ini_files.py
```diff
@@ -223,12 +223,10 @@ def populate_random(random_file, random_template=None, saml_info=None):
     from os import urandom
     from assembl.auth.make_saml import make_saml_key, make_saml_cert
     base = Parser()
-    if random_template is None:
-        random_template = os.path.basename(random_file) + ".tmpl"
```
swergas: Why not keep these 2 lines? except we change random_template to random_templates, like so:
```
if random_templates is None:
    random_templates = os.path.basename(random_file) + ".tmpl"
```

assembl/scripts/ini_files.py
```diff
@@ -198,6 +198,21 @@ def combine_ini(config, overlay, adding=True):
     return config
 
 
+def migrate_beaker_config(random_ini, overlay):
+    """Migrate old-style session... to beaker.session..."""
```
swergas: Could I have more context about beaker, what it is, how is it different to previous session system and why we change? I'm a bit lost here...

fabfile.py
```diff
@@ -1251,6 +1251,22 @@ def check_if_database_exists():
...
+            password=env.db_password, host=env.db_host, user=env.db_user))
+        return not checkDatabase.failed
+
+
+def check_if_first_user():
+    with settings(warn_only=True):
+        checkDatabase = venvcmd('assembl-pypsql -1 -u {user} -p {password} -n {host} -d {database} "{command}"'.format(
+            command="SELECT count(*) from public.user", database=env.db_database,
+            password=env.db_password, host=env.db_host, user=env.db_user))
+        return not checkDatabase.failed and int(checkDatabase.strip('()L,')) > 0
```
swergas: This does not seem to check if we are the first user (like I would understand from function name), but if there are some users.

assembl/alembic/env.py
swergas: Adding this line here means that elasticsearch will be configured at initialization of each migration, right? If yes, why does it need to?

assembl/indexing/changes.py
```diff
@@ -193,4 +194,19 @@ def tpc_abort(self, transaction):
...
+
+
+def configure_indexing(config):
+    global _changes
+    _changes = ElasticChanges(transaction.manager)
+
+
+def includeme(config):
+    """ Initialize changes. """
+    configure_indexing(config)
```
swergas: I would have the same question

docker/docker-compose.yaml.jinja2
```diff
@@ -55,18 +55,79 @@ services:
...
+    command: "sentry run cron"
+    environment:
+      SENTRY_POSTGRES_HOST: postgres
+      SENTRY_DB_USER: postgres
+      SENTRY_DB_PASSWORD: postgres
+      SENTRY_REDIS_HOST: redis
+      SENTRY_SINGLE_ORGANIZATION: "true"
+      SENTRY_SECRET_KEY: jran5_Hnoherknoudr65_Ryaboeeenh40_Bloboo
+
+  sentry_celery_worker:
```
swergas: Sentry needs celery tasks? Is it not part of the sentry container?

assembl/scripts/pshell.py
swergas: Has it been documented since?

fabfile.py
swergas: Maybe then you should document which versions of sentry are supported as of now

doc/configuration.rst
```diff
@@ -143,55 +164,55 @@ piwik_host:
...
-    TODO
+    The identifier of the sentry project of this server
 
 \*sentry_key:\
-    TODO
+    The public key of the sentry project of this server
 
 \*sentry_secret:\
-    TODO
+    The public key of the sentry project of this server
```
swergas: private key

doc/docker.rst
swergas: password recovery proCess

assembl/scripts/pshell.py
swergas: A bit in https://github.com/assembl/assembl/pull/106/commits/675360d3c0c5f0f8a5fcbe2ca8b2dafd0e4d8a3c

assembl/scripts/ini_files.py
```diff
@@ -144,72 +166,392 @@ def main():
...
+                                default=sys.stdout,
+                                help='The output file')
+    parser_combine.add_argument('input', type=FileType('r'), nargs='+',
+                                help='Input ini files')
+
+    # ini compose
+    parser_compose = subparsers.add_parser('compose', help=short_help(compose))
+    parser_compose.add_argument('--output', '-o', type=FileType('w'),
+                                default=sys.stdout, help='The output file')
+    parser_compose.add_argument('--random', '-r', help='random.ini file',
```
maparent: It's in the docs, yes.

docker/Dockerfile
maparent: done after merge

fabfile.py
maparent: added to function comment.

docker/docker-compose.yaml
maparent: corrected in later commits

fabfile.py
maparent: Look again: It's the folder enclosing both.

configs/docker.rc
maparent: No, when we install with clients we will have an example. I've been using the base docker files in my tests. But you made me realize something else.

assembl/scripts/ini_files.py
maparent: now that there are many templates, there is no explicit one. Explicit is better than implicit etc.

assembl/scripts/ini_files.py
maparent: Beaker takes care of our sessions, no change there. The config variable name can be prefixed by `beaker.` or not; the migration was about trying to have more explicit namespaces. (Some variables were harder to change and remained in the global namespace.)

fabfile.py
maparent: renamed the function.

docker/docker-compose.yaml.jinja2
maparent: taken from other example sentry docker-compose.yml files, such as https://gist.github.com/ebuildy/270f4ef3abd41e1490c1
While answering, found this which is more official:
https://github.com/getsentry/onpremise/blob/master/docker-compose.yml

PR #107, opened 2017-06-12
Fix hot reloading of css and js

vincentfretin:
I updated to react hot loader 3, that has better support for hot reloading function components.

The reason why hot reload didn't fully work until now is the publicPath in output, it had to point to webpack url, not pyramid url.

In the console, you will have the following warning "Warning: [react-router] You cannot change <Router routes>; it will be ignored". It's just a warning, not an error, you can ignore it, the components used in the routes are actually patched. If you add new routes, you will have to refresh the page yourself, that's it.

I removed the -d option from webpack-dev-server command, it implied --debug and sourcemap, for sourcemap. I now use devtool: 'eval'. I read eval is best for hot reload, otherwise the sourcemap is not fully update on the browser. The -d option was delaying the hot reload by 2s. Now it takes less than 1s as it should be.

If you are wondering, <AppContainer> is used to keep the local state between hot reload and is actually disabled in production, it just render the children.

ExtractTextPlugin is not recommended with hot reload, so in dev mode, I actually put back the default theme in the bundle.js and the css is injected via the js. I explicitly remove bundle.css and theme_default_web.css from the build folder when webpack start in dev so the css is not included twice, via the not updated css files and via the injected css from bundle.js.


vincentfretin:
I now serve the css bundles only for production. I removed the hack in webpack.dev.config.js to remove the css bundles. You don't have 404 for development anymore.

PR #108, opened 2017-06-12
Move title and description fields from thematic/question to idea

vincentfretin:


assembl/alembic/versions/b0a17a1c42cb_move_title_and_description_fields_from_.py
```diff
@@ -0,0 +1,74 @@
...
+            sa.Column('description_id', sa.Integer,
+                sa.ForeignKey('langstring.id')))
+
+    # Do stuff with the app's models here.
+    from assembl import models as m
+    db = m.get_session_maker()()
+    with transaction.manager:
+        db.execute("UPDATE idea SET title_id = (SELECT title_id FROM thematic WHERE id=idea.id)")
+        db.execute("UPDATE idea SET description_id = (SELECT description_id FROM thematic WHERE id=idea.id)")
+        db.execute("UPDATE idea SET title_id = (SELECT title_id FROM question WHERE id=idea.id)")
```
ayyazdaniaryan: question overrides idea titles?

ayyazdaniaryan: Oh, a question is a child, so it would not override, but fill in the blanks...

assembl/alembic/versions/b0a17a1c42cb_move_title_and_description_fields_from_.py
```diff
@@ -0,0 +1,74 @@
...
+        op.add_column(
+            'question',
+            sa.Column('title_id', sa.Integer,
+                sa.ForeignKey('langstring.id')))
+
+    # Do stuff with the app's models here.
+    from assembl import models as m
+    db = m.get_session_maker()()
+    with transaction.manager:
+        db.execute("UPDATE thematic SET title_id = (SELECT title_id FROM idea WHERE id=thematic.id AND sqla_type='thematic')")
```
ayyazdaniaryan: Since I've always done this in `sqlalchemy`, the column `type`, which is the `polymorphic identifier` is actually `sqla_type` ? If so, this is correct.

vincentfretin: yes, sqla_type is the polymorphic identifier

assembl/alembic/versions/b0a17a1c42cb_move_title_and_description_fields_from_.py
vincentfretin: question inherits from idea. We had a title field in question before, so now it uses the one in idea.

PR #109, opened 2017-06-14
graphql changes for phase 2 (WIP)

vincentfretin:
Work in progress, currently based on phase1_admin_section branch until it is merged to develop.


assembl/graphql/schema.py
```diff
@@ -336,6 +342,31 @@ def is_type_of(cls, root, context, info):
         # return isinstance(root, cls._meta.model)  # this was the original code
         return type(root) == cls._meta.model or type(root) == models.RootIdea
 
+    def resolve_title(self, args, context, info):
+        title = resolve_langstring(self.title, args.get('lang'))
+        # If the idea was created from the old api or v1 interface, we don't
+        # have title, return short_title.
+        if title is None:
+            return self.short_title
```
vincentfretin: I implemented this fallback so we can see the title of existing ideas.

assembl/graphql/schema.py
```diff
@@ -480,13 +511,17 @@ def resolve_order(self, args, context, info):
 class Query(graphene.ObjectType):
     node = Node.Field()
     posts = SQLAlchemyConnectionField(PostConnection, idea_id=graphene.ID())
-    ideas = SQLAlchemyConnectionField(Idea)
+    root_idea = graphene.Field(Idea)
+    ideas = graphene.List(Idea)
```
vincentfretin: ideas query return a list directly, not pagination, like what we have for thematics query.

assembl/graphql/schema.py
```diff
@@ -480,13 +511,17 @@ def resolve_order(self, args, context, info):
 class Query(graphene.ObjectType):
     node = Node.Field()
     posts = SQLAlchemyConnectionField(PostConnection, idea_id=graphene.ID())
-    ideas = SQLAlchemyConnectionField(Idea)
+    root_idea = graphene.Field(Idea)
```
vincentfretin: I implemented a root_idea query. It may be a good idea with the proposed UI to actually get the ideas for a given level when you actually need it. For the first level, do a `rootIdea { children { title } }`. For level 2, so the children ideas of a given idea `node(id:$ideaId) { children { title } }` and so on...

assembl/graphql/schema.py
vincentfretin: ideas query results contain all ideas, so in the browser you need to reconstruct a tree based on parentId (root idea has parentId None)

PR #103, opened 2017-05-31
Phase1 admin section

paolina92:

PR #113, opened 2017-06-22
Graphql loader clean

cedricmessiant:
and some other improvements


assembl/static2/webpack.dev.config.js
```diff
@@ -93,10 +93,15 @@ module.exports = {
             test: /\.eot(\?v=\d+\.\d+\.\d+)?$/,
             loader: 'file'
         },
+        {
+          test: /\.(graphql|gql)$/,
+          exclude: /node_modules/,
+          loader: 'graphql-tag/loader'
```
ayyazdaniaryan: Loader is already part of the apollo dependencies?

assembl/static2/js/app/pages/administration.jsx
```diff
@@ -79,4 +79,10 @@ const mapDispatchToProps = (dispatch) => {
   };
 };
 
-export default compose(graphql(ThematicsQuery), withLoadingIndicator, connect(null, mapDispatchToProps))(Administration);
\ No newline at end of file
+export default compose(
```
ayyazdaniaryan: Nice use of compose.

assembl/static2/js/app/components/debate/survey/question.jsx
```diff
@@ -78,19 +79,21 @@ class Question extends React.Component {
...
-        showSubmitButton: false,
-        remainingChars: maxChars
+    this.props
+      .mutate({ variables: { ideaId: questionId, body: body } })
+      .then(() => {
+        scrollToQuestion(true, index + 1);
+        displayAlert('success', I18n.t('debate.survey.postSuccess'));
+        refetchTheme();
+        this.setState({
+          postBody: '',
```
ayyazdaniaryan: You sure `""` is okay instead of `null`?

assembl/static2/js/app/components/debate/survey/question.jsx
```diff
@@ -140,7 +136,9 @@ class Question extends React.Component {
                   }}
                   value={this.state.postBody}
                   maxLength={300}
-                  ref={(t) => { this.txtarea = t; }}
+                  ref={(t) => {
+                    this.txtarea = t;
```
ayyazdaniaryan: This ref object, it is a DOM object yes? Or is a React simulated? I see that above it's used as `this.txtarea.props.maxLength`. Wasn't aware that a textbox had props property

assembl/static2/js/app/components/debate/survey/question.jsx
```diff
@@ -78,19 +79,21 @@ class Question extends React.Component {
...
-      this.setState({
-        postBody: '',
-        showSubmitButton: false,
-        remainingChars: maxChars
+    this.props
+      .mutate({ variables: { ideaId: questionId, body: body } })
+      .then(() => {
+        scrollToQuestion(true, index + 1);
+        displayAlert('success', I18n.t('debate.survey.postSuccess'));
+        refetchTheme();
```
ayyazdaniaryan: From server or from Apollo cache? Can't see that in this PR. Its likely from PR #103

assembl/static2/js/app/components/debate/survey/post.jsx
```diff
@@ -51,31 +53,35 @@ class Post extends React.Component {
...
-        disagree: sentiments.data.deleteSentiment.post.sentimentCounts.disagree,
-        mySentiment: sentiments.data.deleteSentiment.post.mySentiment
+    this.props
+      .deleteSentiment({ variables: { postId: id } })
+      .then((sentiments) => {
+        target.setAttribute('class', 'sentiment');
+        this.setState({
+          like: sentiments.data.deleteSentiment.post.sentimentCounts.like,
+          disagree: sentiments.data.deleteSentiment.post.sentimentCounts.disagree,
+          mySentiment: sentiments.data.deleteSentiment.post.mySentiment
```
ayyazdaniaryan: Oh ouch. @cedricmessiant What do you think, maybe we can make models out of these things, like sentiments, posts, etc etc, so that we can have greater control over the state of these objects when they are live. Of course, what I'm saying requires LOTs of boilerplate to keep models consistent with the Apollo state, future requests, etc, etc. What is your opinion on this?

assembl/static2/js/app/components/debate/survey/question.jsx
paolina92: From server, to refresh the view with the new post.

assembl/static2/js/app/components/debate/survey/question.jsx
paolina92: I want to empty the textarea after submitting

assembl/static2/js/app/components/debate/survey/post.jsx
paolina92: Honestly, it's a bad idea to use the local state to store this data. I knew the problem and I should refactor this part. @cedricmessiant : Now that the admin section is done, I have a better understanding on how to do that. Get data with a graphQL query and put it in the store. Right?

assembl/static2/js/app/components/common/withLoadingIndicator.jsx
```diff
@@ -2,12 +2,14 @@
...
-const withLoadingIndicator = (WrappedComponent) => {
-  return (props) => {
-    if (props.data.loading) {
-      return <Loader />;
-    }
-    return <WrappedComponent {...props} />;
+const withLoadingIndicator = (loaderProps = {}) => {
+  return (WrappedComponent) => {
+    return (props) => {
+      if (props.data.loading) {
```
vincentfretin: We should be able to specify it with flow.

assembl/static2/webpack.dev.config.js
vincentfretin: graphql-tag is a dependency of react-apollo

assembl/static2/js/app/components/debate/survey/post.jsx
vincentfretin: @paolina92 for this case, no, don't store in redux store, we should use the update function in the mutation. We first tried updateQuery if your remember but failed, see DEVAS-458

assembl/static2/js/app/components/debate/survey/question.jsx
cedricmessiant: you should take a look at the doc about refs : https://facebook.github.io/react/docs/refs-and-the-dom.html
You don't need these very often but sometimes it's very useful. It's a reference to a DOM element or an instance of a class component. So you can trigger an event on an element when another event occur on another element

assembl/static2/js/app/components/debate/survey/post.jsx
cedricmessiant: Yes, we should avoid the use of the redux store to store data that comes from apollo when we don't need to. 
Here we probably just need a `post` variable to lighten the code. We should also stay aware of code that might be use lots of time and refactor as soon as we are about to duplicate some lines of code.

PR #116, opened 2017-06-27
use a graphql query to get the contributions and participants counters

vincentfretin:
use a graphql query to get the contributions and participants counters on the landing page (DEVAS-658)


assembl/static2/js/app/client.js
```diff
@@ -1,10 +1,25 @@
...
+        {
+          kind: 'INTERFACE',
+          name: 'IdeaInterface',
+          possibleTypes: [{ name: 'Idea' }, { name: 'Thematic' }]
+        }
+      ]
+    }
+  }
+});
+
```
vincentfretin: It's documented here http://dev.apollodata.com/react/initialization.html#fragment-matcher
Apollo needs to know the schema, that Idea and Thematic ObjectTypes both implements IdeaInterface.

PR #118, opened 2017-06-30
Show specific unauthorized page when unauthorized user tries to access administration section

swergas:
This corresponds to story DEVAS-650

### DIFF
--- a/assembl/static2/js/app/pages/unauthorizedAdministration.jsx
+++ b/assembl/static2/js/app/pages/unauthorizedAdministration.jsx
@@ -4,24 +4,20 @@ import { Link } from 'react-router';
 import { get } from '../utils/routeMap';
 import { getDiscussionSlug } from '../utils/globalFunctions';
 
-class UnauthorizedAdministration extends React.Component {
-
-  render() {
-    const slug = getDiscussionSlug();
-    const routeParams = { slug: slug };
-    return (
-      <div className="unauthorized-administration">
-        <div>
-          <Translate value="unauthorizedAdministration.unauthorizedMessage" />
-        </div>
-        <div> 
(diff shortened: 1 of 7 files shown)


ayyazdaniaryan:
So okay, I wasn't very clear on what this task was. But I do not see the need for a React page to be created for Unauthorized Access and then link them back to Assembl.

In `assembl/templates/react_unauthorized.tmpl`, there is a template in which you can use to redirect the user to. It was made to be extended as more 400s and perhaps even 500 level errors were to be presented to the user from Pyramid itself, rather than React. You could have kept the `react_view` intact, and added slightly more logic, that if you are requesting a route that you do not have permission to, the `administration*` routes, then you could redirected to that template (and passed in your error code).

Nothing against this method, just so much extra work and hand-shaking between Pyramid and React in order to do something that Pyramid could have handled already, and the templates/setup was already in place. Otherwise, looks good.

swergas:
Yes I know, but this was a UI requirement of the story to show the header etc

assembl/static2/js/app/pages/unauthorizedAdministration.jsx
```diff
@@ -0,0 +1,27 @@
...
+          <Link className="button-link button-dark margin-l" href={`${get('home', routeParams)}`}>
+            <Translate value="unauthorizedAdministration.returnButton" />
+          </Link>
+        </div>
+      </div>
+    );
+  }
+}
+
+export default UnauthorizedAdministration;
```
paolina92: Use a stateless component

ayyazdaniaryan: I'll take care of that.

PR #121, opened 2017-07-06
Feat(DEVAS-477) TopPost creation on Thread phase

nicolaszil:
https://bluenove.atlassian.net/browse/DEVAS-477


assembl/static2/js/app/components/debate/thread/topPostForm.jsx
```diff
@@ -0,0 +1,223 @@
...
+  updateSubjectChars,
+  bodyRemainingChars,
+  updateBodyChars
+}) => {
+  const displayForm = (isActive) => {
+    return updateFormStatus(isActive);
+  };
+
+  const redirectToLogin = () => {
+    const isUserConnected = getConnectedUserId(); // TO DO put isUserConnected in the store
```
swergas: Is this TODO a nice to have for later or is it mandatory for reasonable code quality?

assembl/static2/js/app/components/debate/thread/topPostForm.jsx
```diff
@@ -0,0 +1,223 @@
...
+  subjectRemainingChars,
+  updateSubjectChars,
+  bodyRemainingChars,
+  updateBodyChars
+}) => {
+  const displayForm = (isActive) => {
+    return updateFormStatus(isActive);
+  };
+
+  const redirectToLogin = () => {
```
swergas: Bad function name because it "redirects" (displays a modal actually) only on condition (user logged out) and else does something completely different.

assembl/static2/js/app/reducers/postsReducer.js
```diff
@@ -11,4 +11,56 @@ export const postsById = (state = Map(), action) => {
...
+export const subjectRemainingChars = (state = 10000, action) => {
+  switch (action.type) {
+  case 'UPDATE_SUBJECT_REMAINING_CHARS':
+    return action.subjectRemainingChars;
+  default:
+    return state;
+  }
+};
+
+export const bodyRemainingChars = (state = 10000, action) => {
```
swergas: These action methods all look the same, isn't there a way to factorize some of this code?

assembl/static2/tests/unit/actions/postsActions.spec.js
```diff
@@ -0,0 +1,75 @@
...
+      const expected = {
+        isTopPostFormActive: true,
+        type: 'UPDATE_TOP_POST_FORM_STATUS'
+      };
+      const actual = updateTopPostFormStatus(true);
+      expect(actual).toEqual(expected);
+    });
+  });
+
+  describe('updateSubjectRemaingChars action', () => {
```
swergas: Are these tests really useful? Couldn't we come up with better tests? Maybe more like functional tests?

assembl/static2/tests/unit/reducers/postsReducer.spec.js
```diff
@@ -0,0 +1,128 @@
...
+        type: 'UPDATE_TOP_POST_BODY',
+        topPostBody: 'New body'
+      };
+      const actual = topPostBody(state, action);
+      const expected = 'New body';
+      expect(actual).toEqual(expected);
+    });
+  });
+
+  describe('topPostFormStatus reducer', () => {
```
swergas: Same here, I find these tests not really useful, they could be autogenerated or replaced by better ones.

assembl/static2/css/components/posts.scss
```diff
@@ -141,7 +141,7 @@
 
     .body {
         padding: 20px 0;
-        word-break: break-all;
+        word-wrap: break-word;
```
swergas: Maybe replace `word-wrap` occurrences here by `overflow-wrap` because it seems to be the real official correct name of the property according to https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap

assembl/static2/js/app/components/debate/thread/topPostForm.jsx
```diff
@@ -0,0 +1,217 @@
...
+    const isUserConnected = getConnectedUserId(); // TO DO put isUserConnected in the store
```
ayyazdaniaryan: Do not see the benefit of moving to store. The user's connectivity is passed from the python-rendered template to the front-end. It's in a `script` tag. It is globally available. And by design, cannot change unless page is re-rendered by backend, which point, react must re-init and set up store all over again.

assembl/static2/js/app/components/debate/thread/topPostForm.jsx
```diff
@@ -0,0 +1,217 @@
...
+    }
+  };
+
+  const handleInputFocus = () => {
+    return redirectToLogin();
+  };
+
+  const handleSubjectChange = (e) => {
+    const maxChars = TEXT_INPUT_MAX_LENGTH;
+    const remaining = maxChars - e.target.value.length;
```
ayyazdaniaryan: ```const value = e.target.value.length```
Less hits to the target object, which will fractionally improve the perf.

assembl/static2/js/app/components/debate/thread/topPostForm.jsx
```diff
@@ -0,0 +1,217 @@
...
+    updateSubjectChars(remaining);
+    updateSubject(e.target.value);
+  };
+
+  const handleBodyChange = (e) => {
+    const maxChars = TEXT_AREA_MAX_LENGTH;
+    const remaining = maxChars - e.target.value.length;
```
ayyazdaniaryan: Same comment as above.

assembl/static2/js/app/components/debate/thread/topPostForm.jsx
```diff
@@ -0,0 +1,217 @@
...
+              onFocus={handleInputFocus}
+              onChange={handleSubjectChange}
+            />
+            <div className="annotation margin-s">
+              <Translate
+                value="debate.remaining_x_characters"
+                nbCharacters={subjectRemainingChars < 10000 ? subjectRemainingChars : TEXT_INPUT_MAX_LENGTH}
+              />
+            </div>
+            <div className={isFormActive ? 'margin-m' : 'hidden'}>
```
ayyazdaniaryan: Is there a css animation to transition from `hidden` to `margin-m`? Fade out/slide down? Don't think so by looks of this. Let's see how this looks during QA. TODO.

assembl/static2/js/app/graphql/mutations/createPost.graphql
```diff
@@ -1,8 +1,9 @@
-mutation createPost($ideaId: ID!, $body: String!) {
-  createPost(ideaId:$ideaId, body: $body) {
+mutation createPost($ideaId: ID!, $subject: String, $body: String!) {
```
ayyazdaniaryan: Hmm, interesting to see if the Phase1 proposition posts will still work now....

assembl/static2/tests/unit/reducers/postsReducer.spec.js
ayyazdaniaryan: Brought this up with @cedricmessiant in previous PRs. His response was they are quick wins to test the actions/reducers of the redux train. They are cheap and easy to write and catch the very low hanging fruit on the front-end. Rigorous testing is done using Jest's snapshot feature.

ayyazdaniaryan: And of course, integration testing and QA.

assembl/static2/js/app/components/debate/thread/topPostForm.jsx
paolina92: True. I change the name to `inviteToLogin`

assembl/static2/js/app/reducers/postsReducer.js
paolina92: It`s 2 different values. I don't know how...

assembl/static2/tests/unit/actions/postsActions.spec.js
paolina92: The goal is to check if something is broken when we will change the action. It's useful. Functional tests are different.

assembl/static2/css/components/posts.scss
paolina92: Hum... Not sure. I tested and it failed...

assembl/static2/js/app/components/debate/thread/topPostForm.jsx
paolina92: The goal is to test the component. As is, it's difficult to test the component. It's not a "pure" component

assembl/static2/js/app/components/debate/thread/topPostForm.jsx
paolina92: Ok... I didn't know.

assembl/static2/js/app/components/debate/thread/topPostForm.jsx
paolina92: I changed.

assembl/static2/css/components/posts.scss
swergas: OK yes it seems that IE and Edge still need the old name http://caniuse.com/#search=overflow-wrap

assembl/static2/js/app/reducers/postsReducer.js
swergas: Something like this:
```
export const basicActionFactory = (initial_state = '', handled_action='', returned_prop='') => {
  return (state=initial_state, action) => {
    switch (action.type) {
      case handled_action:
        return action[returned_prop];
      default:
        return state;
    }
  };
};

export const topPostSubject = basicActionFactory('', 'UPDATE_TOP_POST_SUBJECT', 'topPostSubject');
export const topPostBody = basicActionFactory('', 'UPDATE_TOP_POST_BODY', 'topPostBody');
export const topPostFormStatus = basicActionFactory(false, 'UPDATE_TOP_POST_FORM_STATUS', 'isTopPostFormActive');
export const subjectRemainingChars = basicActionFactory(10000, 'UPDATE_SUBJECT_REMAINING_CHARS', 'subjectRemainingChars');
export const bodyRemainingChars = basicActionFactory(10000, 'UPDATE_BODY_REMAINING_CHARS', 'bodyRemainingChars');
```

assembl/static2/tests/unit/reducers/postsReducer.spec.js
swergas: OK then

PR #123, opened 2017-07-12
Answer to a post devas565

paolina92:


assembl/static2/js/app/components/debate/thread/answerForm.jsx
```diff
@@ -1,36 +1,102 @@
...
+    updateBody(e.target.value);
+  };
+
+  const resetForm = () => {
+    updateAnswerFormStatus(false);
+    updateBodyChars(TEXT_AREA_MAX_LENGTH);
+    updateBody('');
+  };
+
+  const inviteToLogin = () => {
```
swergas: inviteToLogin function is duplicated between answerForm.jsx and topPostForm.jsx. I guess it should be imported from a separate file (and optionnaly call it with configuration parameters).

assembl/static2/js/app/reducers/postsReducer.js
```diff
@@ -29,38 +38,58 @@ export const topPostBody = (state = '', action) => {
...
-  case 'UPDATE_BODY_REMAINING_CHARS':
-    return action.bodyRemainingChars;
+  case 'UPDATE_ANSWER_POST_FORM_STATUS':
+    return action.isAnswerPostFormActive;
   default:
     return state;
   }
 };
 
+export const answerPostBody = (state = '', action) => {
```
swergas: I would make the same comment as previous PR: https://github.com/assembl/assembl/pull/121#discussion_r126692963 . How about the solution I proposed? (a better name for basicActionFactory could be basicReducerFactory)

assembl/static2/js/app/components/debate/thread/answerForm.jsx
```diff
@@ -0,0 +1,136 @@
...
+    updateBody(e.target.value);
+  };
+
+  const resetForm = () => {
+    hideAnswerForm();
+    updateBodyChars(TEXT_AREA_MAX_LENGTH);
+    updateBody('');
+  };
+
+  const inviteToLogin = () => {
```
paolina92: I would like to dispatch an action and store if the user is connected. Then I should move this method in the `main.jsx`. We need to display this modal at several places in the app. What do you think?

assembl/static2/js/app/graphql/IdeaWithPosts.graphql
```diff
@@ -8,7 +8,7 @@ query IdeaWithPosts($lang: String!, $id: ID!) {
       posts(first: 20) {
         edges {
           node {
-            ... on Post {
+            ... on AssemblPost {
```
swergas: I tried to use an union or interface type instead (defined in graphql/schema.py) but I could not get it to work... Any idea @vincentfretin @maparent @cedricmessiant ?

assembl/static2/js/app/reducers/postsReducer.js
paolina92: I don't know how to write it... No idea. If you have an example...

swergas: See example I wrote there ;)

assembl/static2/js/app/graphql/IdeaWithPosts.graphql
vincentfretin: Why do want to change that? We don't have PropositionPost on an idea, only posts. Post object type can actually be a sqlalchemy AssemblPost, ImportedPost, Email... Maybe this confused you?

assembl/static2/js/app/components/debate/thread/answerForm.jsx
paolina92: Finally, I created a method in a higher level (`utils/utilityManager`) that we can call every where.

assembl/static2/js/app/reducers/postsReducer.js
paolina92: Ok I followed your example

assembl/graphql/schema.py
```diff
@@ -1021,29 +1135,46 @@ def mutate(root, args, context, info):
...
             subject_langstring = langstring_from_input_entries(subject_entries)
             body_langstring = langstring_from_input_entries(body_entries)
-            discussion = models.Discussion.get(discussion_id)
-            new_post = cls(
-                discussion=discussion,
-                subject=subject_langstring,
-                body=body_langstring,
-                creator_id=user_id
-            )
+            if in_reply_to_post: # Is there an unified way to do this?
```
cedricmessiant: @vincentfretin ?

vincentfretin: Yeah, just use a params dict to put parent conditionnaly in it, and create new post with cls(**params)

vincentfretin: Can't you just always use parent=in_reply_to_post here? If in_reply_to_post is None, it should keep the parent field to null.

assembl/static2/js/app/graphql/mutations/createPost.graphql
```diff
@@ -1,15 +1,16 @@
-mutation createPost($ideaId: ID!, $subject: String, $body: String!) {
-  createPost(ideaId:$ideaId, subject: $subject, body: $body) {
+mutation createPost($ideaId: ID!, $subject: String, $body: String!, $parentId: ID) {
+  createPost(ideaId:$ideaId, subject: $subject, body: $body, parentId: $parentId) {
     post {
-      ... on PropositionPost {
+      ... on AssemblPost {
```
vincentfretin: If you want to reuse this createPost mutation, you should actually keep the PropositionPost fragment and write the same for Post fragment (you should remove the AssemblPost objecttype in the backend code)

paolina92: ok. I'll do

assembl/static2/js/app/graphql/IdeaWithPosts.graphql
vincentfretin: same here, put back Post in the query or you won't get the ImportedPost instances.

vincentfretin:
@swergas @paolina92 I simplified all this in 1c3133388f6db8e5a305562ae4b30f0874aad2e5 and cc816e4e300e948a8d921c0c4e9e698cf51a7d8

PR #130, opened 2017-07-26
Blur active element in displayModal (DEVAS-680)

n0izn0iz:
Should we move this behaviour in the displayModal function?


vincentfretin:
LGTM

vincentfretin:
I think you can use `document.activeElement = null` too. You're right, it may be a good idea to put this in displayModal.

vincentfretin:
never mind my proposal, `document.activeElement = null` will just remove the outline, but the element will still have the focus, I think.

PR #132, opened 2017-07-31
Add piwik tracking to new frontend

swergas:
This corresponds to story DEVAS-655.
This triggers piwik tracking everytime the React application changes the browser URL.
Adding custom piwik actions types on specific user behavior will be for a next time.

### DIFF
--- a/assembl/static2/package.json
+++ b/assembl/static2/package.json
@@ -66,7 +66,7 @@
     "json-loader": "^0.5.4",
     "lodash": "^4.17.4",
     "moment": "^2.18.1",
-    "piwik-react-router": "^0.9.0",
+    "piwik-react-router": "assembl/piwik-react-router#skip_piwik_initialization",
     "prop-types": "^15.5.8",
     "react": "^15.4.1",
     "react-apollo": "^1.2.0",
--- a/assembl/static2/package.json
+++ b/assembl/static2/package.json
@@ -66,6 +66,7 @@
     "json-loader": "^0.5.4",
     "lodash": "^4.17.4",
     "moment": "^2.18.1",
+    "piwik-react-router": "^0.9.0",
     "prop-types": "^15.5.8",
     "react": "^15.4.1",
     "react-apollo": "^1.2.0",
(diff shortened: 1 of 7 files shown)


assembl/views/discussion/views.py
```diff
@@ -295,6 +296,7 @@ def react_view(request, required_permission=P_READ):
         get_route=get_route,
         **common_context
     )
+    context.update(common_context)
```
vincentfretin: You don't need this line, there is **common_context just before.

assembl/static2/package.json
```diff
@@ -66,6 +66,7 @@
     "json-loader": "^0.5.4",
     "lodash": "^4.17.4",
     "moment": "^2.18.1",
+    "piwik-react-router": "assembl/piwik-react-router#skip_piwik_initialization",
```
ayyazdaniaryan: Interesting choice. By doing this, we have eliminated part of the analytics lib. It was meant to be a general purpose tool, and it included code on setting up multiple analytics engines on a page, if you so wished. I understand the need (to connect this analytics engine to react-router), I wouldn't recommend extending our own library to support react-router anyway.

assembl/views/discussion/views.py
swergas: oh yes, thanks good catch

assembl/static2/package.json
swergas: Oh okay, I did not know that we wanted to make our code support several analytics tracking services. Is it really? I've only heard of piwik support until now. Are you sure it is part of the requirements, or are we going scope creep? (in local.ini, the variable is called `web_analytics_piwik_script`, there is no generic `web_analytics_script`)

Do you have an idea how we could plug a generic tracking system to listen to changes in URL which are made by the React application? I have the intuition that it would imply a much more complicated architecture, and re-implementing big parts of what this library does and which is really tight to React router (so using an external lib which is maintained helps us for this, in case there are future changes in how react router does things which would create a need for changing this code). All this for no real added value at the moment, as I don't see on our roadmap anything about other tracking tools.

vincentfretin:
you can merge

PR #133, opened 2017-08-01
Fix misaligned carets (with flex) (DEVAS-611)

n0izn0iz:
Should work on all browsers except IE<=9
Tested on firefox, safari and chrome on osx


vincentfretin:
LGTM

PR #134, opened 2017-08-03
Add motion.ai chatbot

n0izn0iz:


assembl/templates/index_react.jinja2
```diff
@@ -42,5 +42,17 @@
                 <script src="{{REACT_URL}}/build/theme_{{theme_name}}_web.js"></script>
             {% endif %}
         {% endblock %}
+        <link href="https://api.motion.ai/sdk/webchat.css" rel="stylesheet" type="text/css">
```
swergas: I guess the chatbot feature should load only on the debates where it is active, not on every debate, right? And the parameters (id, api key, etc) should be set per debate, or something like this?

PR #138, opened 2017-08-07
Features rich text / delete and edit post

vincentfretin:
Features rich text / delete and edit post, and infinite scroll for an idea's threads


assembl/static2/css/_assembl_base_styles.scss
```diff
@@ -10,16 +10,18 @@
 @import 'components/doughnut';
 @import 'components/dropdown';
 @import 'components/footer';
-@import 'components/loader';
 @import 'components/goUp';
+@import 'components/loader';
 @import 'components/navbar';
+@import 'components/overflowMenu';
```
vincentfretin: The overflow menu, the ellipsis "..." on a post to show for now the edit and delete actions.

swergas:
Let's respect our process and do a code review before merging ;)

assembl/static2/js/app/components/debate/thread/topPostForm.jsx
```diff
@@ -1,193 +1,148 @@
+// FIXME: cancel button
```
swergas: Did you forget to remove the fixme or to implement it? Or is the implementation of a cancel button out of scope of this story?

assembl/static2/css/common.scss
```diff
@@ -51,6 +51,10 @@ input.placeholder, textarea.placeholder {
   font-family: $font-family6;
 }
 
+.pre-wrap {
```
swergas: There is probably a better to do it in CSS/Sass.

assembl/static2/js/app/components/common/richTextEditor/index.jsx
```diff
@@ -0,0 +1,191 @@
...
+    const contentState = this.state.editorState.getCurrentContent();
+    if (!contentState.hasText()) {
+      if (contentState.getBlockMap().first().getType() !== 'unstyled') {
+        return true;
+      }
+    }
+    return false;
+  }
+
+  focusEditor = (): void => {
```
swergas: Out of curiosity, what is the underlying reason for this setTimeout? I had to do similar things for the table of ideas...

assembl/models/post.py
```diff
@@ -631,12 +631,19 @@ def generate_message_id(cls):
         onupdate='CASCADE'
     ), primary_key=True)
 
+    modification_date = Column(DateTime)
```
swergas: Why not use HistoryMixin? It provides tombstone date of the previous version, which corresponds to the last modification date.

assembl/static2/package.json
```diff
@@ -9,7 +9,7 @@
   "scripts": {
     "test": "jest",
     "test:watch": "npm test -- --watch",
-    "build": "webpack --progress --colors",
+    "build": "rm -rf node_modules/react-i18nify/node_modules;webpack --progress --colors",
```
swergas: does this re-download all react-i18nify dependencies everytime we update assembl's dependencies? (and have built it before)

assembl/static2/.flowconfig
```diff
@@ -0,0 +1,15 @@
...
+
+[include]
+
+[libs]
+
+[lints]
+
+[options]
+# map .graphql files to an empty object, see https://github.com/facebook/flow/issues/2852
+module.name_mapper='.*\(.graphql\)' -> 'empty/object'
```
vincentfretin: There seems to be a lot of dependency for that indeed, but really we don't care here, it's a dev dependency. This dependency is not included in the final bundle. An empty file is not enough, it must return an empty object :) The actual implementation is:
```
'use strict';

module.exports = {};

if ('production' != process.env.NODE_ENV) {
  Object.freeze(module.exports);
}
```

assembl/graphql/schema.py
```diff
@@ -1039,7 +1039,7 @@ def mutate(root, args, context, info):
 
         permissions = get_permissions(user_id, discussion_id)
         allowed = thematic.user_can(user_id, CrudPermissions.UPDATE, permissions)
-        if not allowed or (allowed == IF_OWNED and user_id == Everyone):
+        if not allowed:
```
vincentfretin: Yes, absolutely sure. MAP confirmed it.

assembl/static2/js/app/components/debate/thread/topPostForm.jsx
vincentfretin: we forget to remove the comment

assembl/static2/css/common.scss
vincentfretin: What do you mean? I don't understand.

assembl/static2/js/app/components/common/richTextEditor/index.jsx
vincentfretin: I'm not sure. @cedricmessiant ?

assembl/models/post.py
vincentfretin: After discussion with MAP, there was to much work to enable HistoryMixin on Post. This is not simply inheriting of HistoryMixin. So we decided to just add a modification_date field. The only fields that are versionned are subject and body via tombstoned langstringentry.

assembl/static2/package.json
vincentfretin: Actually no, when you run `yarn` again, it doesn't put the dependencies again.
Just to explain why I did that,
react-i18nify package.json defines a fixed version of prop-types 15.5.8 and react 15.5.4, which is a bad practice for libraries and react-redux-i18n depends on a fixed version of react-i18nify.
So bundle.js contained a second version of react 18 KB.
I just did a PR for that https://github.com/JSxMachina/react-i18nify/pull/35
When there is a new release of react-i18nify and react-redux-i18n, we can remove this.

assembl/static2/css/variables_cascade.scss
```diff
@@ -1,6 +1,8 @@
 
 /* This file is meant to define cascade variables. Cascade variables are variables which definition in a theme is not mandatory, but which by default have a value that depend on other theme variables. So if these theme variables they depend on are re-defined in a theme, cascade variables would have to use changed values. For this, we use the "!default" keyword of SCSS. */
 
+$first-color-opacity-70: rgba($first-color, 0.7) !default;
```
vincentfretin: Well I needed it at two places, so why not. There is already other variable with 0.5 opacity.

assembl/static2/js/app/components/common/richTextEditor/index.jsx
vincentfretin: It doesn't seem necessary actually, maybe it was in the early development. I removing it.

cedricmessiant: if the user clicks on a button in the toolbar, if you focus without a timeout you may lose the formating that was chosen. I tried without the timeout and finally I did like in `react-rte` package.

```
  _focusEditor() {
    // Hacky: Wait to focus the editor so we don't lose selection.
    setTimeout(() => {
      this.props.focusEditor();
    }, 50);
}
```

vincentfretin: Ah right, the toolbar doesn't work at all without the setTimeout. I'll add a comment in the code.

assembl/static2/js/app/client.js
```diff
@@ -18,12 +18,30 @@ const myFragmentMatcher = new IntrospectionFragmentMatcher({
   }
 });
 
+// The object id retrieved is already unique, it's actually
+// ObjectType:primaryKey encoded in base64, so we define our
+// own dataIdFromObject instead of using the default one `${o.__typename}:o.id`.
+// This allows us to define a custom resolver for the node query.
+// for more info about customResolvers, read
+// http://dev.apollodata.com/react/query-splitting.html
```
swergas: One more thing to learn :D So I'm not able to evaluate this commit (5d03f509be33ab03e9fe4f4936a748e4c6298533) yet.

assembl/graphql/schema.py
```diff
@@ -714,37 +713,6 @@ def resolve_ideas(self, args, context, info):
...
-                joinedload_all(Post.creator),
-                undefer(Post.idea_content_links_above_post)
-            )
-        if len(discussion.discussion_locales) > 1:
-            query = query.options(*models.Content.subqueryload_options())
-        else:
-            query = query.options(*models.Content.joinedload_options())
-
-        # pagination is done after that, no need to do it ourself
-        return query
```
swergas: As the implementation code of this method is quite complex, why has it been implemented if it is not used? Was it used before but not anymore? Last changes to this code were made by you.

assembl/graphql/schema.py
```diff
@@ -1324,6 +1334,71 @@ def mutate(root, args, context, info):
...
+        post_id = args.get('post_id')
+        post_id = int(Node.from_global_id(post_id)[1])
+        post = models.Post.get(post_id)
+
+        permissions = get_permissions(user_id, discussion_id)
+        allowed = post.user_can(user_id, CrudPermissions.DELETE, permissions)
+        if not allowed:
+            raise HTTPUnauthorized()
+
+        # Same logic as in assembl/views/api2/post.py:delete_post_instance
```
swergas: Did you try to factorize the code instead of repeating it? (DRY)

assembl/static2/js/app/index.jsx
```diff
@@ -10,17 +10,17 @@ import createAppStore from './store';
...
 This is not the official way of doing it.
 The official way is using the `history` library, called with `import history from './utils/history';`.
-For details, see https://github.com/ReactTraining/react-router/blob/master/FAQ.md#how-do-i-access-the-history-object-outside-of-components
+For details, see
+https://github.com/ReactTraining/react-router/blob/master/FAQ.md#how-do-i-access-the-history-object-outside-of-components
 */
 const history = browserHistory;
 
-const store = createAppStore(client);
+const store = createAppStore();
```
vincentfretin: It's to not include apollo cache in the redux store. Apollo will create its own redux store to manage its cache. This is to avoid unncessary call to mapStateToProps of redux connect containers when the apollo store is updated when receiving a query response.
Ah, I just said what I said in my commit actually. Is is clearer?

vincentfretin: There is no real reason to have the apollo cache in the same redux store, except if you use redux-persist with localstorage to have some offline features.
In next version of apollo-client, you can't integrate the apollo cache inside your redux store anyway. They removed the redux dependency.

vincentfretin: With having apollo cache in the redux store, the Idea component mapStateToProps was called each time we received a post query response.

assembl/static2/js/app/pages/idea.jsx
```diff
@@ -9,38 +9,62 @@ import IdeaWithPosts from '../graphql/IdeaWithPosts.graphql';
...
+};
+
 class Idea extends React.Component {
+  componentWillMount() {
+    this.refetchIdea = this.props.data.refetch;
+  }
+
+  componentWillReceiveProps(nextProps) {
+    // data.refetchIdea is a new instance each time Idea component is rerendered
+    // so shallowEqual fails on Post components and we have a perf issue.
```
vincentfretin: This is indeed hackish, but I found no other way to fix this perf issue.
The issue is that the graphql() function create a component that redo the query if the props changed, even if those props are not used as variables for the query, and in the process, the data props that is passed to the wrapped component is a new instance because some methods are rebound, including refetch, see https://github.com/apollographql/react-apollo/blob/master/src/graphql.tsx#L513
so when evaluating the shouldComponentUpdate of Post components, the identity equality is false for the refetchIdea=data.refetch prop that is passed down the chain to Post components, and the posts are always rerendered.

Actually now that we removed answerPostFormId from redux store and that we only have lang in mapStateToProps, we normally can remove this hack. There should be no perf issue anymore.

vincentfretin: Ok, deleted in https://github.com/assembl/assembl/commit/1c67ae082f66ba174242b6eed109f9b9254b0ee7

assembl/graphql/schema.py
vincentfretin: Yes, I implemented it in the early days, in advance, but we actually never used it and there was no tests. So yeah, deleting code that we don't use.

assembl/static2/js/app/components/common/formControlWithLabel.jsx
```diff
@@ -29,27 +31,51 @@ class FormControlWithLabel extends React.Component {
     this.setState({ errorMessage: errorMessage, validationState: validationState });
   }
 
+  renderRichTextEditor = () => {
```
vincentfretin: It doesn't replace it, there is a condition line 49 see below.

assembl/graphql/schema.py
vincentfretin: No I didn't try to refactor. It may be a good idea though. What about post.delete_post_and_extracts(permissions) ?

vincentfretin: I would need to give user_id parameter as well. and return the number_of_extracts removed. I don't like this api. I prefer not to refactor here.

vincentfretin:
Ok I think I answered all comments here.

vincentfretin:
FYI, the PR was merged, and then reverted in #142. Some commits were added to this branch. Then the revert was reverted and I merged into develop only the last commits from a feature_rich_text_review branch https://github.com/assembl/assembl/commit/7f7f320b881b09890308ec3bb299782371e53d90

PR #139, opened 2017-08-09
Table of ideas navigation (DEVAS-551+DEVAS-552) (status: Code Review)

vincentfretin:


assembl/graphql/schema.py
```diff
@@ -323,6 +323,7 @@ class Video(graphene.ObjectType):
 class IdeaInterface(graphene.Interface):
     num_posts = graphene.Int()
     num_contributors = graphene.Int()
+    num_children = graphene.Int()
```
vincentfretin: Could you please add a simple test, by adding numChildren in the following test tests/views_tests/test_graphql_ideas.py:test_graphql_get_all_ideas for example. Thanks!

assembl/static2/js/app/components/debate/common/ideas.jsx
```diff
@@ -0,0 +1,65 @@
+import React from 'react';
+import { Translate } from 'react-redux-i18n';
+import { Grid } from 'react-bootstrap';
+import '../../../../../css/components/ideas.scss';
```
vincentfretin: Please remove this import, it doesn't take into account the theme. And you already added it in _assembl_base_styles.scss anyway.

assembl/static2/js/app/components/debate/common/ideasLevel.jsx
```diff
@@ -0,0 +1,201 @@
+import React from 'react';
+import { Row, Col } from 'react-bootstrap';
+import getValue from 'lodash/get';
+import IdeaPreview from '../../common/ideaPreview';
+import VisibilityComponent from '../../common/visibilityComponent';
+import { get as getRoute } from '../../../utils/routeMap';
+import { getDiscussionSlug } from '../../../utils/globalFunctions';
+import '../../../../../css/components/ideas.scss';
```
vincentfretin: remove this scss import here

assembl/static2/js/app/utils/translations.js
```diff
@@ -118,8 +118,7 @@ const Translations = {
       existingUsername: 'Le nom d’utilisateur que vous avez renseigné existe déjà',
       existingEmail: 'L’e-mail que vous avez renseigné existe déjà',
       emailNotFound: 'L’e-mail que vous avez renseigné n’a pas été trouvé. Veuillez réessayer.',
-      incorrectPasswordLogin:
-        'Le mot de passe que vous avez entré ne correspond pas avec l’utilisateur saisi. Veuillez réessayer.',
+      incorrectPasswordLogin: 'Le mot de passe que vous avez entré ne correspond pas avec l’utilisateur saisi. Veuillez réessayer.',
```
vincentfretin: Oh, I know there is no real change here, but it's a bad security practice to divulge that the login actually exist! I can't find where this message is used in the code, maybe it's not used anymore. If this is the case, we should remove it.

swergas:
Work on this branch is far from done, it's really work in progress. I'm doing a lot of trial and errors and trying several solutions and progressively refactoring. If things change too much, I will probably re-write some commits. So please don't open a pull request for me. Lots of things you comment now will probably not exist anymore at the time it goes to CR. Yes I will try to add as many tests as I can but I doubt there will be many, as there is a short deadline for this feature.

vincentfretin:
I created the PR because I wanted to know if you changed the debateThread.jsx file, I want to change something in it, but now I see you changed it, I can wait that this feature is merged before doing the change, this is to avoid you a conflict error to resolve. ;)
While I was at it, I just did a quick review, not a deep review.
But sure, I'll stop commenting on it until the feature is done, as you wish.

For the tests, I just ask that all graphql code on the backend are tested like it is today. Really this is just 2 lines to add in an existing test.

assembl/static2/js/app/components/debate/common/ideasLevel.jsx
```diff
@@ -0,0 +1,408 @@
...
+        }
+      }
+    }
+  }
+
+  scrollToIdeaIndexIfNecessary(index) {
+    const thematicWidth = this.getOption('thematicWidth');
+    const carouselWidth = this.getOption('carouselWidth');
+    const ideaInitialX = index * thematicWidth;
+    if (ideaInitialX + thematicWidth > carouselWidth) {
```
ayyazdaniaryan: And what about the `else` condition? Is it handled outside of this method?

assembl/static2/js/app/components/debate/common/ideasLevel.jsx
```diff
@@ -0,0 +1,408 @@
...
+      this.scrollLeft.setState({ isVisible: targetScrollLeftVisibility });
+      this.scrollLeft.forceUpdate(); // FIXME: Why is this needed? component should re-render and show correct UI by itself
+      this.scrollRight.setState({ isVisible: targetScrollRightVisibility });
+      this.scrollRight.forceUpdate(); // FIXME: Why is this needed? component should re-render and show correct UI by itself
+    }
+  }
+
+  initializeConstants() {
+    let parentWidth = 1400;
+    if (this.me && this.me.parentNode && this.me.parentNode.clientWidth) {
```
ayyazdaniaryan: This code makes the component very brittle to refactoring in a parent component. I would recommend passing the parent-most `clientWidth` to this component instead. A) Easier to document usage B) More flexible usage

assembl/static2/js/app/components/debate/common/ideasLevel.jsx
```diff
@@ -0,0 +1,408 @@
...
+      parentWidth = this.me.parentNode.clientWidth;
+    }
+
+    const scrollDisplacement = this.getOption('scrollDisplacement') || 0;
+
+    let thematicWidth = 350;
+    let thematicWidthPercent = 100;
+    let numberOfThematicsToShow = 4.5;
```
ayyazdaniaryan: So, this is a bit more about readability, but I would, at top of this file, declare these as consts, for readability, awareness, and ease of change.

eg.
```
const NUM_THEMATICS_TOSHOW_BIG = 4.5;
const NUM_THEMATICS_TOSHOW_MED = 3.5;
const NUM_THEMATICS_TOSHOW_SMALL = 2.5;
```

I was not at all aware that these were in JS until I came 240 lines deep into this component. Other developers might miss it if they ever want to change it.

assembl/static2/css/components/ideas.scss
```diff
@@ -0,0 +1,104 @@
...
+            z-index: 1001;
+            right: 0;
+            &:before {
+                content: ">";
+            }
+        }
+
+        .row {
+            position: relative;
+            display: inline-flex;
```
vincentfretin: Yes we have, v2 targets IE11.

assembl/static2/css/components/ideas.scss
```diff
@@ -0,0 +1,104 @@
...
+                &:nth-child(16) {
+                    margin-top: 0;
+                    color: #fafbca;
+                }
+            }
+        }
+    }
+
+    .theme {
+        transition: all 0.5s;
```
vincentfretin: Please don't imagine use cases. If a customer didn't ask for it, don't do it.

assembl/static2/js/app/components/debate/common/ideasLevel.jsx
```diff
@@ -0,0 +1,408 @@
...
+        <VisibilityComponent
+          isVisible={isScrollLeftButtonVisible}
+          ref={(el) => {
+            this.scrollLeft = el;
+          }}
+          className="scroll-left-container"
+        >
+          <div className="scroll-left" onClick={this.onScrollLeftClick} />
+        </VisibilityComponent>
+        <Row className="no-margin" ref={this.refRow}>
```
vincentfretin: There is no issue here. The refRow is a bound function which take an el.

assembl/static2/css/components/ideas.scss
```diff
@@ -0,0 +1,104 @@
...
+        .scroll-left, .scroll-right {
+            display: none;
+            position: absolute;
+            width: 50px;
+            height: 50px;
+            background-color: rgba(195, 195, 195, 0.6);
+            top: calc(50% - 25px);
+            cursor: pointer;
+            &:before {
+                font-size: 42px;
```
paolina92: For the `font-size`, use or create a variable.

assembl/static2/css/components/ideas.scss
```diff
@@ -0,0 +1,104 @@
...
+                &:nth-child(7),
+                &:nth-child(8),
+                &:nth-child(9),
+                &:nth-child(10),
+                &:nth-child(11),
+                &:nth-child(12),
+                &:nth-child(13),
+                &:nth-child(14),
+                &:nth-child(15),
+                &:nth-child(16) {
```
paolina92: +1

assembl/static2/css/components/ideas.scss
```diff
@@ -0,0 +1,104 @@
...
+                    margin-top: 0;
+                    color: #fafbca;
```
paolina92: Create a variable to set the color!

assembl/static2/js/app/components/common/ideaPreview.jsx
```diff
@@ -0,0 +1,52 @@
+import React from 'react';
+import { Link } from 'react-router';
+import { Translate } from 'react-redux-i18n';
+import Statistic from './statistic';
+
+class IdeaPreview extends React.Component {
```
paolina92: You should use a pure component here.

assembl/static2/js/app/components/common/visibilityComponent.jsx
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
```
paolina92: I don't understand the meaning of `visibilityComponent`. Can you explain? May be we can find a better name for this react component...

assembl/static2/js/app/components/common/visibilityComponent.jsx
```diff
@@ -0,0 +1,24 @@
...
+
+class VisibilityComponent extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
```
paolina92: If we don't use the state, may be you can create a pure component?

assembl/static2/js/app/components/debate/common/ideas.jsx
```diff
@@ -0,0 +1,69 @@
+import React from 'react';
+import { Translate } from 'react-redux-i18n';
+import { Grid } from 'react-bootstrap';
+import _filter from 'lodash/filter';
+import IdeasLevel from './ideasLevel';
+
+class Ideas extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
```
paolina92: Same comment here, can we use the store instead of the state and create a pure component?

assembl/static2/js/app/components/common/ideaPreview.jsx
vincentfretin: Functional component you mean? This is different. Only React.PureComponent or React.Component implementing shouldComponentUpdate, or using connect() are pure components.

vincentfretin: Pure component means the render method is not called when the prev props and next props are the same. This is not the case for functional component, render is always called.

assembl/static2/js/app/components/common/visibilityComponent.jsx
```diff
@@ -0,0 +1,24 @@
...
+      isVisible: props.isVisible
+    };
+  }
+  render() {
+    const { className } = this.props;
+    const { isVisible } = this.state;
+    const myClassNames = isVisible === true ? className : `${className} hidden`;
```
cedricmessiant: now we can use the `classnames` module for such stuff : https://www.npmjs.com/package/classnames (we added the dependency in the richtexteditor branch)

assembl/static2/js/app/utils/translations.js
swergas: +1, but in another PR

assembl/models/idea.py
```diff
@@ -698,7 +698,8 @@ def get_discussion_id(self):
         return self.discussion_id or self.discussion.id
 
     def get_definition_preview(self):
-        body = self.definition.strip()
+        definition = self.definition or ""
```
swergas: It just makes sure that `definition` is a string so that I can call `strip()` on it on the next line (whereas it was creating an error before in the case of a None value)

assembl/static2/css/components/ideas.scss
```diff
@@ -0,0 +1,104 @@
...
+            right: 0;
+        }
+
+        .scroll-left, .scroll-right {
+            display: none;
+            position: absolute;
+            width: 50px;
+            height: 50px;
+            background-color: rgba(195, 195, 195, 0.6);
+            top: calc(50% - 25px);
```
swergas: @ayyazdaniaryan the scroll button is vertically centered, so this corresponds to setting a height, and positionning it at 50% of parent height minus half of its height. So here 25 is 50/2, where 50 is the width defined line 25.

@paolina92 OK that's a good idea to use variables for width and height as some things are computed from them. But is the `variables.scss` file the best place to declare them? These width/height variables are completely internal to this component, and are not meant to be used elsewhere. I propose for these cases we declare variables in the component's .scss file instead. What do you guys think?

assembl/static2/css/components/ideas.scss
```diff
@@ -0,0 +1,104 @@
...
+        &:hover {
+            .scroll-left, .scroll-right {
+                display: block;
+            }
+        }
+        .scroll-right {
+            z-index: 1001;
+            right: 0;
+            &:before {
+                content: ">";
```
swergas: Yes, this is a first pass for the appearance of the scroll buttons. Next iterations will probably need fontello or an image.

assembl/static2/css/components/ideas.scss
swergas: Yeah I think probability that someone asks to customize the transition durations of their debate is pretty low

assembl/static2/js/app/components/common/visibilityComponent.jsx
swergas: It's a simple higher order component that displays its children only if a visibility prop given to it is true. It's the react equivalent of our CSS `hidden` class technique.

assembl/static2/js/app/components/common/visibilityComponent.jsx
swergas: My knowledge of React is still quite limited so for sure I don't use the best solutions. But I had already tried to use only props instead of state for this component, and it was not refreshing when its props changed. Also a constraint I have is that its parent component (ideasLevel) should not re-render() in the middle of an animation otherwise animation is cut, so I can't just lift the state up. This makes things difficult. I would welcome a pair programming session on this.

assembl/static2/js/app/components/common/visibilityComponent.jsx
swergas: that's just a style preference

assembl/static2/js/app/components/debate/common/ideas.jsx
```diff
@@ -0,0 +1,69 @@
...
+      selectedIdeas: [''],
+      isInitialState: true
+    };
+    this.onSeeSubIdeasClick = this.onSeeSubIdeasClick.bind(this);
+  }
+  onSeeSubIdeasClick(ideaId, level) {
+    const updatedSelectedIdeas = this.state.selectedIdeas.slice(0, level + 2);
+    updatedSelectedIdeas[level] = ideaId;
+    updatedSelectedIdeas[level + 1] = '';
```
swergas: The ideas level which shows at the bottom is always represented by an empty string because this means that no idea is selected in this level

assembl/static2/js/app/components/debate/common/ideasLevel.jsx
```diff
@@ -0,0 +1,408 @@
...
+
+        this.scrollToIdeaIndexIfNecessary(index);
+
+        const f = () => {
+          this.onSeeSubIdeasClick(ideaId, index, step + 1);
+          this.forceUpdate();
+        };
+        f.bind(this);
+
+        setTimeout(f, 500);
```
swergas: Yes, that's because animation is in progress. As animation is in several steps, we wait until the animation of a step is finished before executing the next step.

assembl/static2/js/app/components/debate/common/ideasLevel.jsx
swergas: the else is do nothing (which means don't do any scroll), hence the name of the method scrollToIdeaIndexIfNecessary

assembl/static2/css/components/ideas.scss
```diff
@@ -0,0 +1,104 @@
...
+            display: none;
+            position: absolute;
+            width: 50px;
+            height: 50px;
+            background-color: rgba(195, 195, 195, 0.6);
+            top: calc(50% - 25px);
+            cursor: pointer;
+            &:before {
+                font-size: 42px;
+                color: #000;
```
swergas: black actually

assembl/static2/css/components/ideas.scss
swergas: this was just a kind of copy paste of what was already done for the list of thematics of phase1, overriding their style. But now @paolina92 has improved the implementation, which removed this code.

assembl/static2/css/components/ideas.scss
swergas: this code has been removed since

assembl/static2/js/app/components/common/ideaPreview.jsx
swergas: OK I just converted this component to a functional component (but from my current knowledge I'm really not sure it's worth the time spent)

assembl/static2/js/app/components/debate/common/ideas.jsx
swergas: What are the reasons for this?

assembl/static2/js/app/components/debate/common/ideasLevel.jsx
swergas: It would not be ideal either: I think the parent component does not know its clientWidth at the moment when execution is in its its render() method and it's calling the child component.

assembl/static2/js/app/components/debate/common/ideasLevel.jsx
swergas: I think this algorithm will soon be improved or refactored and these numbers will probably disappear or change their meaning, so I prefer to wait until then.

assembl/graphql/schema.py
swergas: responded in https://github.com/assembl/assembl/pull/139#issuecomment-321191247

swergas:
OK I have replied to all code review comments. Not all of them close the discussion, so feel free to continue it, but if you don't think it's necessary the branch can now be merged :)

PR #141, opened 2017-08-10
Langstring graphql

vincentfretin:
This corresponds to story DEVAS-700

### DIFF
--- a/assembl/models/discussion.py
+++ b/assembl/models/discussion.py
@@ -839,26 +839,21 @@ class Discussion(DiscussionBoundBase, NamedClassMixin):
         return self.preferences["translation_service"]
 
     def translation_service(self):
-        service_class = self.translation_service_class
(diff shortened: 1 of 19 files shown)


maparent:
OK, we have tests and corrections from those tests. Please have a look when you can.

vincentfretin:
LGTM

assembl/models/discussion.py
```diff
@@ -839,26 +839,21 @@ def translation_service_class(self):
         return self.preferences["translation_service"]
 
     def translation_service(self):
-        service_class = self.translation_service_class
+        service_class = (self.translation_service_class or
```
swergas: I don't understand the title of your commit and this replacement. Why can a language identification service be considered as a translation service? It does only identify which language is used, but it can't translate something into another language, right? Further, I see that you differenciate the two with `service.canTranslate`

PR #147, opened 2017-08-26
Sentiment mapping feature

paolina92:


assembl/static2/js/app/components/debate/thread/sentimentStats.jsx
```diff
@@ -0,0 +1,54 @@
...
+    return { color: '#FAC16F', svg: dontUnderstand };
+  case 'moreInfo':
+    return { color: '#9374FF', svg: moreInfo };
+  default:
+    return { color: '#46D081', svg: 'like' };
+  }
+};
+
+const getSentimentDetails = (totalSentimentsCount, sentimentCounts) => {
+  const sentimentStats = (
```
cedricmessiant: why `const` and then `return` ? you should directly return the jsx here

assembl/static2/js/app/components/debate/thread/postActions.jsx
```diff
@@ -97,71 +92,52 @@ class PostActions extends React.Component {
...
+                      <div className="min-sentiment" key={index} style={{ left: `${(count += 1 * 6)}px` }}>
+                        <DontUnderstand size={15} />
+                      </div>
+                    );
+                  }
+                  if (sentimentCounts[sentiment] > 0 && sentiment === 'moreInfo') {
+                    return (
+                      <div className="min-sentiment" key={index} style={{ left: `${(count += 1 * 6)}px` }}>
+                        <MoreInfo size={15} />
+                      </div>
```
n0izn0iz: refactoring this itches me ^^

cedricmessiant: I agree, just keep the list of sentiments in order and for each render something like : 

```
<div className="min-sentiment" key={index} style={{ left: `${(count += 1 * 6)}px` }}>
   <SvgComponent size={15} />
</div>
```

n0izn0iz: I'm trying to do it but I have a "react array element missing key" error while the divs seems to have keys :/

n0izn0iz: Oh, there seems to be a 'should have a unique "key" prop' error in postActions even if I comment out this part

n0izn0iz:
One feature from the user story is missing: the sentiment tooltips are black instead of the sentiment color

PR #148, opened 2017-08-28
What you need to know UI (DEVAS-630)

n0izn0iz:


assembl/static2/js/app/utils/translations.js
```diff
@@ -462,6 +463,7 @@ const Translations = {
       confirmDeletionBody: 'Are you sure you want to delete this message?',
       confirmDeletionButtonDelete: 'Delete',
       confirmDeletionButtonCancel: 'Cancel',
+      whatWeNeedToKnow: 'What we need to know',
```
ayyazdaniaryan: In English, often we wouldn't say "What we need to know", it would likely say, "What you need to know". This is a design choice though. @FrankEscoubes @paolina92, thoughts?

assembl/static2/js/app/components/debate/common/whatWeNeedToKnow.jsx
```diff
@@ -0,0 +1,14 @@
...
+import { Translate } from 'react-redux-i18n';
+
+export default ({ longTitle }) => {
+  return (
+    <div className="insert-box wwntk-box">
+      <h3 className="dark-title-3 wwntk-title">
+        <Translate value="debate.whatWeNeedToKnow" />
+      </h3>
+      <div className="box-hyphen" />
+      <p className="wwntk-text" dangerouslySetInnerHTML={{ __html: longTitle }} />
```
ayyazdaniaryan: Can you add a TODO comment here that we need to cleanse longTitle before setting HTML content into the `<p>` tag? This is part of our ongoing struggle with security, to not allow an attacker to override our API response and pass xss script here and we blindly add. Thanks

FrankEscoubes:
"What you need to know"

PR #149, opened 2017-08-28
Add google+ facebook and twitter icons (DEVAS-710)

n0izn0iz:
https://bluenove.atlassian.net/browse/DEVAS-710


assembl/static2/config.json
```diff
@@ -521,6 +521,24 @@
       "src": "fontawesome"
     },
     {
+      "uid": "b945f4ac2439565661e8e4878e35d379",
```
ayyazdaniaryan: Will these now be the replacement for the `Glyphicon`s that are currently used in the footer and maybe the login page?

ayyazdaniaryan: Those icons were svg's, mind you...

n0izn0iz: The icons spots were empty on the login page and I can't find icons in footers so I have no idea, the task says:
> Reintroduce the icons with name twitter facebook google in the fontello font

n0izn0iz: Mb, the glyphicons were not used on the login page, there were already <i> tags with the correct ccs classes (ex: assembl-icon-google) but they are indeed used in the footer with no plan to replace them to my knowledge

ayyazdaniaryan: @paolina92 Perhaps we can remove these glyphs in favour of these new social media icons? One is a font (and loads much quicker), the other is an svg, which takes longer to load and more computation on CPU per render.

n0izn0iz: The glyphicons in the footer are not displayed if there is an empty socialMedias variable and I can't find how to populate it

n0izn0iz: Found it

n0izn0iz: The icons in the footer are different (white circle with black icon) from the ones in the login page (white icon without circle).
I would need to add the circle icons to the fontello font and create assembl-icon-{socialMediaName}-circle classes

vincentfretin:
LGTM, please merge.

PR #150, opened 2017-08-29
Auto subject prefix derived from post level and index

n0izn0iz:
Double check the backend changes :P

### DIFF
--- a/assembl/static2/css/variables.scss
+++ b/assembl/static2/css/variables.scss
@@ -86,6 +86,12 @@ $img-dir: '/static2/img';
   src: local('☺'), url('#{$fonts-dir}/Lato-LightItalic.woff') format('woff'), url('#{$fonts-dir}/Lato-LightItalic.ttf') format('truetype');
 }
 
+@font-face {
+  font-family: 'LatoHairline';
+  src: url('#{$fonts-dir}/Lato-Hairline.eot');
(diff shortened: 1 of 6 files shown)


assembl/static2/css/variables.scss
```diff
@@ -86,6 +86,12 @@ $img-dir: '/static2/img';
   src: local('☺'), url('#{$fonts-dir}/Lato-LightItalic.woff') format('woff'), url('#{$fonts-dir}/Lato-LightItalic.ttf') format('truetype');
 }
 
+@font-face {
```
ayyazdaniaryan: If you're going to add this font-family to our css, you should also add this to our styleguide. Nobody outside of those who review this PR will know of this in the future. @paolina92 thoughts?

assembl/views/api/post.py
```diff
@@ -669,7 +669,7 @@ def create_post(request):
             locale = discussion.main_locale
         # print subject
         if subject is not None and len(subject):
-            new_subject = "Re: " + restrip_pat.sub('', subject).strip()
```
vincentfretin: @n0izn0iz I disagree. The old interface should stay unchanged. I think you should revert this commit and remove the 'Re: ' in javascript.

vincentfretin: Ok, I reverted in https://github.com/assembl/assembl/commit/f43d413412f2e9ad27c4242ced6e6e1d213d82e8 and did it in js https://github.com/assembl/assembl/commit/6e9a20adfc30545605d149205ba2264aa0f51c6d

PR #157, opened 2017-09-01
Fix colored tooltips arrow (DEVAS-720)

n0izn0iz:
+ https://bluenove.atlassian.net/browse/DEVAS-720


vincentfretin:
@paolina92 Is this good for you? Can we merge?

PR #162, opened 2017-09-11
bug fix title editing in responses

MilanBarande:
add state to post to track if it's top post and edit post component will not allow title edit on responses

### DIFF
--- a/assembl/static2/js/app/components/debate/thread/editPostForm.jsx
+++ b/assembl/static2/js/app/components/debate/thread/editPostForm.jsx
@@ -20,6 +20,7 @@ type EditPostFormProps = {
   body: string,
   id: string,
   subject: string,
+  readOnly: boolean,
   goBackToViewMode: Function,
   mutate: Function
 };
@@ -107,6 +108,7 @@ class EditPostForm extends React.PureComponent<void, EditPostFormProps, EditPost
               value={this.state.subject}
               handleTxtChange={this.updateSubject}
               maxLength={TEXT_INPUT_MAX_LENGTH}
+              readOnly={this.props.readOnly}
             />
             <FormGroup>
               <div className="form-label">
--- a/assembl/static2/js/app/components/debate/thread/post.jsx
+++ b/assembl/static2/js/app/components/debate/thread/post.jsx
@@ -35,8 +35,8 @@ const getFullLevelString = (fullLevel) => {
 // We only have french and english for en, fr, ja for now.
 
 class Post extends React.PureComponent {
-  constructor() {
-    super();
+  constructor(props) {
+    super(props);
     this.state = {
       showAnswerForm: false,
       mode: 'view',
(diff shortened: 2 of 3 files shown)


assembl/static2/js/app/components/debate/thread/post.jsx
```diff
@@ -35,12 +35,13 @@ const getFullLevelString = (fullLevel) => {
...
-  constructor() {
-    super();
+  constructor(props) {
+    super(props);
     this.state = {
       showAnswerForm: false,
       mode: 'view',
-      showOriginal: false
+      showOriginal: false,
+      isTopPost: !props.data.post.parentId
```
vincentfretin: there is no need to set that is the state

assembl/static2/js/app/graphql/Post.graphql
```diff
@@ -28,7 +28,7 @@ fragment Post on Post {
   creator {
     ...AgentProfileInfo
   }
-  # parentId
+  parentId
```
vincentfretin: no need, you should already have parentId in props.parentId from the IdeaWithPosts query

assembl/static2/js/app/components/debate/thread/editPostForm.jsx
```diff
@@ -107,6 +108,7 @@ class EditPostForm extends React.PureComponent<void, EditPostFormProps, EditPost
               value={this.state.subject}
               handleTxtChange={this.updateSubject}
               maxLength={TEXT_INPUT_MAX_LENGTH}
+              isTopPost={this.props.isTopPost}
```
vincentfretin: please use a generic name, like readOnly

assembl/static2/js/app/graphql/Post.graphql
ayyazdaniaryan: Because this is happening on the post component itself, we observed that parentId is not available in the constructor. Why was this commented in the first place?

assembl/static2/js/app/components/debate/thread/post.jsx
ayyazdaniaryan: This is true. Can be removed. Use the `this.props.data.post.parentId` further below.

assembl/static2/js/app/graphql/Post.graphql
vincentfretin: Why do you want it in the constructor, to set an initial state? There is no need for a state here. It is commented to not fetch it again, we already get this data with the IdeaWithPosts query. See http://dev.apollodata.com/react/query-splitting.html for the concept of query splitting. The IdeaWithPosts and PostQuery should not overlap.

MilanBarande: Ok c'est réglé, merci des commentaires Vincent!

assembl/static2/js/app/components/debate/thread/editPostForm.jsx
cedricmessiant: +1 please don't use specific stuff in common components

vincentfretin:
Ok pour merge, mais tu devrais faire qu'un seul commit au lieu de 3.

PR #163, opened 2017-09-11
Per-discussion phase marker of interface used

maparent:


vincentfretin:
LGTM but I don't see where you create an entry in discussion_phase table when you add a new entry in timeline. Did you forgot it?

maparent:
No need to create table entry, that is done by the sqla ORM. The difference is between making that class have a table or not.

PR #168, opened 2017-09-13
Fix survey doughnut on small screens

n0izn0iz:
https://bluenove.atlassian.net/browse/DEVAS-751


assembl/static2/js/app/constants.js
```diff
@@ -1,3 +1,4 @@
+export const EXTRA_SMALL_SCREEN_WIDTH = 600;
 export const SMALL_SCREEN_WIDTH = 768;
```
paolina92: One day we should use the same value everywhere....

PR #171, opened 2017-09-19
Display the nuggets next to the post (DEVAS-629)

n0izn0iz:
https://bluenove.atlassian.net/browse/DEVAS-629

### DIFF
--- a/assembl/static2/css/components/posts.scss
+++ b/assembl/static2/css/components/posts.scss
@@ -260,7 +260,7 @@ $deleted-post-border: 1px solid $grey1 !default;
 
       .badges {
         line-height: 26px;
-        font-family: 'LatoLight';
+        font-family: $font-family4;
 
         .nugget-img {
           text-align: center;
@@ -277,7 +277,7 @@ $deleted-post-border: 1px solid $grey1 !default;
         .nugget {
 
           .nugget-txt {
-            font-style: italic;
+            font-family: $font-family6;
             color: $first-color;
             margin-bottom: 25px;
           }
--- a/assembl/static2/css/components/posts.scss
+++ b/assembl/static2/css/components/posts.scss
@@ -250,6 +250,44 @@ $deleted-post-border: 1px solid $grey1 !default;
             margin: 3px 0 8px;
         }
     }
+
+    .extracts {
+      position: absolute;
+      left: 2vw;
+      max-width: 230px;
+      width: 14vw;
+      line-height: 30px;
+
+      .badges {
+        line-height: 26px;
+        font-family: 'LatoLight';
+
+        .nugget-img {
+          text-align: center;
+          margin-bottom: 25px;
+
+          span {
+            display: inline-block;
+          }
+          span::before {
+            font-size: 50px;
+          }
+        }
+
+        .nugget {
+
+          .nugget-txt {
+            font-style: italic;
+            color: $first-color;
+            margin-bottom: 25px;
+          }
+
+          .box-hyphen {
+            border-top-width: 3px;
+          }
+        }
+      }
+  }
 }
 
 @media screen and (min-width: 600px) {
@@ -475,6 +513,10 @@ $deleted-post-border: 1px solid $grey1 !default;
             .assembl-icon-back-arrow {
                 padding: 0;
             }
+        }
+
+        .extracts {
+          display: none;
         }
     }
 }
--- a/assembl/static2/js/app/components/debate/thread/nuggets.jsx
+++ b/assembl/static2/js/app/components/debate/thread/nuggets.jsx
@@ -1,0 +1,106 @@
+import React from 'react';
+import range from 'lodash/range';
+
+import { getDomElementOffset, setDomElementOffset } from '../../../utils/globalFunctions';
+
+class Nuggets extends React.Component {
+  static get SPACER_SIZE() {
+    // Vertical space between overflowing extracts in px
+    return 100;
+  }
+
+  static completeLevelFromId(id) {
+    return id.split('-').slice(1).map((ident) => {
+      return Number(ident);
+    });
+  }
+
+  static completeLevel(rowIndex, fullLevel) {
+    return fullLevel ? [rowIndex, ...fullLevel] : [rowIndex];
+  }
+
+  static completeLevelGreater(a, b) {
+    return range(Math.min(a.length, b.length)).reduce((result, index) => {
+      if (result !== null) return result;
+      if (a[index] === b[index] && (a.length === index + 1 || b.length === index + 1)) return a.length > b.length;
+      return a[index] !== b[index] ? a[index] > b[index] : result;
+    }, null);
+  }
+
+  static previousExtract(completeLevel, otherExtracts) {
+    return !otherExtracts || !otherExtracts.length
+      ? null
+      : otherExtracts.reduce((prevExtract, extract) => {
+        const currentCompleteLevel = prevExtract ? Nuggets.completeLevelFromId(prevExtract.getAttribute('id')) : null;
+        const otherCompleteLevel = Nuggets.completeLevelFromId(extract.getAttribute('id'));
+        if (
+          Nuggets.completeLevelGreater(completeLevel, otherCompleteLevel) &&
+            (prevExtract === null || Nuggets.completeLevelGreater(otherCompleteLevel, currentCompleteLevel))
+        ) {
+          return extract;
+        }
+        return prevExtract;
+      }, null);
+  }
+
+  componentDidMount() {
+    this.callForceUpdate = this.callForceUpdate.bind(this);
+    document.addEventListener('rowHeightRecomputed', this.callForceUpdate);
+  }
+
+  componentDidUpdate() {
+    const { fullLevel, rowIndex, postId } = this.props;
+    const extracts = [].slice.call(document.getElementsByClassName('extracts'));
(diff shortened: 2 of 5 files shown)


assembl/static2/js/app/components/debate/thread/nuggets.jsx
```diff
@@ -0,0 +1,106 @@
...
+
+  render() {
+    const { extracts, fullLevel, rowIndex } = this.props;
+    const thisCompleteLevel = fullLevel ? [rowIndex, ...fullLevel] : [rowIndex];
+    return extracts.length > 0
+      ? <div
+        id={`extracts-${thisCompleteLevel.join('-')}`}
+        ref={(node) => {
+          this.node = node;
+          return this.node;
```
vincentfretin: You don't need to return anything here. This is not used.

assembl/static2/js/app/components/debate/thread/nuggets.jsx
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('rowHeightRecomputed', this.callForceUpdate);
+  }
+
+  callForceUpdate() {
+    // This is to prevent the event from passing arguments to forceUpdate, raising an error.
+    // I did not use an arrow function because we need to pass the function to removeEventListener
+    this.forceUpdate();
```
paolina92: Another approach... Instead of the `forceUpdate` you can do that:

```
componentDidMount() {
    this.setState({rowHeight: height});
    window.addEventListener('resize', rowHeightRecomputed);
}
componentWillUnmount() {
    window.removeEventListener('resize', rowHeightRecomputed);
  }
```

assembl/static2/js/app/components/debate/thread/nuggets.jsx
```diff
@@ -0,0 +1,106 @@
...
+import { getDomElementOffset, setDomElementOffset } from '../../../utils/globalFunctions';
+
+class Nuggets extends React.Component {
+  static get SPACER_SIZE() {
+    // Vertical space between overflowing extracts in px
+    return 100;
+  }
+
+  static completeLevelFromId(id) {
+    return id.split('-').slice(1).map((ident) => {
```
paolina92: Why do you use the DOM id?

assembl/static2/js/app/components/debate/thread/nuggets.jsx
n0izn0iz: The row height is not used, the nuggets component needs to rerender whenever a post is loaded which is what the rowHeightRecomputed event signals (there is no rowHeightRecomputed function)

assembl/static2/js/app/components/debate/thread/nuggets.jsx
n0izn0iz: because I did not found a better way to find the preceding extract

assembl/static2/js/app/utils/globalFunctions.js
```diff
@@ -131,4 +131,21 @@ export const getDomElementOffset = (el) => {
   const scrollTop = getDocumentScrollTop();
   const scrollLeft = window.pageXOffset || document.documentElement.scrollLeft || document.body.scrollLeft || 0;
   return { top: rect.top + scrollTop, left: rect.left + scrollLeft };
+};
+
+export const setDomElementOffset = (el, offset) => {
```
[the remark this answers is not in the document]
n0izn0iz: I need this function to compute the new absolute top. (dom offsets are weird)
But I could modify it to return an object with the new top and use react state + style prop instead of 'elem.style.top = smth' you're right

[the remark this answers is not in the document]
n0izn0iz: Or not, since I need the dom element to compute the top I need to set.. Will look into it.

assembl/static2/js/app/components/debate/thread/nuggets.jsx
```diff
@@ -0,0 +1,105 @@
...
+      if (a[index] === b[index] && (a.length === index + 1 || b.length === index + 1)) return a.length > b.length;
+      return a[index] !== b[index] ? a[index] > b[index] : result;
+    }, null);
+  }
+
+  static previousExtract(completeLevel, otherExtracts) {
+    return !otherExtracts || !otherExtracts.length
+      ? null
+      : otherExtracts.reduce((prevExtract, extract) => {
+        const currentCompleteLevel = prevExtract ? Nuggets.completeLevelFromId(prevExtract.getAttribute('id')) : null;
```
ayyazdaniaryan: So you're using the DOM id to fetch the previous nugget. But you have a react Ref. Why not use the `this.node` to find the component (Maybe you will have to keep a state with the name of the component so you can easily find the node). Check to see if this is feasable. Perhaps the methods `completeLevelFromId` and `completeLevel` might also not be needed with this method.

assembl/static2/css/components/posts.scss
```diff
@@ -444,5 +482,9 @@ $deleted-post-border: 1px solid $grey1 !default;
                 padding: 0;
             }
         }
+
+        .extracts {
```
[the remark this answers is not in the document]
n0izn0iz: Nope

assembl/static2/js/app/components/debate/thread/nuggets.jsx
n0izn0iz: I found a way to not use forceUpdate

assembl/static2/js/app/components/debate/thread/nuggets.jsx
n0izn0iz: I think I can do this but walking through the dom doesn't seems very resilient (if a component structure changes)

assembl/static2/js/app/components/debate/thread/post.jsx
```diff
@@ -167,6 +171,29 @@ class Post extends React.PureComponent {
     };
     return (
       <div className="posts" id={id}>
+        {extracts.length
+          ? <div className="extracts" style={{ position: 'absolute', left: '100px', width: '250px', lineHeight: '30px' }}>
```
ayyazdaniaryan: The styles here, don't look to be changing dynamically. You should instead either add these styles to the css class `extracts` or create a new class and add that to the list of classnames.

assembl/static2/js/app/components/debate/thread/post.jsx
```diff
@@ -167,6 +171,29 @@ class Post extends React.PureComponent {
...
+            <div className="badges">
+              {extracts.map((extract) => {
+                return (
+                  <div className="nugget">
+                    <div className="nugget-img">
+                      <span className="assembl-icon-pepite color2" style={{ fontSize: '50px' }}>
```
ayyazdaniaryan: Same CSS comment for this style.

assembl/static2/js/app/components/debate/thread/post.jsx
```diff
@@ -167,6 +171,29 @@ class Post extends React.PureComponent {
...
+                          &nbsp;
+                      </span>
+                    </div>
+                    <div className="nugget-txt" style={{ padding: '25px 0', marginRight: '20px', fontStyle: 'italic' }}>
+                      <span key={extract.id} style={{ color: '#5404D7' }}>
```
ayyazdaniaryan: Also for the styles here. Also, this colour is defined as `$purple` in `_variables.scss`. Use the variables as much as possible. This will ensure that updating themes in 1 place will be propagated across the entire code-base.

assembl/static2/js/app/components/debate/thread/post.jsx
n0izn0iz: You're commenting on old commits

assembl/static2/js/app/components/debate/thread/post.jsx
n0izn0iz: same

assembl/static2/js/app/components/debate/thread/post.jsx
n0izn0iz: same

assembl/static2/js/app/components/debate/thread/nuggets.jsx
```diff
@@ -4,8 +4,8 @@ import { getDomElementOffset, computeDomElementOffset } from '../../../utils/glo
 
 class Nuggets extends React.Component {
   static nodeExtracts(node) {
-    let result = node.getElementsByClassName('posts')[0];
-    if (result) result = result.getElementsByClassName('extracts')[0];
+    let result = Nuggets.getChildsByClassName(node, 'posts')[0];
+    if (result) result = Nuggets.getChildsByClassName(result, 'extracts')[0];
```
ayyazdaniaryan: Ahh okay, you used this same method as before. Yes, traversing the DOM is restrictive. What you could have done, in fact, is changed the component structure a bit. You could have passed the parent node as a prop to each nugget. And then, using `this.node.props` discovered if there is a parent, and computed the level that way. But, anyways, this isn't a bad option either. It's more jQuery-style than React style. Keep that in mind for next time.

assembl/static2/js/app/components/debate/thread/nuggets.jsx
```diff
@@ -74,7 +76,7 @@ class Nuggets extends React.Component {
   render() {
     const { extracts, fullLevel, rowIndex } = this.props;
     const style = this.state.top === null ? {} : { top: this.state.top };
-    return extracts.length > 0
+    return Array.isArray(extracts) && extracts.length > 0
```
ayyazdaniaryan: Good

PR #172, opened 2017-09-19
Local login

maparent:


assembl/models/auth.py
```diff
@@ -1682,6 +1693,7 @@ def getCurrent(cls, req=None):
             req = get_current_request()
         assert req
         if getattr(req, "lang_prefs", 0) is 0:
+            # lang. prefs remain effective even without shared login
             user_id = req.authenticated_userid
```
vincentfretin: while you're at it, use `authenticated_userid(req)`
The syntax req.authenticated_userid is deprecated.

assembl/models/preferences.py
```diff
@@ -554,6 +554,21 @@ def get_preference_data_list(self):
             "default": True
         },
 
+        # Does the Idea panel automatically open when an idea is clicked? (and close when a special section is clicked)
```
vincentfretin: Is this a bad copy-paste? The comment doesn't seem related to this new config.

assembl/static2/js/app/components/common/avatar.jsx
```diff
@@ -24,10 +25,17 @@ class ProfileIcon extends React.Component {
       next: location
     });
   }
+  redirectToJoin() {
+    const slug = { slug: getDiscussionSlug() };
+    browserHistory.push(get('join', slug));
+  }
```
vincentfretin: I think you need to push to customBrowserHistory created in index.jsx
You should export the history variable from index.jsx and use it here.

assembl/views/__init__.py
```diff
@@ -316,6 +317,7 @@ def process_export_list(ls):
         application_url=application_url,
         get_route=get_route,
         user=user,
+        loggedin_userid=request.authenticated_userid,
```
vincentfretin: `authenticated_userid(request)`

vincentfretin:
Other than that, LGTM.

vincentfretin:
execute `npm run format` to format with prettier and eslint --fix the js code in v2. This is the same as using the prettier-atom plugin for atom.

maparent:
Adjusted to the comments. Good catch about req.authenticated_userid being deprecated.
One thing before we merge: The join page needs to be made prettier. I think someone else will be better positioned to do it. 
Also, I think everyone is aware that this only works as such for v2 discussions, and other limits discussed by email. (Adding them to jira while I'm at it.)

assembl/static2/js/app/components/common/avatar.jsx
```diff
@@ -1,9 +1,10 @@
 import React from 'react';
+import { customBrowserHistory } from '../../index';
```
vincentfretin: customBrowserHistory is not currently exported in index.jsx. I don't see the change in your PR. You need to add `export customBrowserHistory` in index.jsx.

maparent: That's right... why didn't webpack complain? (I even tested!)

vincentfretin: webpack will not complain, eslint should have. Try `npm run eslint`

maparent: ok, that was helpful, thank you. Have a look now?

vincentfretin: Ok, should be good.

assembl/static2/js/app/components/common/avatar.jsx
```diff
@@ -24,10 +25,17 @@ class ProfileIcon extends React.Component {
...
+    const slug = { slug: getDiscussionSlug() };
+    customBrowserHistory.push(get('join', slug));
+  }
   render() {
     const slug = { slug: getDiscussionSlug() };
     const connectedUserId = getConnectedUserId();
     const connectedUserName = getConnectedUserName();
+    const alreadyLoggedIn = !!getLoggedInUserId();
+    const loginOrJoinUrl = alreadyLoggedIn ? get('join', slug) : getContextual('login', slug);
+    const onClickRedirect = alreadyLoggedIn ? this.redirectToJoin : null;
```
ayyazdaniaryan: I think instead of `null`, the standard way is to return `false`

assembl/views/__init__.py
```diff
@@ -316,6 +317,7 @@ def process_export_list(ls):
         application_url=application_url,
         get_route=get_route,
         user=user,
+        loggedin_userid=authenticated_userid(request),
```
ayyazdaniaryan: Shouldn't it be `effectiveuserid` here instead?

assembl/views/discussion/views.py
```diff
@@ -189,6 +193,11 @@ def is_login_route(route_name):
         "do_password_change")
 
 
+def is_join_route(request):
+    # TODO: Get a real shared route system for react.
```
ayyazdaniaryan: Agreed. Absolute TODO.

assembl/views/discussion/views.py
```diff
@@ -374,6 +394,54 @@ def purl_post(request):
...
+    notifications = [localizer.translate(n.get_human_readable_description())
+                     for n in ut.notification_subscriptions
+                     if n.status == NotificationSubscriptionStatus.ACTIVE]
+    data = get_default_context(request)
+    data['notifications'] = notifications
+    return data
+
+
+@view_config(route_name='join', request_method='POST')
+def do_join_discussion(request):
```
ayyazdaniaryan: As the ultimate response of this view is to redirect to either a login page or throw an error, we should utilize the new error page that was created for exactly these cases. Instead of returning `HTTPUnauthorized` or others, should redirect to error page, and pass the error in.

Downstream, that view can respond to the appropriate `Error` with a client-friendly view (this is not done for all cases, only 404, 401 and 500 I recall)

maparent: You have an error page now? Then I think that the proper thing to do is to have HTTPUnauthorized go to that error page! Pyramid allows that. Let's discuss it.

ayyazdaniaryan: Yup, I discussed it with you a long while back, when logged in users wanted to access a private discussion. Yes, let's discuss when you have a moment.

maparent: ok now that we have a generic error page, this code stands.

assembl/views/__init__.py
maparent: loggedin_userid is explicitly about knowing that the person is logged in even though they have no role in the discussion. Adding a comment.

maparent:
@vincentfretin  BTW, I was revisiting authenticated_userid: It's actually the security function that's deprecated, not the request attribute. One of us should revert this.

vincentfretin:
Oh 💩

vincentfretin:
We'll see that when we know what do to with this branch. It's on standby for the moment.

paolina92:
@maparent What do we do with this PR?

PR #173, opened 2017-09-20
changed font size on metrics and call to action, reversed hyphen rota…

MilanBarande:
…tion

### DIFF
--- a/assembl/static2/css/components/posts.scss
+++ b/assembl/static2/css/components/posts.scss
@@ -112,6 +112,9 @@ $deleted-post-border: 1px solid $grey1 !default;
 
     .post-left {
         padding: 35px;
+        .dark-title-3 {
+          line-height: 1.4
+        }
     }
 
     .post-right {
@@ -160,6 +163,7 @@ $deleted-post-border: 1px solid $grey1 !default;
             color: $grey2;
             font-size: $font-size-xs;
             padding: 0 0 0 10px;
+            font-family: 'LatoLight'
         }
     }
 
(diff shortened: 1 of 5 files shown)


assembl/static2/css/components/posts.scss
```diff
@@ -160,6 +163,7 @@ $deleted-post-border: 1px solid $grey1 !default;
             color: $grey2;
             font-size: $font-size-xs;
             padding: 0 0 0 10px;
+            font-family: 'LatoLight'
```
ayyazdaniaryan: You want to use variables instead, so that if a custom theme changes the variable, these styles will be affected. Use `$font-family4` here, instead.

assembl/static2/css/components/posts.scss
```diff
@@ -179,13 +183,15 @@ $deleted-post-border: 1px solid $grey1 !default;
         .label {
             color: $grey2;
             font-size: $font-size-xs;
+            font-family: 'LatoLight'
```
ayyazdaniaryan: Same

assembl/static2/css/components/posts.scss
```diff
@@ -179,13 +183,15 @@ $deleted-post-border: 1px solid $grey1 !default;
...
         }
 
         .badges {
             padding: 10px 0;
 
             .badge {
-                background-color: $grey1;
+              display: inline;
+                background-color: #e5e5e5;
```
ayyazdaniaryan: We try to keep to a colour palette that is defined the `variables.scss`. We have 6 grey variables. If this colour is to be used elsewhere, then it should warrant the creation of another grey variable. If not, then find the closest grey variable that already exists to this value. Thanks

PR #183, opened 2017-09-28
Share a post

vincentfretin:


assembl/static2/js/app/pages/idea.jsx
```diff
@@ -54,8 +54,35 @@ const noRowsRenderer = () => {
 };
 
 class Idea extends React.Component {
+  topPostsToLoad = (topPosts, edges) => {
+    const { hash } = window.location;
```
ayyazdaniaryan: Again, we're using the framework, which recommends the use its props than window.location. Should use the react-router's prop (passed from parent to its children).

assembl/static2/js/app/routes.jsx
```diff
@@ -29,6 +29,24 @@ import TwoColumnsAdmin from './pages/twoColumnsAdmin';
...
+  const { hash } = window.location;
+  if (hash !== '') {
+    const id = hash.replace('#', '');
+    // Push onto callback queue so it runs after the DOM is updated,
+    // this is required when navigating from a different page so that
+    // the element is rendered on the page before trying to getElementById.
+    setTimeout(() => {
+      const element = document.getElementById(id);
+      if (element) {
+        const offset = getDomElementOffset(element).top - 160;
```
ayyazdaniaryan: Minor: I would make 160px a variables inside of the global variables file and use it here.

assembl/static2/js/app/components/debate/thread/postActions.jsx
```diff
@@ -55,6 +59,23 @@ class PostActions extends React.Component {
...
+      const url = `${window.location.protocol}//${window.location.host}${get('debate', {
+        slug: slug,
+        phase: phase
+      })}${get('theme', {
+        themeId: themeId
+      })}/#${postId}`;
+      const social = debateData.useSocialMedia;
+      const src = `/static/widget/share/index.html?u=${encodeURI(url)}&t=${encodeURI(title)}&s=${encodeURI(social)}`;
+      const iframeTitle = I18n.t('Social buttons to share the post');
+      const body = <iframe src={src} width="100%" height="200" frameBorder="0" title={iframeTitle} />;
```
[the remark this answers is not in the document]
cedricmessiant: yes, we should avoid `window` and `document` in components when it is possible

assembl/static2/js/app/pages/idea.jsx
```diff
@@ -54,8 +54,35 @@ const noRowsRenderer = () => {
...
+    if (hash !== '') {
+      const id = hash.replace('#', '');
+      const post = edges.filter((edge) => {
```
cedricmessiant: use `.find` instead of `.filter` if it is for only one element : https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

assembl/static2/js/app/pages/idea.jsx
```diff
@@ -54,8 +54,35 @@ const noRowsRenderer = () => {
...
+        return edge.node.id === id;
+      });
+      if (post.length === 0) {
+        return null;
+      }
+      let topPostId = post[0].node.parentId;
```
cedricmessiant: it's more `parentPostId` here than `topPostId`, isn't it?

cedricmessiant: from what I understand, you are really searching for the top post for this post. So it wont work for a post that is at level 3 or more ? you have to recursively search for the first post in the ancestors of `post` that have no `parentId`

assembl/static2/js/app/pages/idea.jsx
```diff
@@ -102,6 +134,7 @@ class Idea extends React.Component {
                     contentLocale={contentLocale}
                     lang={lang}
                     data={topPosts}
+                    initialRow={this.topPostsToLoad(topPosts, ideaWithPostsData.idea.posts.edges)}
```
cedricmessiant: I find `topPostsToLoad` name confusing, as it returns the index of the last top post that we want to load before the scroll (if I'm right). Maybe `getInitialRowIndex` or `getTopPostToScrollIndex` (hum...)

assembl/static2/js/app/pages/idea.jsx
```diff
@@ -54,8 +54,35 @@ const noRowsRenderer = () => {
...
+      }
+      let topPostId = post[0].node.parentId;
+      if (!topPostId) {
+        topPostId = post[0].node.id;
+      }
+      const index = topPosts
+        .map((value) => {
+          return value.id;
+        })
+        .indexOf(topPostId);
```
cedricmessiant: `topPosts.findIndex(value => value.id == topPostId)` does the job here, no ? (I haven't tested it)

assembl/static2/js/app/services/debateService.js
```diff
@@ -32,21 +31,25 @@ export const buildDebateData = (debateData, prefs, timeline) => {
...
 export const getDebateData = (debateId) => {
   const url1 = `/data/Discussion/${debateId}`;
   const url2 = `/data/Discussion/${debateId}/preferences`;
   const url3 = `/data/Discussion/${debateId}/timeline_events/`;
+  const url4 = `/data/Discussion/${debateId}/settings/social_sharing`;
   const request1 = xmlHttpRequest({ method: 'GET', url: url1 });
   const request2 = xmlHttpRequest({ method: 'GET', url: url2 });
   const request3 = xmlHttpRequest({ method: 'GET', url: url3 });
-  return Promise.all([request1, request2, request3]).then((results) => {
+  const request4 = xmlHttpRequest({ method: 'GET', url: url4 });
```
cedricmessiant: why not use better names for these urls and requests ?

assembl/static2/js/app/pages/idea.jsx
vincentfretin: Ah you're right! I need to recursively do it.

vincentfretin: So yeah, it's really topPostId I want here.

assembl/static2/js/app/components/debate/thread/postActions.jsx
```diff
@@ -55,6 +59,23 @@ class PostActions extends React.Component {
       (connectedUserId === String(creatorUserId) && connectedUserCan(Permissions.DELETE_MY_POST)) ||
       connectedUserCan(Permissions.DELETE_POST);
     const userCanEditThisMessage = connectedUserId === String(creatorUserId) && connectedUserCan(Permissions.EDIT_MY_POST);
+    const { slug, phase, themeId } = routerParams;
+    const confirmModal = () => {
+      const title = postSubject;
+      const url = `${window.location.protocol}//${window.location.host}${get('debate', {
```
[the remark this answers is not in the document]
vincentfretin: @ayyazdaniaryan What are you talking about? https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/api/location.md ? It doesn't seem to have the domain in here.

assembl/static2/js/app/components/debate/thread/postActions.jsx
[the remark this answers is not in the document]
vincentfretin: @cedricmessiant I think you commented on the wrong comment. You meant for window.location right? It doesn't seem we cant get the domain with react-router's location. Even if I pass window.location as a prop, there will be a window.location somewhere...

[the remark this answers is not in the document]
vincentfretin: @ayyazdaniaryan I'm not sure about the height in pixel vs percent here. I think the height in percent here can have very different behavior with window size. I think it's preferable to have it in pixels here.

assembl/static2/js/app/pages/idea.jsx
vincentfretin: @ayyazdaniaryan Ah here for the hash yes, we can retrieve it from react-router's location. It was not the case above it seems for protocol and host. But the react-router do in the end window.location right? I think here we need to add withRouter to have it, so we complexify the code for almost nothing IMO.

assembl/static2/js/app/routes.jsx
vincentfretin: I'm against adding a constant in a global variable file if it's used only at one place.

assembl/static2/js/app/pages/idea.jsx
vincentfretin: find and findIndex are much better indeed.

assembl/static2/js/app/pages/idea.jsx
vincentfretin: ok, getInitialRowIndex it is

assembl/static2/js/app/pages/idea.jsx
vincentfretin: yeah, it should do it

assembl/static2/js/app/pages/idea.jsx
vincentfretin: Hum, I tried to use  `<Idea ... location={props.location} >` in routes.jsx but this.props.location was undefined in getInitialRowIndex... I gave up.

assembl/static2/js/app/components/debate/thread/postActions.jsx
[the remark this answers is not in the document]
n0izn0iz: There is a ResponsiveEmbed component in react-bootstrap (don't mind me if this is not relevant)

PR #187, opened 2017-10-03
Langstring ideas: elasticsearch changes

maparent:
New version of elasticsearch. Specify which languages should have a separate index. Add langstring fields. Specify analyzer (may be a mistake or a noop, please advise.) Frontend follows these indexes.

### DIFF
--- a/assembl/static2/js/app/components/search.jsx
+++ b/assembl/static2/js/app/components/search.jsx
@@ -51,6 +51,7 @@ import { getConnectedUserId, getDebateId, getLocale } from '../reducers/contextR
 import { connectedUserIsExpert } from '../utils/permissions';
 
 const FRAGMENT_SIZE = 400;
+const elasticsearchLangIndexes = document.getElementById('elasticsearchLangIndexes').value.split(' ');
 
 const truncateText = (text) => {
   return truncate(text, { length: FRAGMENT_SIZE, separator: ' ', omission: ' [...]' });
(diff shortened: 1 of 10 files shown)


assembl/static2/js/app/components/search.jsx
```diff
@@ -130,26 +131,40 @@ const ImageType = (props) => {
...
+  }
+  elasticsearchLangIndexes.forEach((lang) => {
+    const key = `${prop}_${lang}`;
+    if (source[key]) {
+      result = source[key];
+    }
+  });
+  if (result) {
+    return result;
+  }
```
vincentfretin: This condition should be inside the forEach, no? So it can return the content of the first lang in elasticsearchLangIndexes that is not empty, not the last.

assembl/static2/js/app/components/search.jsx
```diff
@@ -130,26 +131,40 @@ const ImageType = (props) => {
...
   }
-  let body = get(
-    props.result,
-    `highlight.body_${locale}`,
-    get(props.result, `highlight.body_${locale}-x-mtfrom-${mtfromSuffix}`, get(props.result, 'highlight.body_und'))
-  );
+  let body = getFieldAnyLang(props.result.highlight, 'body', locale);
   if (!body) {
-    body = get(source, `body_${locale}`, get(source, `body_${locale}-x-mtfrom-${mtfromSuffix}`, source.body_und));
-    body = truncateText(body);
```
vincentfretin: Why did you remove `body = truncateText(body);`?

assembl/static2/js/app/components/search.jsx
```diff
@@ -252,12 +275,27 @@ const UserHit = (props) => {
...
+  // long title missing?
+  let announceTitle = getFieldAnyLang(props.result.highlight, 'announcement_title', locale);
+  if (!announceTitle) {
+    announceTitle = getFieldAnyLang(source, 'announcement_title', locale);
+  }
+  let announceBody = getFieldAnyLang(props.result.highlight, 'announcement_body', locale);
+  if (!announceBody) {
+    announceBody = getFieldAnyLang(source, 'announcement_body', locale);
+  }
+
```
vincentfretin: Not exactly the same as before. Before, when the text came from source, not source.highlight, I truncated it because it could be too long.

vincentfretin: So in _other you can have several languages? So if you indexed a post japanese and chinese, we will have those two languages in body_other and we show a post in japanese and chinese in the result?

vincentfretin: Ok, so not really an issue, we just ad de, zh, ja in the indexed locales

assembl/static2/js/app/components/search.jsx
maparent: silly error, correcting.

assembl/static2/js/app/components/search.jsx
vincentfretin: Ah, we can't do a return in a forEach, neither a break. Maybe we should just add 'en' at the end of the "elasticsearch_lang_indexes" list in this case so that English is the default?

vincentfretin:
It's okay for the analyzers. At query time, it will use the same analyzer as the one used at indexing time.
https://www.elastic.co/guide/en/elasticsearch/reference/current/search-analyzer.html

assembl/static2/js/app/components/search.jsx
maparent: As discussed: I think it's an improvement over the previous situation, where und was lost. It is robust against mis-identified langstrings. The workaround is to add languages in the index.

assembl/static2/js/app/components/search.jsx
maparent: forEach does not return, it's not a "_.any". After the interface locale is not found, it's a random locale anyway.

maparent: crossed messages. Now, putting en at the end is hackish, so I'll ensure we take the first, at least.

assembl/static2/js/app/components/search.jsx
```diff
@@ -130,27 +131,54 @@ const ImageType = (props) => {
   return <img className={props.className} src={`/static2/img/icon-${props.type}.svg`} alt="" />;
 };
 
+const getFieldAnyLang = (source, prop, locale) => {
+  let result;
+  if (!source) {
+    // Why does this happen?
+    return result;
+  }
+  if (source[`${prop}_${locale}`]) {
```
ayyazdaniaryan: Template strings aren't too performant yet. I would make this a variable and use accordingly.
https://stackoverflow.com/a/29083467

assembl/static2/js/app/components/search.jsx
ayyazdaniaryan: If you wanted to keep this functional, you could have used a fold/reduce function. Result could have been the accumulator. Just a thought...

maparent: reduce could work. do you want me to correct?

production.ini
```diff
@@ -47,6 +47,9 @@ sqlalchemy.strategy = atexit_cleanup
 # Do we use elastic_search indexing
 use_elasticsearch = true
 
+# languages (w/o country) for which we'll have a separate elasticsearch field
+elasticsearch_lang_indexes = en fr
```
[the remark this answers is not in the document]
maparent: production.ini is the base, so it'll end up in local.ini.
You could and probably will override it in some .rc files.

assembl/static2/js/app/components/search.jsx
maparent: Yes, but your eslint settings impose templates. shrug.

vincentfretin: Please keep the template litteral. Performance in browser can change from version to version, and you're arguing for some microseconds here... If this code was in a requestFrameAnimation loop, for VR for example, maybe, and still... And please don't make optimizations based on some stackoverflow comment that can be outdated. Before optimizing something, measure measure measure! ;-) And by the way, the browser is really executing a concatenation here, because babel transform this in ES5. :-D

PR #188, opened 2017-10-04
React share post

MilanBarande:
remake of the share-a-post feature, using the react-share library. The modal is now automaticaly closing once a share button has been clicked. Also added a "copy to clipboard" button using react-copy-to-clipboard


assembl/static2/js/app/components/common/socialShare.jsx
```diff
@@ -0,0 +1,96 @@
+import React from 'react';
+import { CopyToClipboard } from 'react-copy-to-clipboard';
+import { ShareButtons, generateShareIcon } from 'react-share';
+// import { OverlayTrigger } from 'react-bootstrap';
+// import { shareFacebookTooltip } from './tooltips';
```
vincentfretin: please remove the commented lines

assembl/static2/js/app/components/common/tooltips.jsx
```diff
@@ -44,6 +44,12 @@ export const shareTooltip = (
   </Tooltip>
 );
 
+export const shareFacebookTooltip = (
+  <Tooltip id="shareFacebookTooltip">
+    <Translate value="debate.shareSocial" />
+  </Tooltip>
+);
+
```
vincentfretin: remove this code if you don't use it

assembl/static2/js/app/components/common/socialShare.jsx
```diff
@@ -0,0 +1,96 @@
...
+            </div>
+          </div>}
+        <CopyToClipboard
+          text={url}
+          onCopy={() => {
+            return this.setState({ copied: true });
+          }}
+        >
+          <button className="btn btn-default btn-copy">
+            {this.state.copied ? 'Link copied' : 'Copy link to Clipboard'}
```
vincentfretin: please translate this with Ii18n.t

vincentfretin: I don't think it make sense to change the label of the button here. You can add a message 'Link copied' next to the button if you want, but I think the button should keep "Copy link to Clipboard".

MilanBarande: I had done this at first (the message showing after the click) the thing is the modal is automatically growing when you add text inside it and it made a weird effect. So Roxana wants it this way as an MVP and she's creating a ticket with the requirements for something more sophisticated (icons in the button and a tooltip for "link copied" message)

vincentfretin: write a '&nbsp;' maybe so that the place is reserved so the modal won't change size?
Having a button with 'Link copied' that when you click copy the link is a no no for me.

ayyazdaniaryan: You could use a green alert to signal a copy (like giphy), or do as @vincentfretin suggested. Vincent, for the sake of Milan's learning, we should clarify why something is a no no.

assembl/static2/js/app/components/common/socialShare.jsx
```diff
@@ -0,0 +1,95 @@
...
+  WhatsappShareButton,
+  TelegramShareButton
+} = ShareButtons;
+const FacebookIcon = generateShareIcon('facebook');
+const GooglePlusIcon = generateShareIcon('google');
+const LinkedinIcon = generateShareIcon('linkedin');
+const TwitterIcon = generateShareIcon('twitter');
+const EmailIcon = generateShareIcon('email');
+const WhatsappIcon = generateShareIcon('whatsapp');
+const TelegramIcon = generateShareIcon('telegram');
```
ayyazdaniaryan: for this lot, you could do something like this:
```
const icons = {};
['facebook', 'google', 'linkedin', ...].forEach((service) => {
  icons[service] = generateShareIcon(service);
});

File: assembl/static2/js/app/components/common/socialShare.jsx [hunk 5]
```diff
@@ -0,0 +1,95 @@
...
+    };
+  }
+  render() {
+    const { url, onClose, social } = this.props;
+    return (
+      <div className="share-buttons-container">
+        {social
+          ? <div className="social-share-buttons-container">
+            <div className="social-share-button">
+              <EmailShareButton url={url} onShareWindowClose={onClose}>
```
ayyazdaniaryan: Seeing as to how you have to pass the same props everywhere, let's make a Higher Order Component (https://reactjs.org/docs/higher-order-components.html) that abstracts away the passing of these props.
```
const superShareButton = (Component, data) => {
  return (props) => {
    <Component {...data} {...props} />
  }
}
```

And use it with this example:

```
const EmailShare = superShareButton(EmailShareButton, {url: url, onShareWindowClose: onClose, size: 32});
return (<EmailShare />);
```

ayyazdaniaryan: Thought, right now, I'm realizing this doesn't reduce your code. Well, take it as a learning opportunity for HOCs.

assembl/static2/package.json
```diff
@@ -87,13 +87,15 @@
     "react": "^15.4.1",
     "react-apollo": "^1.4.12",
     "react-bootstrap": "^0.31.2",
+    "react-copy-to-clipboard": "^5.0.1",
```
ayyazdaniaryan: Before merging, we'll do another bundle size check to see how much these packages affected the size of our bundle. Hopefully not by much.

assembl/static2/js/app/components/common/socialShare.jsx
vincentfretin: Seriously? This is common sense. It's a button with "link copied" label! what does it means? What does it do when I click on it? A button needs to describe an action, should be a verb.

assembl/static2/js/app/components/common/socialShare.jsx
ayyazdaniaryan: Okay, I was too tired to give a good example of HOC. Here is the simplified solution:
```
const superShareButton = (Component) => {
  const url = 'someUrl';
  const size = 32;
  const onClose = () => console.log('something magical happened');
  const data = {url: url, size: size, onShareWindowClose: onClose}
  return (props) => {
    <Component {...data} {...props} />
  }
}
```

Then you can simply write:

```
const EmailShare = superShareButton(EmailShareButton);
return (<EmailShare />);
...
```

assembl/static2/js/app/components/common/socialShare.jsx
MilanBarande: I don't disagree with you, it's just that Roxana liked it this way so I have left like that in the meantime (until we get her favourite solution which she made a mock up for). I can do what you said but it can be confusing to know what to do when getting different requests...

vincentfretin: Sure, I understand. Let's keep it that way for now.

paolina92: Guys, at the end, it's the PO who decides even if you disagree. Of course you can give your opinion and the most of time the PO will follow it.

swergas:
ok for me

assembl/static2/css/components/socialShare.scss
```diff
@@ -12,16 +12,28 @@
   padding-bottom: 20px;
 }
 
-*:focus {
+.modal-content
+  div:focus {
   outline: none;
```
cedricmessiant: indentation

assembl/static2/css/components/socialShare.scss
```diff
@@ -12,16 +12,28 @@
   padding-bottom: 20px;
 }
 
-*:focus {
+.modal-content
+  div:focus {
```
cedricmessiant: indentation

assembl/static2/css/components/socialShare.scss
```diff
@@ -12,16 +12,28 @@
   padding-bottom: 20px;
 }
 
-*:focus {
+.modal-content
```
cedricmessiant: it will modify all modals, no?

assembl/static2/js/app/components/common/socialShare.jsx
```diff
@@ -12,13 +12,31 @@ const {
...
+
+const FacebookIcon = icons.facebook;
+const GooglePlusIcon = icons.google;
+const LinkedinIcon = icons.linkedin;
+const TwitterIcon = icons.twitter;
+const EmailIcon = icons.email;
+const WhatsappIcon = icons.whatsapp;
+const TelegramIcon = icons.telegram;
+
+const SuperShareButton = ({ Component, Icon, url, onClose, ...props }) => {
```
cedricmessiant: you should anotate new components with flow

cedricmessiant: and add a snapshot test

assembl/static2/css/components/socialShare.scss
MilanBarande: this was put to deal with an undesirable highlighting of divs (different depending on the browser), would we ever want such a thing to happen in a modal ? even anywhere else... this is an annoying behaviour

cedricmessiant: no problem, it was more a remark/"warning"

PR #189, opened 2017-10-04
Nuggets manager to ensure proper (and simpler) nuggets rendering

n0izn0iz:
https://bluenove.atlassian.net/browse/DEVAS-814

### DIFF
--- a/assembl/static2/js/app/components/common/nuggetsManager.js
+++ b/assembl/static2/js/app/components/common/nuggetsManager.js
@@ -30,8 +30,8 @@ class NuggetsManager {
   }
   sort() {
     this.nuggetsList.sort((a, b) => {
-      const aLevel = a.props.completeLevel;
-      const bLevel = b.props.completeLevel;
+      const aLevel = a.props.completeLevel.split('-');
+      const bLevel = b.props.completeLevel.split('-');
       return aLevel.some((aLevelValue, index) => {
         if (index >= bLevel.length) return true;
         return aLevelValue > bLevel[index];
--- a/assembl/static2/js/app/components/common/nuggetsManager.js
+++ b/assembl/static2/js/app/components/common/nuggetsManager.js
@@ -2,7 +2,7 @@
 
 import Nuggets from '../debate/thread/nuggets';
 
-const invalidIndex = (index) => {
+const indexNotFound = (index) => {
   return index === -1;
 };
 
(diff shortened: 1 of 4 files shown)


assembl/static2/js/app/components/common/nuggetsManager.js
```diff
@@ -0,0 +1,43 @@
...
+  constructor() {
+    this.nuggetsList = [];
+  }
+  add(nuggets: Nuggets) {
+    this.nuggetsList.push(nuggets);
+    this.sort();
+    this.update();
+  }
+  update = () => {
+    let previous = null;
```
ayyazdaniaryan: No no. Please don't use mutations like this. You're using a functional approach, keep to it.
```
this.nuggestList.forEach( (nugget, index, nuggets) => {
  if (index == 0) {
    nugget.updateTop(null);
  }
  nugget.updateTop(nuggets[index-1]);
});

File: assembl/static2/js/app/components/common/nuggetsManager.js [hunk 2]
```diff
@@ -0,0 +1,43 @@
...
+    this.nuggetsList.forEach((nuggets) => {
+      nuggets.updateTop(previous);
+      previous = nuggets;
+    });
+  };
+  remove(nuggets: Nuggets) {
+    const index = this.nuggetsList.indexOf(nuggets);
+    if (invalidIndex(index)) throw new Error('Tried to remove nuggets that are not managed by this NuggetsManager');
```
ayyazdaniaryan: Let the error include the index that caused the error.

assembl/static2/js/app/components/common/nuggetsManager.js
```diff
@@ -0,0 +1,43 @@
+// @flow
+
+import Nuggets from '../debate/thread/nuggets';
+
+const invalidIndex = (index) => {
+  return index === -1;
```
ayyazdaniaryan: actually, an invalid index is `index < 0`. But, I understand this case will never be hit in this PR.

assembl/static2/js/app/components/debate/thread/post.jsx
```diff
@@ -194,7 +205,19 @@ class Post extends React.PureComponent {
...
+          postId={id}
+          nuggetsManager={nuggetsManager}
+          completeLevel={(fullLevel
+            ? [
+              rowIndex,
+              ...fullLevel.split('-').map((string) => {
+                return Number(string);
+              })
+            ]
+            : [rowIndex]).join('-')}
```
ayyazdaniaryan: This entire if block is next to unreadable for a developer not very familiar with this part of the code. For the sake of clarity, can it be written outside of the Component creation as a variable and cleaned up?

assembl/static2/js/app/components/common/nuggetsManager.js
n0izn0iz: from MDN
>Array.prototype.indexOf()
>....
>Return value
>The first index of the element in the array; -1 if not found.

ayyazdaniaryan: I understand. The method name implies it checks all cases of invalid indices, not just the returned value of `indexOf`. Perhaps change the method name?

n0izn0iz: yup

PR #190, opened 2017-10-04
Adapt post for n-columns

n0izn0iz:
https://bluenove.atlassian.net/browse/DEVAS-827


assembl/static2/css/components/posts.scss
```diff
@@ -532,3 +532,12 @@ $deleted-post-border: 1px solid $grey1 !default;
         }
     }
 }
+
+.column-hint {
+  font-family: $font-family4;
+  font-size: $font-size-m;
+  font-weight: 300;
```
paolina92: You should use the dedicated font-family and don't use the font-weight property

PR #193, opened 2017-10-10
Graphql twocolumns

ayyazdaniaryan:
GraphQL queries to be consumed by the front-end


assembl/graphql/schema.py
```diff
@@ -546,7 +546,7 @@ class Post(SecureObjectType, SQLAlchemyObjectType):
     class Meta:
         model = models.Post
         interfaces = (Node, PostInterface)
-        only_fields = ('id', )  # inherits fields from Post interface only
+        only_fields = ('id', 'message_classifier')  # inherits fields from Post interface only
```
vincentfretin: You can remove message_classifier here. It is said in the comment that it inherits the fields from PostInterface.

PR #195, opened 2017-10-10
Trying travis integration (status: code review)

swergas:


swergas:
Github does still not show the last 2 commits on this page, they are here: https://github.com/assembl/assembl/compare/travis

.travis.yml
```diff
@@ -1,12 +1,50 @@
...
+  #- fab -c configs/travis.rc database_delete
+  #- fab -c configs/travis.rc app_update_dependencies
+  #- fab -c configs/travis.rc app_compile_nodbupdate
+  - fab -c configs/travis.rc database_create
+  - assembl-db-manage testing.ini bootstrap
+script:
+  # TODO: fix error (production.ini: Bad value substitution)
+  # fab -c configs/travis.rc build_doc
+  - py.test --cov assembl --cov-report term --cov-report html --junit-xml junit.xml assembl
+  - cd assembl/static2; npm test; npm run eslint; cd ..
```
vincentfretin: You can add `npm run flow` too. `cd ..` is really needed here?
Maybe you can run in parallel `py.test` and `npm test; npm run eslint, npm run flow` with concurrent jobs?

swergas: OK thanks for the remark, I've added `npm run flow` and remove the final `cd ..`. I think further optimizations of the build (parallelizing tasks and caching an installed version of assembl) is out of scope for this first implementation and will need a second prioritized task.

swergas:
Any other remark @ayyazdaniaryan or @maparent ? Otherwise it's ready for merge

.travis.yml
```diff
@@ -1,12 +1,50 @@
...
- - sudo apt-get -qq update
+ # The following line is needed, otherwise the set_file_permissions fab command yells error "Operation not permitted"
+ - sudo usermod -a -G www-data $USER
  - sudo apt-get install -y fabric git openssh-server
- - fab -c configs/develop.rc install_single_server
- - fab -c configs/develop.rc bootstrap_from_checkout
-install: true
+ - ssh-keygen -P '' -f ~/.ssh/id_rsa && cat ~/.ssh/id_rsa.pub >> ~/.ssh/authorized_keys
+ - echo "_extends = ./testing.rc" > configs/travis.rc; echo "_gitbranch = $TRAVIS_COMMIT" >> configs/travis.rc;
+ # The following line produces warning "uses weak digest algorithm" during its `apt-get update`
```
ayyazdaniaryan: Likely coming from one of the dependencies being stale/old?
https://askubuntu.com/questions/760796/how-to-fix-apt-signature-by-key-uses-weak-digest-algorithm-sha1

.travis.yml
```diff
@@ -1,12 +1,50 @@
...
+ - fab -c configs/travis.rc install_assembl_server_deps
+ - fab -c configs/travis.rc install_redis
+ - fab -c configs/travis.rc install_memcached
+install:
+ # The set_file_permissions fabric command needs to be encapsulated into this sudo su command, because in before_install we have added $USER in www-data group but have not logged out and logged in again for changes to take effect in this terminal
+ - sudo -E su $USER -c 'fab -c configs/travis.rc set_file_permissions'
+ # TODO: fix fabfile's check_and_create_database_user command which produces an error when run by travis. For now we instead run the following command
+ - pip install psycopg2
+ - python2 $PWD/assembl/scripts/pypsql.py -u postgres -n localhost  "CREATE USER assembl_test WITH CREATEDB ENCRYPTED PASSWORD 'assembl_test'; COMMIT;"
```
ayyazdaniaryan: ooh boy, this isn't good. Ahh, is it failing because the script `assembl-pypsql` doesn't exist yet?
If so, prior to running this, you can run:
```
pip install -e $PWD
```

And that would install the setup.py scripts in "editable mode", giving you access to them. 

https://pip.pypa.io/en/stable/reference/pip_install/#editable-installs

ayyazdaniaryan: Maybe could resolve this issue.

.travis.yml
```diff
@@ -1,12 +1,50 @@
...
+  #- supervisord
+  #- supervisorctl shutdown
+  #- fab -c configs/travis.rc database_delete
+  #- fab -c configs/travis.rc app_update_dependencies
+  #- fab -c configs/travis.rc app_compile_nodbupdate
+  - fab -c configs/travis.rc database_create
+  - assembl-db-manage testing.ini bootstrap
+script:
+  # TODO: fix error (production.ini: Bad value substitution)
+  # fab -c configs/travis.rc build_doc
```
ayyazdaniaryan: Nah, not needed in this environment.

.travis.yml
```diff
@@ -1,12 +1,50 @@
...
+  - py.test --cov assembl --cov-report term --cov-report html --junit-xml junit.xml assembl
```
ayyazdaniaryan: Does travis account build success/fail based on whether the scripts output into the stderr?

vincentfretin: stderr doesn't matter, it's the returned code that matter.

.travis.yml
```diff
@@ -1,12 +1,50 @@
...
+ - echo "_extends = ./testing.rc" > configs/travis.rc; echo "_gitbranch = $TRAVIS_COMMIT" >> configs/travis.rc;
+ # The following line produces warning "uses weak digest algorithm" during its `apt-get update`
+ - fab -c configs/travis.rc install_assembl_server_deps
+ - fab -c configs/travis.rc install_redis
+ - fab -c configs/travis.rc install_memcached
+install:
+ # The set_file_permissions fabric command needs to be encapsulated into this sudo su command, because in before_install we have added $USER in www-data group but have not logged out and logged in again for changes to take effect in this terminal
+ - sudo -E su $USER -c 'fab -c configs/travis.rc set_file_permissions'
+ # TODO: fix fabfile's check_and_create_database_user command which produces an error when run by travis. For now we instead run the following command
+ - pip install psycopg2
```
maparent: That and a few other things could be part of install_basetools
I guess you decided you did not need install_single_server ? So no testing elasticsearch?

vincentfretin: There is no elasticsearch tests currently.

.travis.yml
maparent: That may not be necessary. We install psycopg2 globally so the script can be run before the whole assembl machinery. `python assembl/scripts/pypsql.py ...` should work.

vincentfretin:
follow-up commit to fix return code of npm test; npm run eslint that was ignored. Travis would be green if there was errors in eslint or jest and no error on flow. And I split the tests into BACKEND and FRONTEND.

https://github.com/assembl/assembl/commit/15714f3ff1484817ae5901101662cf77010c5a09

PR #196, opened 2017-10-11
Search in v2

vincentfretin:
without responsive


vincentfretin:
It's ready for merge as it is.

assembl/static2/js/app/components/search.jsx
```diff
@@ -91,19 +91,33 @@ if (__resourceQuery) {
...
     default:
       // post
-      return `posts/${id}`;
+      ideaId = hit._source.idea_id.length > 0 ? hit._source.idea_id[0] : null;
+      if (!ideaId) {
+        return undefined;
+      }
+      ideaBase64id = btoa(`Idea:${ideaId}`);
+      postBase64id = btoa(`Post:${id}`);
+      return `/${slug}/debate/thread/theme/${ideaBase64id}/#${postBase64id}`;
```
cedricmessiant: don't you have an inconsistent return error in eslint ?

assembl/static2/js/app/components/search.jsx
```diff
@@ -91,19 +91,33 @@ if (__resourceQuery) {
...
     const id = hit._source.id;
+    let ideaBase64id;
+    let postBase64id;
+    let ideaId;
     switch (hit._type) {
     case 'synthesis':
-      return `posts/${id}`;
+      return undefined;
     case 'user':
-      return `profile/${id}`;
```
cedricmessiant: those 2 cases were copy/paste from v1 ?

cedricmessiant:
LGTM

assembl/static2/js/app/components/search.jsx
vincentfretin: nope

assembl/static2/js/app/components/search.jsx
vincentfretin: yep, this was when we didn't even had v2 routes

fabfile.py
```diff
@@ -1818,6 +1818,9 @@ def install_elasticsearch():
         move(join(extract_path, 'elasticsearch-{version}'.format(version=ELASTICSEARCH_VERSION)), extract_path+'.tmp')
         rmtree(extract_path)
         move(extract_path+'.tmp', extract_path)
+        run(env.projectpath + '/var/elasticsearch/bin/elasticsearch-plugin install https://artifacts.elastic.co/downloads/elasticsearch-plugins/analysis-smartcn/analysis-smartcn-{version}.zip'.format(version=ELASTICSEARCH_VERSION))
+        run(env.projectpath + '/var/elasticsearch/bin/elasticsearch-plugin install https://artifacts.elastic.co/downloads/elasticsearch-plugins/analysis-kuromoji/analysis-kuromoji-{version}.zip'.format(version=ELASTICSEARCH_VERSION))
```
ayyazdaniaryan: Cool. For the automation process however, I would have this method return a boolean for success/fail so that later tasks can bail out. Eg. somewhere along installation process, network failed, didn't install, etc. Possible?

assembl/static2/js/app/components/search.jsx
```diff
@@ -98,19 +91,33 @@ if (__resourceQuery) {
...
+      return undefined;
     case 'idea':
-      return `ideas/${id}`;
+      ideaBase64id = btoa(`Idea:${id}`);
```
ayyazdaniaryan: Learn something new everyday. `btoa` 👍

assembl/indexing/utils.py
```diff
@@ -58,11 +59,16 @@ def populate_from_langstring(ls, data, dataPropName):
             if not entry.value:
                 continue
             locale_code = strip_country(entry.base_locale)
+            if locale_code == 'zh_Hans':
+                locale_code = 'zh_CN'
             if locale_code in langs:
                 dataPropNameL = "_".join((dataPropName, locale_code))
-                data[dataPropNameL] = entry.value
+                # Japanese for example is stored in db with a html entity for
```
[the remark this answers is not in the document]
vincentfretin: Other lang can have entities, an apostrophe can be an entity too. ASCII of course is not represented as entities.

assembl/scripts/reindex_all_contents.py
```diff
@@ -21,11 +20,11 @@ def main():
         "configuration",
         help="configuration file with destination database configuration")
     args = parser.parse_args()
+    env = bootstrap(args.configuration)
```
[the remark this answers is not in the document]
vincentfretin: Yeah, apparently this configure the registry an put it in threadlocal.

assembl/static2/js/app/components/search.jsx
```diff
@@ -98,19 +91,33 @@ if (__resourceQuery) {
   };
   // }
 } else {
-  Link = require('react-router').Link; // eslint-disable-line
+  // Link = require('react-router').Link; // eslint-disable-line
+  Link = (props) => {
+    return <a href={props.to} dangerouslySetInnerHTML={props.dangerouslySetInnerHTML} />;
```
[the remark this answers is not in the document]
vincentfretin: Ah, yeah forgot about that one. Yes, you can use react-router, but the page load in the background, you need a way to close the search modal after clicking on a link. I didn't find a clear way to do that, so I don't use react-router and do a full page load.

fabfile.py
vincentfretin: This is not automatic with fabric? It doesn't stop if the command return an error?

assembl/indexing/settings.py
```diff
@@ -81,6 +81,7 @@ def add_index_languages(props, names):
         'creation_date': DATE,
         'id': LONG,
         'parent_id': {'type': 'long', 'null_value': 0},
+        'idea_id': LONG,
```
swergas: A message can be associated to several ideas, so type should probably not be LONG but an array of LONG, no? And so idea_id would not be the best name, rather something like ideas_ids

assembl/indexing/utils.py
```diff
@@ -120,6 +120,9 @@ def get_data(content):
                      'creator_id', 'sentiment_counts'):
             data[attr] = getattr(content, attr)
 
+        data['idea_id'] = [link.idea_id
```
swergas: A message can be associated to several ideas, so idea_id would not be the best name, rather something like ideas_ids, no?

assembl/indexing/settings.py
vincentfretin: With the elasticsearch mapping syntax, there is no array per se. Declaring here idea_id as long can contain long or an array of long.

assembl/indexing/utils.py
vincentfretin: Yeah, on the db model side. In v2, a message can only be associated to one idea though. The purpose of indexing idea_id here is to generate an url to the post, so I really only need one idea in the end.

swergas: Maybe it can be only associated to one idea for now in v2, but that is temporary, as harvesting is a core feature of assembl, it will arrive sooner or later in v2

PR #198, opened 2017-10-13
fix running_locally when a public_hostname is set to prevent the...

n0izn0iz:
… env.mac variable from being set to False


fabfile.py
```diff
@@ -103,6 +102,8 @@ def sanitize_env():
     # nor env.host_string are set properly. Revisit with Fabric2.
     if not env.get('host_string', None):
         env.host_string = env.hosts[0]
+
```
ayyazdaniaryan: ```if not env.hostname ... ```

PR #199, opened 2017-10-13
Attachment image prev styling

MilanBarande:
The modal is now adapting to properly display an image preview in case the uploaded file is an image. If it's any other format of file it will just display the file name and keep its original size.

### DIFF
--- a/assembl/static2/js/app/components/common/attachFileForm.jsx
+++ b/assembl/static2/js/app/components/common/attachFileForm.jsx
@@ -36,12 +36,18 @@ class AttachFileForm extends React.Component<*, AttachFileFormProps, AttachFileF
   };
 
   render() {
+    let isImage;
+    const file = this.state.file;
+    if (file !== null) {
+      isImage = file.type.includes('image');
+    }
+
     return (
-      <div className="attach-file">
+      <div className={isImage ? 'attach-file with-preview' : 'attach-file'}>
         <label htmlFor="attachment">
           <Translate value="common.attachFileForm.label" />
         </label>
-        <FileUploader handleChange={this.handleFileChange} fileOrUrl={this.state.file} withPreview={false} />
+        <FileUploader handleChange={this.handleFileChange} fileOrUrl={this.state.file} isImage={isImage} withPreview />
         <Button className="button-submit button-dark btn btn-default" onClick={this.handleSubmit}>
           <Translate value="common.attachFileForm.submit" />
         </Button>
--- a/assembl/static2/js/app/components/common/fileUploader.jsx
+++ b/assembl/static2/js/app/components/common/fileUploader.jsx
@@ -67,7 +67,7 @@ class FileUploader extends React.Component {
   render() {
     const fileSrc = this.state.fileSrc;
     const fileIsImage = this.props.isImage;
-    const urlIsImage = fileSrc && fileSrc.includes('png' || 'jpeg' || 'jpg' || 'gif');
+    const urlIsImage = fileSrc && fileSrc.endsWith('png' || 'jpeg' || 'jpg' || 'gif');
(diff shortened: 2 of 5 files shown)


assembl/static2/js/app/components/common/fileUploader.jsx
```diff
@@ -65,6 +65,9 @@ class FileUploader extends React.Component {
   }
 
   render() {
+    const fileSrc = this.state.fileSrc;
+    const fileIsImage = this.props.isImage;
+    const urlIsImage = fileSrc && fileSrc.includes('png' || 'jpeg' || 'jpg' || 'gif');
```
ayyazdaniaryan: This is not bad, but there are other file types that can be images. You can make a better estimation of the type by checking the `File` object's `type`. It will often be `application/<type>`. 

https://developer.mozilla.org/en-US/docs/Web/API/File 
If there is a failing test on this, we can work on it together.

assembl/static2/js/app/components/common/attachFileForm.jsx
```diff
@@ -36,12 +36,18 @@ class AttachFileForm extends React.Component<*, AttachFileFormProps, AttachFileF
...
   render() {
+    let isImage;
+    const file = this.state.file;
+    if (file !== null) {
+      isImage = file.type.includes('image');
+    }
+
     return (
-      <div className="attach-file">
+      <div className={isImage ? 'attach-file with-preview' : 'attach-file'}>
```
cedricmessiant: it is a good habit to use `classnames` lib for this kind of stuff

assembl/static2/js/app/components/common/fileUploader.jsx
```diff
@@ -73,17 +76,18 @@ class FileUploader extends React.Component {
         {this.props.withPreview
           ? <div className={this.state.fileSrc ? 'preview' : 'hidden'}>
             <img
-              src={this.state.fileSrc}
+              src={fileSrc}
               ref={(p) => {
                 this.preview = p;
               }}
-              alt="preview"
+              alt={fileIsImage || urlIsImage ? 'preview' : ''}
```
cedricmessiant: we shouldn't render an img tag if this is not an image, aren't we ?

assembl/static2/js/app/components/common/attachFileForm.jsx
```diff
@@ -36,12 +36,18 @@ class AttachFileForm extends React.Component<*, AttachFileFormProps, AttachFileF
   };
 
   render() {
+    let isImage;
+    const file = this.state.file;
+    if (file !== null) {
+      isImage = file.type.includes('image');
```
cedricmessiant: ok, you can also use `.startsWith` but I think it doesn't matter here

assembl/static2/js/app/components/common/fileUploader.jsx
MilanBarande: this was made to satisfy a test (hence the comment made by Aryan up top), "preview" was displayed next to the file title when the uploaded file was not an image. The test mocks an update using an image url, so the condition isImage was false because it only checks if the file type is "image", which does not work with url. So I will get rid of this and update the test instead

assembl/static2/js/app/components/common/attachFileForm.jsx
MilanBarande: Allright, I thought the classnames lib was only pertinent when we had more than 2 classes alternating

assembl/static2/js/app/components/common/fileUploader.jsx
MilanBarande: answered this issue in my answer to Cédric, I will take care of this by updating the fileUploader test

PR #202, opened 2017-10-17
Synthesis share

MilanBarande:


assembl/static2/js/app/components/debate/multiColumns/columnsView.jsx
```diff
@@ -53,13 +64,30 @@ class ColumnsView extends React.Component {
       return countSynthesis >= 1;
     });
 
+    const { slug, phase, themeId } = routerParams;
+    const openShareSynthesisModal = (id) => {
+      const title = <Translate value="debate.shareSynthesis" />;
+      const url = `${window.location.protocol}//${window.location.host}${get('debate', {
```
ayyazdaniaryan: Try to use the react-router props provided from the highest level component to here. It's just a practice, not a hard-set rule.

assembl/static2/js/app/components/debate/multiColumns/columnsView.jsx
```diff
@@ -53,13 +64,30 @@ class ColumnsView extends React.Component {
       return countSynthesis >= 1;
     });
 
+    const { slug, phase, themeId } = routerParams;
+    const openShareSynthesisModal = (id) => {
```
ayyazdaniaryan: There was a method to `openSharePostModal`. Now you're adding specific method for this. Looking quickly, there isn't much of a difference between the two. This is often a good practice to abstract this operation to a method that is independent of the type, perhaps put it on the highest level component, or as a function in the `utils` or pass it as a prop.

assembl/static2/js/app/components/debate/multiColumns/columnsView.jsx
```diff
@@ -70,7 +98,7 @@ class ColumnsView extends React.Component {
                     style={isColumnViewInline ? { width: `${MIN_WIDTH_COLUMN}px` } : {}}
                   >
                     {isSynthesis
-                      ? <div id={`synthesis-${classifier}`} className="box synthesis">
+                      ? <div id={synthesisId} className="box synthesis">
```
ayyazdaniaryan: Why does the Synthesis have an `id` instead of a `class`?

assembl/static2/js/app/components/debate/multiColumns/columnsView.jsx
ayyazdaniaryan: I mean that the router has `location` as well. You could perhaps pass that in as well.

assembl/static2/js/app/components/debate/multiColumns/columnsView.jsx
MilanBarande: to be able to scroll down to it

assembl/static2/js/app/components/debate/multiColumns/columnsView.jsx
MilanBarande: okay, I just rename both openShareModal ?

ayyazdaniaryan: I'm referring to writing one method and using it across all components. If that is what you're referring to, then yes an `openShareModal`.

assembl/static2/js/app/components/debate/multiColumns/columnsView.jsx
ayyazdaniaryan: Ah okay.

assembl/static2/js/app/components/debate/multiColumns/columnsView.jsx
MilanBarande: Done.

assembl/static2/js/app/utils/utilityManager.jsx
```diff
@@ -72,6 +74,21 @@ export const closeModal = () => {
   modalManager.component.setState({ showModal: false });
 };
 
+export const openShareModal = (title, routerParams, elementId, social, isFooter, footer) => {
```
ayyazdaniaryan: As these entities are rather long and in each case not needed, pass an `option` object around, and take the appropriate keys in the method below. Would be easier :)

assembl/static2/js/app/utils/utilityManager.jsx
```diff
@@ -72,6 +74,21 @@ export const closeModal = () => {
...
+  const modalTitle = title;
+  const { slug, phase, themeId } = routerParams;
+  const url = `${window.location.protocol}//${window.location.host}${get('debate', {
```
ayyazdaniaryan: Next step here, in `routeMap`, you can add a route for thematic element, and instead use the `getFullPath` method, whilst only passing the keys needed.

assembl/static2/js/app/components/debate/common/postActions.jsx
```diff
@@ -106,10 +91,18 @@ class PostActions extends React.Component {
...
+                title: modalTitle,
+                routerParams: routerParams,
+                elementId: postId,
+                social: useSocial
+              });
             }}
           >
-            <OverlayTrigger placement={this.state.screenWidth >= MEDIUM_SCREEN_WIDTH ? 'right' : 'top'} overlay={shareTooltip}>
+            <OverlayTrigger
+              placement={this.state.screenWidth >= MEDIUM_SCREEN_WIDTH ? 'right' : 'top'}
```
ayyazdaniaryan: This comment isn't pertinent to just your PR, but to all use cases of places. This check isn't complete enough. It should be expanded to check for the item's `width` and `scrollLeft` properties against the screen width, and placed accordingly.

This would resolve issues such as https://bluenove.atlassian.net/browse/DEVAS-888

ayyazdaniaryan: You do not have to resolve it in this PR. But, be aware that to resolve that bug, these are the kinds of checks that you should be making for tooltip placement

assembl/static2/js/app/utils/routeMap.js
```diff
@@ -35,7 +35,8 @@ const routes = {
   theme: 'theme/${themeId}',
   administration: '${slug}/administration',
   unauthorizedAdministration: '${slug}/unauthorizedAdministration',
-  adminPhase: '${phase}'
+  adminPhase: '${phase}',
+  post: '${slug}/debate/${phase}/theme/${themeId}/#${element}'
```
ayyazdaniaryan: Maybe to be more clear, you could rename `element` to `elementId` to be consistent with the `themeId` that is just before it.

assembl/static2/js/app/utils/utilityManager.jsx
cedricmessiant: I agree

assembl/static2/js/app/utils/utilityManager.jsx
cedricmessiant: yes, try to avoid the use of `window` in components for purity/testabilty

assembl/static2/js/app/utils/utilityManager.jsx
```diff
@@ -72,6 +74,21 @@ export const closeModal = () => {
...
+    slug: slug,
+    phase: phase
+  })}${get('theme', {
+    themeId: themeId
+  })}/#${elementId}`;
+  const modalBody = <SocialShare url={url} onClose={closeModal} social={social} />;
+  const isModalFooter = isFooter;
```
cedricmessiant: these 2 `const` are really not useful

assembl/static2/js/app/utils/utilityManager.jsx
```diff
@@ -72,6 +74,21 @@ export const closeModal = () => {
   modalManager.component.setState({ showModal: false });
 };
 
+export const openShareModal = (title, routerParams, elementId, social, isFooter, footer) => {
+  const modalTitle = title;
```
cedricmessiant: it is the same for this `const`

assembl/static2/js/app/utils/utilityManager.jsx
```diff
@@ -84,9 +84,7 @@ export const openShareModal = (title, routerParams, elementId, social, isFooter,
     themeId: themeId
   })}/#${elementId}`;
   const modalBody = <SocialShare url={url} onClose={closeModal} social={social} />;
-  const isModalFooter = isFooter;
-  const modalFooter = footer;
-  return displayModal(modalTitle, modalBody, isModalFooter, modalFooter);
+  return displayModal(title, modalBody, isFooter, footer);
```
cedricmessiant: just a remark here, `displayModal` would benefit of an `options` param too (and its `footer` boolean param is kind of redundant with `footerTxt`)

ayyazdaniaryan: I agree on this point. Allows for the display modal to be more modular.

PR #203, opened 2017-10-19
set min-height instead of fixed height for the form

MilanBarande:


MilanBarande:
shouldn't have made a branch for this, I thought there would be more work, I merge it

PR #204, opened 2017-10-19
remove global state from tree.jsx (fix for DEVAS-891)

n0izn0iz:
https://bluenove.atlassian.net/browse/DEVAS-891

+ Moved prevStopIndex global in Tree
+ Nuggets use events with the tree root div as dom element instead of document

### DIFF
--- a/assembl/static2/js/app/components/common/tree.jsx
+++ b/assembl/static2/js/app/components/common/tree.jsx
@@ -5,25 +5,6 @@ import { AutoSizer, CellMeasurer, CellMeasurerCache, List, WindowScroller } from
 import { getDomElementOffset, createEvent } from '../../utils/globalFunctions';
 import NuggetsManager from './nuggetsManager';
 
-let prevStopIndex = 0;
-// override overscanIndicesGetter to not remove from the dom the posts once rendered
-// to fix various issue with scrolling with WindowScroller
-function overscanIndicesGetter({ cellCount, overscanCellsCount, stopIndex }) {
-  let overscanStopIndex;
-  if (cellCount === 1) {
-    // overscanIndicesGetter is called for columns, not rows
-    // use default implementation
-    overscanStopIndex = Math.min(cellCount - 1, stopIndex + overscanCellsCount);
-  } else {
-    prevStopIndex = Math.max(prevStopIndex, stopIndex);
-    overscanStopIndex = Math.min(cellCount - 1, prevStopIndex + overscanCellsCount);
-  }
-  return {
-    overscanStartIndex: 0,
-    overscanStopIndex: overscanStopIndex
-  };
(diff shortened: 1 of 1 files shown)


assembl/static2/js/app/components/common/tree.jsx
```diff
@@ -280,13 +288,12 @@ class Tree extends React.Component {
...
                 this.cache.clearAll();
                 this.listRef.recomputeRowHeights();
-                document.dispatchEvent(rowHeightRecomputed);
+                if (this.rootRef) this.rootRef.dispatchEvent(rowHeightRecomputed);
               }}
             >
               {({ width }) => {
                 return (
                   <List
-                    key={data.length}
```
cedricmessiant: why did you removed the key on `List`? (it may leads to strange bugs)

n0izn0iz: woops, did not meant to remove it. Because data.length is not a key? what if two List have the same length ? why not just put 0 ?

cedricmessiant: it is indeed a strange value for a key. In case of 2 columns we could have 2 List with the same length so it may not be a good idea. I think that it was originally meant to ensure that if the size of the List changes, we re-render it.

assembl/static2/js/app/components/common/tree.jsx
```diff
@@ -210,9 +199,9 @@ class Tree extends React.Component {
     // Reset the global prevStopIndex to not overfetch posts when changing idea
     // or to avoid recreating all dom nodes if we go back to the same idea.
     this.cache.clearAll();
-    prevStopIndex = 0;
+    this.prevStopIndex = 0;
 
-    document.addEventListener('rowHeightRecomputed', this.nuggetsManager.update);
+    if (this.rootRef) this.rootRef.addEventListener('rowHeightRecomputed', this.nuggetsManager.update);
```
ayyazdaniaryan: This part was a bit confusing to me. But I understand now, you don't want a higher level DOM object consuming events from all columns. Each column throws their own event.

More importantly, we *should* pass a prop to this component that enables/disables nugget management on the tree. In case of columns, there will never be nuggets. Therefore you're adding event listener and throwing events for no gain. The component can be more efficient.

assembl/static2/js/app/components/common/tree.jsx
n0izn0iz: but the rowCount prop is already data.length

n0izn0iz: So I just let the key removed right ?
The List is not an element of an array so we shouldn't need the key prop and data.length is passed as an other prop so the rerender behaviour does not change

PR #206, opened 2017-10-24
place post tooltips on left

MilanBarande:
This was done to solve https://bluenove.atlassian.net/browse/DEVAS-888 among other bugs

### DIFF
--- a/assembl/static2/js/app/components/debate/common/postActions.jsx
+++ b/assembl/static2/js/app/components/debate/common/postActions.jsx
@@ -76,13 +76,14 @@ class PostActions extends React.Component {
       return displayModal(title, body, footer, footerTxt);
     };
     let overflowMenu = null;
+    const tooltipPlacement = this.state.screenWidth >= MEDIUM_SCREEN_WIDTH ? 'left' : 'top';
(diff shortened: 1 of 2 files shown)


assembl/static2/js/app/components/debate/common/postActions.jsx
```diff
@@ -61,13 +61,14 @@ class PostActions extends React.Component {
     const modalTitle = <Translate value="debate.sharePost" />;
     const useSocial = debateData.useSocialMedia;
     let overflowMenu = null;
+    const tooltipPlacement = this.state.screenWidth >= MEDIUM_SCREEN_WIDTH ? 'left' : 'top';
```
ayyazdaniaryan: Ahh, this may have strange consequences for the stats hovering on all sentiments. On a smaller screen, you're going to show a big stats box on "top" of what you were reading above. Don't think it's a dev problem at all, but maybe a UX one.

MilanBarande: I didn't notice any shocking problem of this kind when testing. But this is not new anyway, the tooltip was always on top for small screen

PR #207, opened 2017-10-24
Fix image previews in admin section

cedricmessiant:


cedricmessiant:
I'll continue the refactoring of `{ img { externalUrl, mimeType } }` on another branch that will be merge after MEP

PR #210, opened 2017-10-30
Fix landing page with invalid timeline

n0izn0iz:
https://bluenove.atlassian.net/browse/DEVAS-934

### DIFF
--- a/assembl/static2/js/app/main.jsx
+++ b/assembl/static2/js/app/main.jsx
@@ -1,3 +1,5 @@
+// @flow
+
 import React from 'react';
 import { connect } from 'react-redux';
 import { getCurrentPhaseIdentifier } from './utils/timeline';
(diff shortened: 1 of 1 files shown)


assembl/static2/js/app/main.jsx
```diff
@@ -27,13 +30,10 @@ class Main extends React.Component {
       const currentPhase = debateData.timeline.filter((phase) => {
         return phase.identifier === currentPhaseIdentifier;
       });
-      isRedirectionToV1 = currentPhase[0].interface_v1;
+      isRedirectionToV1 = currentPhase[0] && currentPhase[0].interface_v1;
     }
     this.props.addRedirectionToV1(isRedirectionToV1);
   }
-  componentDidMount() {
-    window.addEventListener('scroll', this.displayHeader);
```
ayyazdaniaryan: This isn't relevant any longer?

n0izn0iz: this.displayHeader is undefined I can't see where it could be.

n0izn0iz: there is no "displayHeader" string in the project code except here

ayyazdaniaryan: Okay, dead code. Good kill.

PR #213, opened 2017-10-31
Transform url into links an youtube videos in posts body

cedricmessiant:
for youtube videos it is a quickfix that should probably be replaced by something more generic. I took a look at https://www.npmjs.com/package/oembed-all but it does not seem easy to integrate in our use case (we have raw html as input). 
I think that when we will add links management to the rich text editor, we will not need this anymore as we should probably create the good HTML markup from draft-js entities.

### DIFF
--- a/assembl/static2/js/app/utils/linkify.js
+++ b/assembl/static2/js/app/utils/linkify.js
@@ -1,0 +1,39 @@
+// @flow
+import linkifyHtml from 'linkifyjs/html';
+import * as linkify from 'linkifyjs';
+
+export function transformLinksInHtml(html: string): string {
+  const linksToReplace = linkify
+    .find(html)
+    .map((link) => {
+      const url = link.href;
+      if (url.startsWith('https://www.youtube.com') || url.startsWith('https://youtube.com')) {
+        const re = /(.*watch\?v=(.*?))([\s<,;].*)/;
+        const result = url.match(re);
+        if (result) {
+          const videoUrl = result[1];
+          const videoId = result[2];
+          const rest = result[3];
+          const embedUrl = `https://www.youtube.com/embed/${videoId}`;
+          const embeddedIFrame = `<div><iframe title="" src="${embedUrl}" frameborder="0" allowfullscreen></iframe></div>`;
+          return {
+            origin: url,
+            dest: videoUrl + embeddedIFrame + rest
+          };
+        }
+      }
+
+      return null;
+    })
+    // remove null values
+    .filter((link) => {
+      return link;
+    });
+
+  let transformedHtml = html;
+  linksToReplace.forEach((linkToReplace) => {
+    transformedHtml = html.replace(linkToReplace.origin, linkToReplace.dest);
+  });
+
+  return linkifyHtml(transformedHtml);
+}
--- a/assembl/static2/js/app/utils/linkify.js
+++ b/assembl/static2/js/app/utils/linkify.js
@@ -2,36 +2,52 @@
 import linkifyHtml from 'linkifyjs/html';
 import * as linkify from 'linkifyjs';
 
+type LinkToReplace = {
+  dest: string,
+  origin: string
+};
+
+type LinkifyLink = {
+  href: string,
+  type: string,
+  value: string
+};
+
 export function transformLinksInHtml(html: string): string {
-  const linksToReplace = linkify
+  const linksToReplace: Array<LinkToReplace> = linkify
     .find(html)
-    .map((link) => {
+    .map((link: LinkifyLink) => {
       const url = link.href;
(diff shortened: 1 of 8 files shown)


assembl/static2/package.json
```diff
@@ -82,6 +82,7 @@
     "immutable": "~3.7.4",
     "isomorphic-fetch": "^2.2.1",
     "json-loader": "^0.5.4",
+    "linkifyjs": "^2.1.5",
```
ayyazdaniaryan: Does this pull in all of the dependencies, including the react module, jquery module, html, and others?

ayyazdaniaryan: Hmm, no doesn't pull jQuery dependency in (thank god). http://soapbox.github.io/linkifyjs/docs/2.1.html

ayyazdaniaryan: But, possible to just install `linkify-html` standalone?

assembl/static2/js/app/utils/linkify.js
```diff
@@ -0,0 +1,39 @@
...
+      }
+
+      return null;
+    })
+    // remove null values
+    .filter((link) => {
+      return link;
+    });
+
+  let transformedHtml = html;
```
ayyazdaniaryan: Shallow copy. You're mutating the content of the input here, I believe. Is that the behaviour that you want?

assembl/static2/js/app/utils/linkify.js
```diff
@@ -0,0 +1,39 @@
...
+import linkifyHtml from 'linkifyjs/html';
+import * as linkify from 'linkifyjs';
+
+export function transformLinksInHtml(html: string): string {
+  const linksToReplace = linkify
+    .find(html)
+    .map((link) => {
+      const url = link.href;
+      if (url.startsWith('https://www.youtube.com') || url.startsWith('https://youtube.com')) {
+        const re = /(.*watch\?v=(.*?))([\s<,;].*)/;
```
ayyazdaniaryan: Could you explain this group further to me? `([\s<,;].*)`?

ayyazdaniaryan:
@cedricmessiant You know what I just realized, we kind of took a shortcut approach to this, to make every post calculate their links on render. For many posts, with multiple URLs, this can be expensive. Not to mention that the YouTube injection is limiting (generalizing this PR will be hard).

I'm not sure if it is in your scope, or worth your time, but wouldn't the better solution be capture this once, at creation time? You'll have access to the `EditorState`, so you can use the linkify results to create attachments as well, instead of breaking the view. This is more in line with the attachment model in the backen as well (URL's are also attachments). 

And lo-and-behold, draftjs already thought about it!
https://www.draft-js-plugins.com/plugin/linkify

Thoughts?

assembl/static2/js/app/utils/linkify.js
```diff
@@ -0,0 +1,39 @@
+// @flow
+import linkifyHtml from 'linkifyjs/html';
+import * as linkify from 'linkifyjs';
+
+export function transformLinksInHtml(html: string): string {
+  const linksToReplace = linkify
+    .find(html)
+    .map((link) => {
+      const url = link.href;
+      if (url.startsWith('https://www.youtube.com') || url.startsWith('https://youtube.com')) {
```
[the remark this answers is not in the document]
cedricmessiant: yes, I will enhance the condition

assembl/static2/js/app/utils/linkify.js
cedricmessiant: `html` is a string so it's a "real" copy and since I use `.replace`, it does not modify `transformedHtml` in place

assembl/static2/package.json
cedricmessiant: I need `linkify.find` so I suppose not

cedricmessiant:
I know it is not the best solution but from what I understand, we need to fix this quickly. And I don't think that links and youtube videos will be attachments in v2. With draft-js, links will be draft-js entities and we will manage the html generation in `attachmentsPlugin.js`.

assembl/static2/js/app/utils/linkify.js
cedricmessiant: it ensures that the regexp ends if there is a space, a comma or a `<` (because sometimes, `linkify.find` returns "bad" urls like : `http://www.example.com/</p>`)

cedricmessiant:
I did some enhancements to the condition, I think we can merge

swergas:
Why not do the linkifying the React component way? http://soapbox.github.io/linkifyjs/docs/2.1.html#reactjs-interface

vincentfretin:
Because the content here is dynamic, coming from a variable. @cedricmessiant tested it, it doesn't work.

PR #214, opened 2017-11-03
Now the table of ideas works on all resolutions

paolina92:


assembl/static2/js/app/components/debate/common/ideasLevel.jsx
```diff
@@ -59,6 +53,37 @@ class IdeasLevel extends React.Component {
...
+        });
+      }
+    }
+  }
+  updateDimensions() {
+    this.setState({
+      sliderCount: 0,
+      sliderLeftPosition: 0
+    });
+    this.getDimensions();
```
cedricmessiant: To avoid strange issues, it is preferable to put `this.getDimensions` in the callback param of `setState`:

```this.setState({ ... }, this.getDimensions)
```

assembl/static2/js/app/components/debate/common/ideasLevel.jsx
```diff
@@ -59,6 +53,37 @@ class IdeasLevel extends React.Component {
...
+  }
   getColClassNames(index) {
     const { ideaLevel } = this.props;
     const isFirsStepActif = ideaLevel <= 1;
```
cedricmessiant: maybe fix this typo as you are changing this file : `isFirstStepActive`

assembl/static2/js/app/components/debate/common/ideasLevel.jsx
```diff
@@ -59,6 +53,37 @@ class IdeasLevel extends React.Component {
       }, 500);
     }
   }
+  componentWillUnmount() {
+    window.removeEventListener('resize', this.updateDimensions);
+  }
+  getDimensions() {
```
cedricmessiant: it is more `setDimensions` than `getDimensions` as it changes the state

assembl/static2/css/components/ideas.scss
```diff
@@ -96,3 +96,51 @@
       }
     }
 }
+
+/* iPhone Portrait */
```
ayyazdaniaryan: Well the fix shouldn't just be on iPhone, check on bigger screen (via the chrome simulator) in landscape vs portrait. The changes can still be applied, however.

assembl/static2/css/components/ideas.scss
```diff
@@ -96,3 +96,51 @@
...
+@media screen and (max-device-width: 480px) and (orientation:portrait) {
+  .themes-section {
```
ayyazdaniaryan: This change is repeating. You can make it into a mixin (not function)

http://thesassway.com/advanced/pure-sass-functions

assembl/static2/js/app/components/debate/common/ideasLevel.jsx
```diff
@@ -59,6 +53,39 @@ class IdeasLevel extends React.Component {
...
+      }
+    }
+  }
+  updateDimensions() {
+    this.setState(
+      {
+        sliderCount: 0,
+        sliderLeftPosition: 0
+      },
+      this.setDimensions()
```
ayyazdaniaryan: Why you callback in `setState`?

paolina92: Because `setState` is asynchronous. I want to call `setDimensions()` immediately after setting state.

vincentfretin: Euh @paolina92 you shouldn't call it here, it should be just `this.setDimensions` here.

assembl/static2/css/components/ideas.scss
```diff
@@ -87,6 +87,12 @@
       display: flex;
       float: left;
   }
+  .mobile-tdi {
```
swergas: avoid using "tdi" (if it means "Table Des Idées")

PR #215, opened 2017-11-06
fix Idea.get_ancestors_query to check for long, not only int

vincentfretin:


vincentfretin:
merged in develop and master

PR #223, opened 2017-11-08
Graphql types in flow

vincentfretin:


assembl/static2/flow/README
```diff
@@ -0,0 +1,11 @@
...
+
+    assembl-pshell local.ini
+    from assembl.graphql.schema import generate_schema_json; generate_schema_json()
+    ctr+d
+    cd assembl/static2
+    yarn add apollo-codegen
+    ./node_modules/.bin/apollo-codegen generate js/app/graphql/*.graphql js/app/graphql/mutations/*.graphql --schema /tmp/schema.json --target flow --output flow/graphql_types.flow.js
+    git checkout package.json yarn.lock
+    rm -rf node_modules/
+    yarn
```
vincentfretin: I install apollo-codegen only to generate the file, then I restart fresh without it. I have lots of flow errors if I keep it around, probably a flow version mismatch...

assembl/graphql/schema.py
```diff
@@ -2003,11 +2003,11 @@ class Mutations(graphene.ObjectType):
 Schema = graphene.Schema(query=Query, mutation=Mutations)
 
 
-def print_schema_json(schema):
+def generate_schema_json():
```
cedricmessiant: I imagine it was necessary for apollo-codegen? I don't really like the fact that it takes `Schema` from outside ;-)

cedricmessiant:
this is awesome! :1st_place_medal: 
So we have now to replace types we have written by those in `graphql_types.flow.js`, is that it ?

vincentfretin:
Yes

assembl/graphql/schema.py
vincentfretin: Not necessary. It just simplify the API. You can see the schéma.py module as a facade, and Schema as internal too.

vincentfretin: I added a generate_schema_json_from_schema(schema, output) api

PR #224, opened 2017-11-09
fix global votes and contributions counters

vincentfretin:
Currently global votes counter is counting sentiments on delete posts. It shoudn't.
The contributions counter is wrong too if you do the sum of posts on each idea in phase 1 and 2.

### DIFF
--- a/assembl/graphql/schema.py
+++ b/assembl/graphql/schema.py
@@ -971,7 +971,9 @@ class Query(graphene.ObjectType):
         query = discussion.db.query(models.SentimentOfPost
             ).filter(
                 models.SentimentOfPost.tombstone_condition(),
-                models.Content.tombstone_date == None,
+                models.Content.tombstone_condition(),
+                models.Post.id == models.Content.id,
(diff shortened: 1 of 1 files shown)


assembl/graphql/schema.py
```diff
@@ -961,11 +968,13 @@ class Query(graphene.ObjectType):
...
         discussion_id = context.matchdict['discussion_id']
         discussion = models.Discussion.get(discussion_id)
-        return discussion.db.query(models.SentimentOfPost
+        query = discussion.db.query(models.SentimentOfPost
             ).filter(
-                models.SentimentOfPost.discussion.has(id=discussion_id),
-                models.SentimentOfPost.tombstone_condition()
-            ).count()
+                models.SentimentOfPost.tombstone_condition(),
+                models.Content.tombstone_date == None,
```
ayyazdaniaryan: Perhaps use the `models.Content.tombstone_condition()`, which is basically the same thing as this line.

PR #225, opened 2017-11-09
IdeaSynthesis flow coverage

n0izn0iz:
I turned getSentimentsCount into functional style code so I do a PR instead of direct push

PR #226, opened 2017-11-09
use countable states to count sentiments

vincentfretin:


vincentfretin:
Careful if we merge #225 before that one, it will need to be rebased.

assembl/static2/js/app/constants.js
```diff
@@ -31,4 +31,10 @@ ModeratedPublicationStates[PublicationStates.MODERATED_TEXT_ON_DEMAND] = Publica
 
 export const DeletedPublicationStates = {};
 DeletedPublicationStates[PublicationStates.DELETED_BY_USER] = PublicationStates.DELETED_BY_USER;
-DeletedPublicationStates[PublicationStates.DELETED_BY_ADMIN] = PublicationStates.DELETED_BY_ADMIN;
\ No newline at end of file
+DeletedPublicationStates[PublicationStates.DELETED_BY_ADMIN] = PublicationStates.DELETED_BY_ADMIN;
+
+export const CountablePublicationStates = {};
```
ayyazdaniaryan: Great! All good. One last thing I would add, since this code is duplicated across the front/back boundry, a comment to direct them to which file this is taken from, and a similar comment in Python, so any future changers would be clear that the refactoring has to be propagated across both files/languages. Otherwise, all good.

PR #227, opened 2017-11-09
Better main.jsx flow coverage

n0izn0iz:


assembl/static2/js/app/utils/globalFunctions.js
```diff
@@ -59,54 +72,66 @@ export const getSortedArrayByKey = (arr, key) => {
...
-export const capitalize = (s) => {
+export const capitalize = (s: string) => {
   return s.charAt(0).toUpperCase() + s.slice(1);
 };
 
 export const getDocumentScrollTop = () => {
-  return window.pageYOffset || document.documentElement.scrollTop || document.body.scrollTop || 0;
+  return (
+    window.pageYOffset ||
+    (document.documentElement && document.documentElement && document.documentElement.scrollTop) ||
```
vincentfretin: you have document.documentElement twice here

assembl/static2/js/app/utils/globalFunctions.js
```diff
@@ -59,54 +72,66 @@ export const getSortedArrayByKey = (arr, key) => {
...
 };
 
-export const getDomElementOffset = (el) => {
+export const getDomElementOffset = (el: HTMLElement) => {
   const rect = el.getBoundingClientRect();
   const scrollTop = getDocumentScrollTop();
-  const scrollLeft = window.pageXOffset || document.documentElement.scrollLeft || document.body.scrollLeft || 0;
+  const scrollLeft =
+    window.pageXOffset ||
+    (document.documentElement && document.documentElement && document.documentElement.scrollLeft) ||
```
vincentfretin: here too

assembl/static2/js/app/actions/phaseActions.js
```diff
@@ -1,18 +1,20 @@
-export const updateCurrentPhaseIdentifier = (currentIdentifier) => {
+// @flow
+
+export const updateCurrentPhaseIdentifier = (currentIdentifier: string) => {
```
cedricmessiant: it would be nice to also have typing for return values here. Look at this commit for example 609e7c for an example. You can then use these in reducers (lots of examples in `resourcesCenter` actions and reducers.

n0izn0iz: woops :(

cedricmessiant: I chose to put action types in a separate file but I think it should be a folder if it grows (but it is not a good time to do this)

assembl/static2/js/app/utils/globalFunctions.js
```diff
@@ -1,55 +1,68 @@
...
   return 'token' in location.query ? location.query.token : null;
-};
+}
 
 export const getProvidersData = () => {
-  const data = document.getElementById('login-providers') ? document.getElementById('login-providers').text : null;
+  const data = getScriptText('login-providers');
   try {
-    return JSON.parse(data);
+    return data && JSON.parse(data); // $flowfixme
```
cedricmessiant: what is the problem here ?

assembl/static2/js/app/utils/globalFunctions.js
```diff
@@ -59,54 +72,66 @@ export const getSortedArrayByKey = (arr, key) => {
     return 0;
   });
   return arr;
-};
+}
 
-export const isDateExpired = (date1, date2) => {
+export const isDateExpired = (date1: number, date2: number) => {
```
cedricmessiant: `export const isDateExpired = (date1: number, date2: number): boolean => {` ?
and same for others below (and above)

PR #232, opened 2017-11-13
fix synthesis display bugs in v1 due to bad handling of langstrings

swergas:


assembl/static/js/app/models/langstring.js
```diff
@@ -400,7 +399,8 @@ var LangString = Base.Model.extend({
    * @function app.models.langstrings.LangString.bestValue
    */
   bestValue: function(langPrefs) {
-    return this.best(langPrefs).get("value");
+    var bestLangString = this.best(langPrefs);
+    return bestLangString ? bestLangString.get("value") : null;
```
ayyazdaniaryan: I would pass a `''` instead of null. So it is the last stream of the langstring logic. This is a value after all. I wasn't aware that bestLang would return nothing however.

maparent: LangString returning null should not happen... unless there are no entries (or they all have errors.) If there's no entries, maybe the LangString should not be created? That may be worth checking upstream.

ayyazdaniaryan: That seems to be the real problem here MAP. Either with the ideas or with synthesis, the new CKLSEditor might not be making langstrings appropriately, as there are many issues that are now appearing on newly created syntheses (after langstring changes)

swergas:
This PR is replaced by https://github.com/assembl/assembl/pull/236

PR #233, opened 2017-11-13
fix ideas order in v2

vincentfretin:


assembl/tests/views_tests/test_graphql_ideas.py
```diff
@@ -41,10 +41,10 @@ def test_graphql_get_all_ideas(graphql_request,
...
     assert res.data['rootIdea']['id'] is not None
     assert len(res.data['ideas']) == 4
-    root_idea = res.data['ideas'][0]
-    first_idea = res.data['ideas'][1]
-    second_idea = res.data['ideas'][2]
-    third_idea = res.data['ideas'][3]
+    first_idea = res.data['ideas'][0]
+    second_idea = res.data['ideas'][1]
+    third_idea = res.data['ideas'][2]
+    root_idea = res.data['ideas'][3]
```
cedricmessiant: so root idea is always the last in order ?

vincentfretin: yes, because order is None

assembl/graphql/schema.py
```diff
@@ -997,7 +997,9 @@ def resolve_ideas(self, args, context, info):
         root_idea_id = discussion.root_idea.id
         descendants_query = model.get_descendants_query(
             root_idea_id, inclusive=True)
-        query = query.filter(model.id.in_(descendants_query)
+        query = query.outerjoin(
```
ayyazdaniaryan: I suspect that this is not fully right. There are many idea links that have both a null source and a null target, or a source and null target. But I think it's correct that the order is an ascending number. @maparent Maybe could chime in?

vincentfretin: @ayyazdaniaryan I do exactly the same thing that in the rest api. It's a left outer join, and the order by is on both IdeaLink.order and Idea.creation_date

vincentfretin:
Can I merge @ayyazdaniaryan ?

assembl/graphql/schema.py
ayyazdaniaryan: Alright. All good with me then.

vincentfretin: In the rest api : https://github.com/assembl/assembl/blob/68f3348cfe35de7b962352303b61e893b26584a6/assembl/views/api/idea.py#L125-L130

ayyazdaniaryan:
Si all good with me.

PR #234, opened 2017-11-13
ensure entries are added upon cloning

maparent:
This could lead to an entry-less LangString, which is always a symptom of a bug. (Do not fix downstream without propagating to sentry.)


assembl/tests/model_tests/test_langstring.py
```diff
@@ -643,3 +643,28 @@ def test_identify_ambiguous_body_in_post_by_import(
         first_original().locale_code == 'fr'
     assert post_body_locale_determined_by_import.subject.\
         first_original().locale_code == 'fr'
+
+def test_clone_langstring_has_same_content_as_original(
+        en_langstring, test_session):
+    from datetime import datetime
+
+    tombstone = datetime.utcnow()
+    boba_fett = en_langstring.clone(tombstone=tombstone)
```
ayyazdaniaryan: You should delete `boba_fett` at the end of the test, as this langstring stays in the test db after test execution is completed. Not a problem (that I can think of), but breaks test isolation principle.

swergas: good catch thx!

swergas:
then maybe the cloned langstring should be a fixture, no?

vincentfretin:
This was merged in master develop syntheses

PR #235, opened 2017-11-13
Allow to create CSR (vs self-signed)

maparent:


assembl/auth/make_saml.py
```diff
@@ -45,64 +45,109 @@ def private_key_from_cleaned_text(key_text):
...
-        # Our certificate will be valid for 10 days
-        datetime.datetime.utcnow() + datetime.timedelta(days=days)
-    ).add_extension(
-        x509.BasicConstraints(True, 0), False
     ).add_extension(
-        skid, False
-    ).add_extension(
-        x509.AuthorityKeyIdentifier(
-            skid.digest, [x509.DirectoryName(subject)], serial_number), False
+        x509.KeyUsage(True, False, False, False, True, False, False, False, False), False
```
ayyazdaniaryan: Perhaps add the keynames here for readability? This is daunting to look at if you're not the writer.

configs/docker.rc
```diff
@@ -22,6 +22,9 @@ supervisor__autostart_uwsgi = true
 supervisor__autostart_elasticsearch_server = false
 uwsgi__socket = 0.0.0.0:9000
 uwsgi__processes = 3
+# this means that the exposed nginx is secure (https) even
```
ayyazdaniaryan: Or we could migrate away from having a global nginx and an nginx container. But, not the scope of this PR.

assembl/auth/social_auth.py
```diff
@@ -202,6 +202,12 @@ def get_setting(self, name):
     #     from social.pipeline.utils import partial_from_session
     #     return partial_from_session(self, session)
 
+    def build_absolute_uri(self, path=None):
+        path = super(AssemblStrategy, self).build_absolute_uri(path)
```
ayyazdaniaryan: I see that social_auth has this method, we have our own `get_route` and `frontendUrls` as well that tries to accomplish the same thing. Seems like a lot of ways to do the same thing. Understand that this method is bound to the PyramidStrategy, hence it's usage, but does the result differ from use of frontendUrls drastically?

PR #236, opened 2017-11-14
Fix v1 synthesis second try

swergas:
This fixes some v1 frontend errors when trying to read a synthesis. When trying to read a badly formed synthesis which has empty langstrings, there are still other errors (which is "normal" up to a certain point).


ayyazdaniaryan:
Ah okay, this is on the `messageSend` view. There has been a bug on this for all of V1 for a while, quick fix and a quick win. Other fixes of synthesis will be on it's branch.

PR #239, opened 2017-11-15
Fix displayModal after call to displayCustomModal

n0izn0iz:
https://bluenove.atlassian.net/browse/DEVAS-998


vincentfretin:
LGTM

PR #240, opened 2017-11-16
redirect to the current phase from app.jsx

paolina92:
does anyone have a better idea?


vincentfretin:
I tried a version with Redirect component without success. So I don't have a better solution. I'm merging it.

PR #242, opened 2017-11-21
Update eslint and change arrow-body-style rule

vincentfretin:
To discuss. Changed arrow-body-style rule to `["error", "as-needed", { "requireReturnForObjectLiteral": true }]`, see https://eslint.org/docs/rules/arrow-body-style

Latest commit was done with:
```
npm run eslint:fix
npm run format
edit static2/js/app/components/debate/multiColumns/utils.js with atom to add a line break after the first `=>`
```

This PR probably needs to be rebased once all other branches are merged.
```
git reset --hard HEAD^
git rebase develop
# reformat all files, see above
git add -A
git commit -m"reformat all files with new arrow-body-style rule config"
```

### DIFF
--- a/assembl/static2/.eslintrc.json
+++ b/assembl/static2/.eslintrc.json
@@ -31,7 +31,7 @@
     "no-mixed-operators": 0,
     "no-console": 2,
     "func-names": ["warn", "never"],
-    "arrow-body-style": ["error", "always"],
+    "arrow-body-style": ["error", "as-needed"],
     "comma-dangle": ["warn", "never"],
     "indent": ["warn", 2],
     "eol-last": ["error", "never"],
--- a/assembl/static2/.eslintrc.json
+++ b/assembl/static2/.eslintrc.json
@@ -36,6 +36,7 @@
     "indent": ["warn", 2],
     "eol-last": ["error", "never"],
     "linebreak-style": ["warn", "unix"],
+    "lines-between-class-members": 2,
     "quotes": ["error", "single"],
     "semi": ["error", "always"],
     "object-shorthand": ["error", "never"]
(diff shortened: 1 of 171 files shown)


assembl/static2/.eslintrc.json
```diff
@@ -31,7 +31,7 @@
     "no-mixed-operators": 0,
     "no-console": 2,
     "func-names": ["warn", "never"],
-    "arrow-body-style": ["error", "always"],
+    "arrow-body-style": ["error", "as-needed", { "requireReturnForObjectLiteral": true }],
```
cedricmessiant: so, we don't want this form : 

```() => ({ foo: 'foo' })``` ?

(I understand it can be "strange" and maybe difficult to read but I got used to it (but I'm ok with this rule since prettier allow me to write it like I want ;-) ))

vincentfretin: Yes. We can discuss it of course. I don't really have a preference. There may be less code changes though, that's it.

cedricmessiant:
So, do we plan a meeting about this ?

vincentfretin:
We can discuss it in sprint review.

vincentfretin:
please vote by putting a thumbsup to the following comments

vincentfretin:
`"arrow-body-style": ["error", "as-needed", { "requireReturnForObjectLiteral": true }]`

vincentfretin:
`"arrow-body-style": ["error", "as-needed"]`

vincentfretin:
no change `"arrow-body-style": ["error", "always"]`

ayyazdaniaryan:
I'm not religious about this. Very comfortable with no braces in simple func definitions, so `as-needed` works for me. However, unlike @cedricmessiant, I'm still not used to not seeing a `return` for object literals. The way it's written, it makes me think of Scala. Just my 2 cents.

vincentfretin:
From https://github.com/assembl/assembl/pull/254/files/1c5e786ca9b089725f6cfaa2271f2a2ca930538e#r154110182
we can update to eslint 4.12.1 and add `"lines-between-class-members": 2` and fix the 150 errors.
https://eslint.org/docs/rules/lines-between-class-members

PR #243, opened 2017-11-21
Add stylelint

vincentfretin:
Similar to eslint, add `npm run stylelint`, `npm run stylelint:fix` commands, and a pre-commit hook.
.stylelintrc uses the stylelint-config-standard config.
You can install linter-stylelint plugin in atom.


assembl/static2/css/variables.scss
```diff
@@ -162,26 +162,26 @@ $screen-lg-min: 1200px;
 // Proper media-queries examples https://stackoverflow.com/questions/18424798/twitter-bootstrap-3-how-to-use-media-queries
 
 /* Large desktops and laptops */
-@media (min-width: $screen-lg-min) {
-
-}
+// @media (min-width: $screen-lg-min) {
```
cedricmessiant: uhuh

vincentfretin:
see https://github.com/stylelint/stylelint-config-standard for additional rules we can add.
As you can see, there is not much changes that needed to be done to respect stylelint-config-standard, so we already respect the google & airbnb styleguide actually.

vincentfretin:
I merged and reverted the merge. The glob is wrong, should be `stylelint '**/*.scss''` instead of `stylelint '*/*.scss'` to match all files in subdirectories.

vincentfretin:
merged into develop with https://github.com/assembl/assembl/commit/96b25925567f32562d0fe7b1f647986a9ce58ed3 and https://github.com/assembl/assembl/commit/338f1cca41bfdfc917b71c6920b1bcdde62f1e22

assembl/static2/css/variables.scss
swergas: Why is this commented?

vincentfretin: stylelint doesn't like empty rules. We can remove it completly too.

PR #244, opened 2017-11-22
Do not send personal info to raven

maparent:


maparent:
piwik script only sends user-id in v1; see context.js:273
no clue if it's more verbose in v2.

PR #245, opened 2017-11-23
Fix vote widget langstrings

swergas:
Fixes DEVAS-850 and other similar UX issues (related to langstrings and showing content in different languages)

### DIFF
--- a/assembl/static/js/app/views/tokenVoteSession.js
+++ b/assembl/static/js/app/views/tokenVoteSession.js
@@ -818,6 +818,7 @@ var TokenVoteItemView = Marionette.LayoutView.extend({
   initialize: function(options){
     this.childIndex = options.childIndex;
     this.parent = options.parent;
+    this.userLanguagePreferences = options.userLanguagePreferences;
   },
 
   regions: {
@@ -891,11 +892,15 @@ var TokenVoteCollectionView = Marionette.CompositeView.extend({
   childView: TokenVoteItemView,
   template: '#tmpl-tokenVoteCollection',
   childViewContainer: "tbody",
+  initialize: function(options) {
+    this.options = options;
+  },
   childViewOptions: function(model, index){
     var that = this;
     return {
       childIndex: index,
-      parent: that
+      parent: that,
+      userLanguagePreferences: that.getOption('userLanguagePreferences')
     };
   },
   templateHelpers: function(){
@@ -1523,6 +1528,7 @@ var TokenVoteSessionModal = Backbone.Modal.extend({
         tokenCategories: that.tokenCategories,
         myVotesCollection: that.myVotesCollection,
         tokenSize: tokenSize,
+        userLanguagePreferences: userLanguagePreferences,
         viewComparator: function(idea){
           return _.findIndex(permutation, function(idea2){return idea2.id == idea.id;});
         }
--- a/assembl/static/js/app/views/tokenVoteSession.js
+++ b/assembl/static/js/app/views/tokenVoteSession.js
@@ -16,6 +16,7 @@ var Marionette = require('../shims/marionette.js'),
   BreadCrumbView = require('./breadcrumb.js'),
   CKEditorLSField = require('./reusableDataFields/ckeditorLSField.js'),
   IdeaModel = require('../models/idea.js'),
+  LangString = require('../models/langstring.js'),
   i18n = require('../utils/i18n.js'),
   Ctx = require('../common/context.js'),
   Moment = require('moment'),
@@ -251,7 +252,7 @@ var RemainingCategoryTokensView = Marionette.ItemView.extend({
     el.addClass("description");
 
     var categoryNameLangString = category.get("name");
-    var categoryNameBestTranslation = categoryNameLangString.bestWithErrors(that.languagePreferences, false).entry.value();
+    var categoryNameBestTranslation = categoryNameLangString.bestWithErrors(that.userLanguagePreferences, false).entry.value();
     // var s = i18n.sprintf(i18n.ngettext("<span class='available-tokens-number'>%d</span> remaining %s token", "<span class='available-tokens-number'>%d</span> remaining %s tokens", data["remaining_tokens"]), data["remaining_tokens"], category.get("typename")); // TODO: use "name" field instead (LangString)
     var s = i18n.sprintf(i18n.ngettext("<span class='available-tokens-number'>%d</span> remaining %s token", "<span class='available-tokens-number'>%d</span> remaining %s tokens", data["remaining_tokens"]), data["remaining_tokens"], categoryNameBestTranslation);
     el.html(s);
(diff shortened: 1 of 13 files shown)


assembl/static/js/app/views/tokenVoteSession.js
```diff
@@ -1303,8 +1321,8 @@ var TokenResultView = Marionette.LayoutView.extend({
     var settings = this.model.get("settings") || {},
         items = "items" in settings ? settings.items : [],
         questionItem = items.length ? items[0] : null,
-        questionTitle = "question_title" in questionItem ? questionItem.question_title : "",
-        questionDescription = "question_description" in questionItem ? questionItem.question_description : "",
+        questionTitle = this.questionTitle ? this.questionTitle : "question_title" in questionItem ? questionItem.question_title : "",
```
ayyazdaniaryan: This will show the original langstring value if the user has no langprefs?

assembl/static/js/app/views/tokenVoteSession.js
```diff
@@ -1464,6 +1482,20 @@ var TokenVoteSessionModal = Backbone.Modal.extend({
...
     
     Promise.join(collectionManager.getUserLanguagePreferencesPromise(Ctx), collectionManager.getAllIdeasCollectionPromise(), function(userLanguagePreferences, allIdeasCollection) {
+      that.userLanguagePreferences = userLanguagePreferences;
+      var settings = that.widgetModel.get("settings") || {};
+      var items = "items" in settings ? settings.items : [];
+      var question_item = items.length ? items[0] : null;
+
+      var questionTitleLangString = new LangString.Model();
+      questionTitleLangString.initFromObjectProperty(question_item, 'question_title');
+      that.$('.question-title').text(questionTitleLangString.bestValue(that.userLanguagePreferences));
```
ayyazdaniaryan: I would add more conditions around this (or move this logic into `bestValue`, if there is no `userLanguagePreference`, then `firstOriginal` is returned. Because, currently, if there is no langPref, the method throws an error, so this crash would become very visible.

Perhaps you could get the `bestValue`'ed lang as a var, and if it doesn't exist, return the original in the `$(..).text()`.

assembl/static/js/app/views/tokenVoteSession.js
```diff
@@ -1464,6 +1482,20 @@ var TokenVoteSessionModal = Backbone.Modal.extend({
...
+
+      var questionDescriptionLangString = new LangString.Model();
+      questionDescriptionLangString.initFromObjectProperty(question_item, 'question_description');
+      that.$('.question-description').text(questionDescriptionLangString.bestValue(that.userLanguagePreferences));
```
ayyazdaniaryan: Same comment as above.

assembl/static/widget/creativity/app/services.js
```diff
@@ -13,6 +13,55 @@ creativityServices.service('AssemblToolsService', ['$window', '$rootScope', '$lo
   };
 }]);
 
+creativityServices.service('LangStringService', ['$window', '$rootScope', '$log', function($window, $rootScope, $log) {
+  this.isMachineTranslation = function(langString) {
+    return langString["@language"].indexOf("-x-mtfrom-") > 0;
+  };
+
+  // returns the LangStringEntry (of this `langString` LangString) that is the most adapted to language `lang`
+  this.bestForLang = function(langString, lang){
```
ayyazdaniaryan: Is this logic duplicate of `best` in `models/langstring`?

swergas: yes except that lang is only one language, not a whole ordered set of languages

assembl/static/js/app/views/tokenVoteSession.js
```diff
@@ -1016,12 +1020,13 @@ var TokenVoteResultView = Marionette.LayoutView.extend({
...
   serializeData: function() {
-    var defn = this.model.get('objectConnectedTo').get('definition');
+    var that = this;
+    var defn = that.model.get('objectConnectedTo').get('definition');
     return {
-      ideaTitle: this.model.get('objectConnectedTo').getShortTitleDisplayText(this.userLanguagePreferences),
-      categoryResult: this.results,
-      showDescriptionButton: defn && !defn.isEmptyStripped(this.userLanguagePreferences),
-      descriptionButton: this.descriptionButton
+      ideaTitle: that.model.get('objectConnectedTo').getShortTitleDisplayText(that.userLanguagePreferences),
```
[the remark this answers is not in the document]
swergas: I did not change anything there except the use of `that`, so maybe this improvement remark is out of scope of the bug fix. I would have to dig more on how `objectConnectedTo` is meant to be used

assembl/static/js/app/views/tokenVoteSession.js
```diff
@@ -1177,12 +1182,13 @@ var TokenVoteResultCollectionView = Marionette.CompositeView.extend({
   },
 
   childViewOptions: function(){
+    var that = this;
```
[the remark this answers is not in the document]
swergas: No real reason now (probably a backtrack of a solution I investigated), I can remove this part of the change if you want.

assembl/static/js/app/views/tokenVoteSession.js
swergas: Yes and also until the langprefs have been received, and also if there is any issue with the received langprefs

assembl/static/js/app/views/tokenVoteSession.js
swergas: If a user did not set any langpref a default one is received I think. Ensuring that getUserLanguagePreferencesPromise contains a valid langpref is probably the job of this getUserLanguagePreferencesPromise function?

assembl/static/js/app/views/tokenVoteSession.js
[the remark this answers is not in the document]
ayyazdaniaryan: Yeah, it's one less variable declaration.

assembl/static/js/app/views/tokenVoteSession.js
ayyazdaniaryan: No, the `getUserLanguagePreferencesPromise` doesn't return a default. I don't think the API is returning a default lang_pref from configs, btw. So in rare case (now) that user doesn't have lang_pref, returns empty collection.

assembl/static/js/app/views/tokenVoteSession.js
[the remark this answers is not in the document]
swergas: ok fixed in 2ea2189b330b09a0db809c559919b63a3b4ebad1

assembl/static/js/app/views/tokenVoteSession.js
swergas: ok I think I solved this potential issue in 85dad6b06f0f81bf41dea9d28b36715b47ae7cf0 and 19a7201594254bf3905c993ab7cffe2e4f1e44ef. But for a better solution, I think a full reevaluation/clarification on how langprefs should be used in combination with langstrings is necessary

maparent:
Ah. Just seeing this. Did not look in depth, but this looks again like symptom masking. Do we know why the entry list is empty?

swergas:
no, the entry list is not empty

maparent:
apologies, I did look too fast, then.

PR #246, opened 2017-11-23
Show piwik metrics on landing page (DEVAS-974)

swergas:
Implements DEVAS-974


swergas:
Implementation is now done, the PR is ready for code review. Things that can be improved but I don't really know how:
* Add more tests, but which ones? The only ones I think about probably require to mock a piwik server response, but I don't really know how
* Find a way to not do several piwik API calls for one graphql call (detect that a graphql requests asks for all 3 fields sumVisitsLength, nbPageviews, nbUniqPageviews), and so call only one resolver instead of 3
* Cache response from the piwik server, for example into redis (we have redis as a dependency because it is required by other dependencies but we don't use it ourselves yet)

assembl/static2/js/app/components/home/header/statistic.jsx
```diff
@@ -39,49 +41,35 @@ class Statistic extends React.Component {
...
-          if (minutes < 10) {
-            minutes = `0${minutes}`;
-          }
-          if (seconds < 10) {
-            secondsRemaining = `0${secondsRemaining}`;
-          }
-          return `${hours}h ${minutes}m ${secondsRemaining}s`;
-        };
-        const readableDuration = totalSecondsToReadableDuration(totalSeconds);
+        const secondsAsNumber = parseInt(totalSeconds, 10); // don't forget the second param
```
cedricmessiant: why do we need the second param here? According to https://www.w3schools.com/jsref/jsref_parseInt.asp if the string does not starts with `0x` or `0` we don't need it

assembl/lib/piwik.py
```diff
@@ -0,0 +1,197 @@
...
+    params = {
+        "module": "API",
+        "format": "JSON",
+        "token_auth": piwik_api_token
+    }
+    params["method"] = "UsersManager.addUser"
+    params["userLogin"] = userLogin # Piwik has two different fields for login and email, but a user can have the same value as login and email
+    params["password"] = password
+    params["email"] = email
+    if ( alias ):
```
cedricmessiant: aren't parenthesis useless here?

assembl/lib/piwik.py
```diff
@@ -0,0 +1,197 @@
...
+    Warning: A new Piwik site is created everytime this method is called, even if another Piwik website already exists with the same name and URLs"""
+    params = {
+        "module": "API",
+        "format": "JSON",
+        "token_auth": piwik_api_token
+    }
+    params["method"] = "SitesManager.addSite"
+    params["siteName"] = siteName
+    params["urls"] = urls
+    if ( ecommerce ):
```
cedricmessiant: same as above for all parenthesis

assembl/lib/piwik.py
```diff
@@ -0,0 +1,197 @@
...
+    if result.status_code != 200:
+        raise requests.ConnectionError()
+
+    content = result.json() # Content should be something like {"value": 47}
+    # print "piwik_SitesManager_addSite", content
+
+    if not content:
+        raise requests.ConnectionError()
+
+    if "value" in content:
```
cedricmessiant: it can be replaced by: `return content.get("value", False)`

assembl/lib/piwik.py
```diff
@@ -0,0 +1,197 @@
...
+
+    if result.status_code != 200:
+        raise requests.ConnectionError()
+    content = result.json() # Content should be either an empty array, or an array like [{"idsite":"44"}]
+    # print "piwik_UsersManager_setUserAccess", content
+
+    if not content:
+        raise requests.ConnectionError()
+
+    user_access_is_set = ("result" in content and content["result"] == "success")
```
cedricmessiant: `return content.get("result", "error") == "success")`

assembl/static2/js/app/components/home/header/statistic.jsx
```diff
@@ -37,7 +37,7 @@ class Statistic extends React.Component {
     });
   };
   render() {
-    const { rootIdea, numParticipants, totalSentiments } = this.props.data;
+    const { rootIdea, numParticipants, totalSentiments, visitsAnalytics } = this.props.data;
```
cedricmessiant: it would have been nice to add flow typing and a test for this component

assembl/static2/css/components/assembl-glyph.scss
```diff
@@ -116,3 +116,5 @@
 .assembl-icon-ellipsis:before { content: '\f141'; } /* '' */
 .assembl-icon-ellipsis-vert:before { content: '\f142'; } /* '' */
 .assembl-icon-sentiment-neutral:before { content: '\e86f'; } /* '' */
+.assembl-icon-double-page:before { content: '\e869'; } /* '' */
+.assembl-icon-timer:before { content: '\e86A'; } /* '' */
```
vincentfretin: Be careful, the font has been modified in #250 too

assembl/graphql/discussion.py
```diff
@@ -183,3 +183,30 @@ def mutate(root, args, context, info):
...
+
+    sum_visits_length = graphene.Int()
+    nb_pageviews = graphene.Int()
+    nb_uniq_pageviews = graphene.Int()
+
+    def query_analytics_single_metric(self, args, context, info, single_metric):
+        discussion_id = context.matchdict['discussion_id']
+        discussion = models.Discussion.get(discussion_id)
+        start = args.get('start_date') or None
+        end = args.get('end_date') or None
```
vincentfretin: You didn't define `start_date`, neither `end_date` as args for `sum_visits_length`, `nb_pageviews`, `nb_uniq_pageviews`.

assembl/tests/views_tests/test_graphql.py
```diff
@@ -1938,3 +1938,19 @@ def __init__(self, filename, type):
...
+    nbPageviews
+    nbUniqPageviews
+  }
+}
+"""
+    res = schema.execute(query, context_value=graphql_request)
+    assert res.data['visitsAnalytics'] is not None
+    assert res.data['visitsAnalytics']['sumVisitsLength'] == None
+    assert res.data['visitsAnalytics']['nbPageviews'] == None
+    assert res.data['visitsAnalytics']['nbUniqPageviews'] == None
```
vincentfretin: You should really add a test with non None values by mocking the result of `discussion.get_visits_time_series_analytic` with the `mock` library. There are some examples already in `test_api.py` and `test_view_auth.py`. We don't want to break things when we will upgrade to graphene 2 that changes the api of resolvers.

vincentfretin: And a test with a range to test start_date and end_date args.

assembl/static2/js/app/components/home/header/statistic.jsx
vincentfretin: eslint gives you an error if you don't specify the radix

assembl/static2/css/components/assembl-glyph.scss
swergas: Yes, I noticed it also while reviewing ^^ https://github.com/assembl/assembl/pull/250#discussion_r154043894
So, using Fontello instead of regular SVG files is not that good when developing several branches at the same time that make changes to icons :(

assembl/graphql/discussion.py
swergas: yes these are optional

assembl/tests/views_tests/test_graphql.py
swergas: Yes but I don't know how to do it, as I said in https://github.com/assembl/assembl/pull/246#issuecomment-347901370

assembl/static2/js/app/components/home/header/statistic.jsx
swergas: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt clearly says:
> Always specify this parameter to eliminate reader confusion and to guarantee predictable behavior. Different implementations produce different results when a radix is not specified, usually defaulting the value to 10.

assembl/lib/piwik.py
swergas: yes that is just a taste of aesthetic/readability. Do we use a rule that forces us to use the minimum amount of parentheses?

assembl/graphql/discussion.py
vincentfretin: Currently you can't do a query 
```
query GetVisitsAnalytics {
￼  visitsAnalytics {
￼    sumVisitsLength(start_date: "2017-11-01T00:00:00Z", end_date: ""2017-12-31T29:59:59Z"")
￼    nbPageviews(start_date: "2017-11-01T00:00:00Z", end_date: ""2017-12-31T29:59:59Z"")
￼    nbUniqPageviews(start_date: "2017-11-01T00:00:00Z", end_date: ""2017-12-31T29:59:59Z"")
￼  }
}
```

If you want to do such query, you need:
```
sum_visits_length = graphene.Int(start_date=DateTime(), end_date=DateTime())
nb_pageviews = graphene.Int(start_date=DateTime(), end_date=DateTime())
nb_uniq_pageviews = graphene.Int(start_date=DateTime(), end_date=DateTime())
```

assembl/lib/piwik.py
swergas: oh yes sometimes I forget the existence of this

assembl/static2/js/app/components/home/header/statistic.jsx
cedricmessiant: Ok, good. I thought I missed something when I saw the comment

assembl/tests/views_tests/test_graphql.py
vincentfretin: In this particuliar case, I would do something like this:
```
    def test_query_visits_analytics(discussion, graphql_request):
        def fake_get_visits_time_series_analytics(self, start_date=None, end_date=None, only_fields=None):
             if "sum_visits_length" in only_fields:
                 return 15

        with mock.patch.object(models.Discussion, 'get_visits_time_series_analytics', fake_get_visits_time_series_analytics):
            
            ...do your graphql query...

File: assembl/lib/piwik.py (hunk 2 above)
Comment (cedricmessiant) [reply assembl/lib/piwik.py]: My bad, I thought it was in PEP8 but it seems not. However, most linting tools wont like this (e.g. just tested it with pylint). When there is only one statement, if a comma "appears" inside, it will be seen as a tuple by python and so it will be always true whereas if you don't use parenthesis, your editor (or the interpretor) should complain on `if alias,:`.
Also, we generally don't use spaces inside parenthesis in python code.

File: assembl/graphql/schema.py [hunk 10]
```diff
@@ -189,6 +190,9 @@ def resolve_locales(self, args, context, info):
     def resolve_resources_center(self, args, context, info):
         return ResourcesCenter()
 
+    def resolve_visits_analytics(self, args, context, info):
+        return VisitsAnalytics()
```
vincentfretin: If you want to just do one piwik query instead of three, do the query here and create `VisitsAnalytics(sum_visits_length=sum_visits_length, nb_pageviews=nb_pageviews, nb_uniq_pageviews=nb_uniq_pageviews)`
We do something similar for `Video` in `graphql/idea.py`

vincentfretin:
For cache, you can use dogpile_cache, create a new region, and use `@piwik_cache.cache_on_arguments()` decorator, see example in `views/api2/discussion.py`
Default conf currently use a file for cache, but you can create a new region and use a redis backend, see
http://dogpilecache.readthedocs.io/en/latest/api.html?highlight=dbm#redis-backends
https://pypi.python.org/pypi/pyramid_dogpile_cache

assembl/graphql/schema.py
vincentfretin: If you want to know if the graphql query is asking for the fields, you can use `get_fields`, there is an example in `graphql/idea.py`, probably like this:
```
fields = get_fields(info)
if 'sumVisitsLength' in fields:
    ....
```

assembl/lib/piwik.py
swergas: improvement done

assembl/lib/piwik.py
swergas: improvement done

assembl/lib/piwik.py
swergas: improvement done

assembl/lib/piwik.py
swergas: improvement done

assembl/static2/js/app/components/home/header/statistic.jsx
swergas: flow types have now been added

assembl/graphql/schema.py
swergas: improvement done in https://github.com/assembl/assembl/pull/246/commits/295b190aeffffc3eff225453d6d338c5a28566d8

assembl/tests/views_tests/test_graphql.py
swergas: improvement done in https://github.com/assembl/assembl/pull/246/commits/295b190aeffffc3eff225453d6d338c5a28566d8

assembl/graphql/schema.py
swergas: improvement done in 301fd66e4c948aafcca0efe9427c876ef721e8d6 . I tested it by counting printed statements, which worked, but I could not put it in an automated test because we use mock for python 2.7 which does not have `mock_method.assert_called_once()`

assembl/graphql/discussion.py
vincentfretin: @swergas You didn't address this one.

swergas: Yes this was intentional. Having the right architecture for future extension and having it work from the REST API v2 is enough for now (so we can answer probable queries from clients), because setting time limits feature is out of scope of the story. But yes Maybe this is not clear enough in the code :/

vincentfretin: Ok. In the code, it's misleading, you thing it would work, but it will raise an error. You should put a comment in the code with a link to my previous comment for the future you. :)

PR #248, opened 2017-11-26
install_elasticsearch: do everything remotely

maparent:


fabfile.py
```diff
@@ -1785,51 +1784,44 @@ def install_elasticsearch():
...
+        if ' ' in sha1_effective:
+            sha1_effective = sha1_effective.split(' ')[-1]
+        assert sha1_effective == sha1_expected, "sha1sum of elasticsearch tarball doesn't match, exiting"
+        run('tar zxf ' + tar_filename)
+        run('rm ' + tar_filename)
+        run('mv %s elasticsearch' % base_filename)
 
         # ensure that the folder being scp'ed to belongs to the user/group
-        sudo('chown -R {user}:{group} {path}'.format(
+        run('chown -R {user}:{group} {path}'.format(
```
ayyazdaniaryan: Ah, this may work on our server configurations, but fails on our local setups (especially on mac). Must add a `_group` key to rc files, and default usuage to be `_user`, override as needed.

maparent: good point. Added env.group (still in single-commit form.)

PR #250, opened 2017-11-27
Feature configurable sections

vincentfretin:


assembl/alembic/versions/c98a9b6f6b7f_add_sections.py
```diff
@@ -0,0 +1,117 @@
...
+            sa.Column('type', sa.String(60), nullable=False),
+            sa.Column('discussion_id',
+                      sa.Integer,
+                      sa.ForeignKey(
+                          'discussion.id',
+                          ondelete="CASCADE",
+                          onupdate="CASCADE"), nullable=False, index=False),
+            sa.Column('title_id', sa.Integer, sa.ForeignKey('langstring.id')),
+            sa.Column('url', URLString(1024)),
+            sa.Column('order', sa.Float, default=0.0),
```
vincentfretin: You should set nullable=False for order.

assembl/graphql/section.py
```diff
@@ -0,0 +1,160 @@
...
+                # Better to have this message than
+                # 'NoneType' object has no attribute 'owner_object'
+                # when creating the saobj below if title=None
+
+            title_langstring = langstring_from_input_entries(title_entries)
+
+            kwargs = {}
+            kwargs['section_type'] = args.get('section_type', 'CUSTOM')
+            kwargs['url'] = args.get('url')
+            kwargs['order'] = args.get('order')
```
vincentfretin: You should set a default value for order, like `args.get('order', 10.0)`

assembl/graphql/section.py
```diff
@@ -0,0 +1,160 @@
...
+                # Better to have this message than
+                # 'NoneType' object has no attribute 'owner_object'
+                # when creating the saobj below if title=None
+
+            update_langstring_from_input_entries(
+                section, 'title', title_entries)
+
+            kwargs = {}
+            kwargs['url'] = args.get('url', None)
+            kwargs['order'] = args.get('order', None)
```
vincentfretin: same here, order shouldn't be None.

assembl/models/section.py
```diff
@@ -0,0 +1,94 @@
...
+
+    section_type = Column(
+        Enum(*section_types, name='section_types'),
+        nullable=False,
+        default=SectionTypesEnum.CUSTOM.value,
+        server_default=SectionTypesEnum.CUSTOM.value
+    )
+
+    order = Column(
+        Float, nullable=False, default=0.0)
```
vincentfretin: nullable=False here, good. But this is not the case in alembic step, and the default value you use in UpdateSection.

assembl/models/section.py
```diff
@@ -0,0 +1,94 @@
...
+        cascade="all, delete-orphan")
+
+    url = Column(URLString)
+
+    section_type = Column(
+        Enum(*section_types, name='section_types'),
+        nullable=False,
+        default=SectionTypesEnum.CUSTOM.value,
+        server_default=SectionTypesEnum.CUSTOM.value
+    )
```
vincentfretin: You could have avoided the section_type enum alltogether and just used sqlalchemy polymorphism and the type field, no?

assembl/models/section.py
```diff
@@ -0,0 +1,94 @@
...
+    __mapper_args__ = {
+        'polymorphic_identity': 'section',
+        'polymorphic_on': type,
+        'with_polymorphic': '*'
+    }
+
+    crud_permissions = CrudPermissions(
+        P_ADMIN_DISC, P_READ, P_ADMIN_DISC, P_ADMIN_DISC)
+
+LangString.setup_ownership_load_event(Section, ['title',])
```
vincentfretin: You don't need to have a trailing comma, this is not a tupe.

assembl/static2/css/components/assembl-glyph.scss
```diff
@@ -116,3 +116,5 @@
 .assembl-icon-ellipsis:before { content: '\f141'; } /* '' */
 .assembl-icon-ellipsis-vert:before { content: '\f142'; } /* '' */
 .assembl-icon-sentiment-neutral:before { content: '\e86f'; } /* '' */
+.assembl-icon-up-bold:before { content: '\e80e'; } /* '' */
+.assembl-icon-down-bold:before { content: '\e810'; } /* '' */
```
vincentfretin: please modify the hash in the urls to the font at the top of this file to avoid cache issue

assembl/static2/js/app/components/administration/discussion/editSectionForm.jsx
```diff
@@ -0,0 +1,139 @@
...
+  handleTitleChange,
+  handleCheckboxChange,
+  handleUrlChange,
+  handleDeleteClick,
+  handleDownClick,
+  handleUpClick
+}: EditSectionFormProps) => {
+  const { translations } = i18n;
+  const titlePh = I18n.t('administration.sections.titlePh');
+  const urlPh = I18n.t('administration.sections.urlPh');
```
vincentfretin: What Ph means?

assembl/static2/js/app/components/administration/discussion/manageSectionsForm.jsx
```diff
@@ -0,0 +1,60 @@
...
+import React from 'react';
+import { connect } from 'react-redux';
+import { I18n } from 'react-redux-i18n';
+import { OverlayTrigger } from 'react-bootstrap';
+import SectionTitle from '../sectionTitle';
+import EditSectionForm from './editSectionForm';
+import { addSectionTooltip } from '../../common/tooltips';
+import * as actions from '../../../actions/adminActions/adminSections';
+
+const ManageSectionsForm = ({ sections, selectedLocale, createSection }) => {
```
vincentfretin: flow annotations missing

assembl/static2/js/app/utils/translations.js
```diff
@@ -648,16 +665,33 @@ const Translations = {
...
       edition: "Edit the Discussion",
       landingpage: "Landing page",
+      sectionsTitle: "Set sections",
+      up: "Up",
+      down: "Down",
       menu: {
         phase: "Phase %{count} - %{description}",
-        language: "Discussion Language"
+        language: "Discussion Language",
+        sections: "Edit Debate Sections"
```
vincentfretin: you should add "debate" to new messages

assembl/tests/views_tests/test_graphql.py
```diff
@@ -1938,3 +1940,153 @@ def __init__(self, filename, type):
...
+        {u"value": u"Section personnalisée", u"localeCode": u"fr"},
+        {u"value": u"Custom section", u"localeCode": u"en"}
+    ]
+    variables = {
+        "titleEntries": title_entries,
+        "url": u"http://www.example.com",
+        "order": 5.0
+    }
+    res = schema.execute(u"""
+mutation createSection($titleEntries:[LangStringEntryInput!]!,$url:String,$order:Float) {
```
vincentfretin: Can you please use the same styleguide as prettier to format query in Python tests? space after ',' and ':'. Thanks.

assembl/tests/views_tests/test_graphql.py
```diff
@@ -1938,3 +1940,153 @@ def __init__(self, filename, type):
...
+    id
+} }"""
+    res = schema.execute(query, context_value=graphql_request)
+    result = res.data
+    assert result is not None
+    assert len(result['sections']) == 5
+
+
+def test_mutation_delete_section_fails_for_non_custom_sections(sections, graphql_request):
+    # raise error if we try to delete anything but CUSTOM type
```
vincentfretin: not really and error, just success=False

assembl/graphql/section.py
cedricmessiant: as it is an update, I thought that I should enable None and that it will not modify it

assembl/models/section.py
cedricmessiant: I don't understand how the type field could have helped me

assembl/static2/js/app/components/administration/discussion/editSectionForm.jsx
cedricmessiant: placeholder (took some time for me to figure it out the first time too)

assembl/alembic/versions/c98a9b6f6b7f_add_sections.py
cedricmessiant: done

assembl/graphql/section.py
cedricmessiant: done

assembl/graphql/section.py
cedricmessiant: done

assembl/models/section.py
cedricmessiant: done

assembl/tests/views_tests/test_graphql.py
cedricmessiant: true

assembl/tests/views_tests/test_graphql.py
cedricmessiant: done

assembl/models/section.py
vincentfretin: type field is a string. You could have added types like this:
```
class SectionCustom(Section):
    __mapper_args__ = {
        'polymorphic_identity': 'custom'
    }
                                                                                
    @classproperty
    def name(cls): 
        return 'custom'
```

vincentfretin: But nevermind, keep this enum. We can maybe use type field later to handle footer actions, user actions if there is a need.

assembl/graphql/section.py
vincentfretin: FYI `args.get('order', None)` and `args.get('order')` is actually the same, it's the dict api. If you want to not specify order and keep the current value, you need to add a condition on `args.get('order')` before setting `kwargs['order'] = None`. Now it will raise an error because you added nullable=False.

assembl/static2/js/app/components/administration/discussion/editSectionForm.jsx
vincentfretin: Yeah, for me too the first time, then I forgot! We should really rename this.

assembl/static2/js/app/utils/translations.js
paolina92: For the admin section, we don't care

assembl/models/section.py
swergas: also why a float and not an int?

assembl/alembic/versions/c98a9b6f6b7f_add_sections.py
```diff
@@ -49,9 +49,47 @@ def upgrade(pyramid_env):
...
         )
 
-        # TODO: insert default data
+        # insert default sections
+        with m.Section.default_db.no_autoflush as db:
+            discussions = db.query(m.Discussion.id).all()
+            for discussion_id in discussions:
+                homepage_section = m.Section(
+                    discussion_id=discussion_id,
+                    title=m.LangString.create(u'Home', 'en'),
```
swergas: for each of these, you should probably also add at least their french translations, no?

assembl/models/section.py
vincentfretin: It's like idea order field, a float. So you can possibly change the order of an element without changing order field of all elements. If you want to insert a element between 1.0 and 2.0, you can use 1.5 :)

assembl/static2/css/components/assembl-glyph.scss
cedricmessiant: @paolina92 have you seen this one?

assembl/static2/js/app/components/administration/discussion/editSectionForm.jsx
```diff
@@ -0,0 +1,135 @@
...
+        <FormControlWithLabel label={titlePh} onChange={handleTitleChange} type="text" value={title} />
+      ) : null}
+      {type === 'HOMEPAGE' ? (
+        <FormGroup>
+          <Checkbox checked={hasUrl} onChange={handleCheckboxChange}>
+            <Translate value="administration.sections.externalPage" />
+          </Checkbox>
+        </FormGroup>
+      ) : null}
+      {(type === 'HOMEPAGE' && hasUrl) || type === 'CUSTOM' ? (
```
cedricmessiant: it means that you can't edit the url of the homepage if it is null, no ?

assembl/static2/js/app/components/administration/discussion/editSectionForm.jsx
```diff
@@ -0,0 +1,135 @@
...
+          </OverlayTrigger>
+        ) : null}
+      </div>
+      <div className="clear" />
+      {type !== 'HOMEPAGE' ? (
+        <FormControlWithLabel label={titlePh} onChange={handleTitleChange} type="text" value={title} />
+      ) : null}
+      {type === 'HOMEPAGE' ? (
+        <FormGroup>
+          <Checkbox checked={hasUrl} onChange={handleCheckboxChange}>
```
cedricmessiant: I would change naming of this function, maybe something like : `toggleExternalPageField` ? (not perfectly sure of what it does)

assembl/static2/js/app/components/administration/discussion/editSectionForm.jsx
```diff
@@ -0,0 +1,135 @@
...
+      {(type === 'HOMEPAGE' && hasUrl) || type === 'CUSTOM' ? (
+        <FormControlWithLabel label={urlPh} onChange={handleUrlChange} type="text" value={url} />
+      ) : null}
+      <div className="separator" />
+    </div>
+  );
+};
+
+const mapStateToProps = (state, { id, locale }) => {
+  const sections = state.admin.sections.sectionsById.get(id);
```
cedricmessiant: `sections` -> `section`

assembl/alembic/versions/c98a9b6f6b7f_add_sections.py
cedricmessiant: yes, I did it in another commit

assembl/static2/js/app/components/administration/discussion/editSectionForm.jsx
```diff
@@ -0,0 +1,135 @@
...
+      return dispatch(updateSectionUrl(id, e.target.value));
+    },
+    handleCheckboxChange: () => {
+      return dispatch(toggleExternalPage(id));
+    },
+    handleDeleteClick: () => {
+      return dispatch(deleteSection(id));
+    },
+    handleUpClick: () => {
+      return dispatch(upSection(id));
```
cedricmessiant: `moveSectionUp` / `moveSectionDown` ?

vincentfretin:
I reviewed the commits, I don't have anything more to say.

assembl/static2/js/app/components/administration/discussion/manageSectionsForm.jsx
```diff
@@ -0,0 +1,62 @@
...
+          </OverlayTrigger>
+        </form>
+      </div>
+    </div>
+  );
+};
+
+const mapStateToProps = (state) => {
+  const { sectionsInOrder, sectionsById } = state.admin.sections;
+  const sections = sectionsInOrder.filter((id) => {
```
cedricmessiant: I think that this `.filter` is useless (if you removed the id from `sectionsInOrder` when a section is deleted which you should to be able to move sections properly)

assembl/static2/js/app/components/administration/saveButton.jsx
```diff
@@ -215,12 +237,33 @@ const SaveButton = ({
...
+        deleteMutation: deleteSection
+      });
+
+      runSerial(mutationsPromises)
+        .then(() => {
+          refetchSections();
+          displayAlert('success', I18n.t('administration.sections.successSave'));
+        })
+        .catch((error) => {
+          displayAlert('danger', `${error}`, false, 30000);
```
cedricmessiant: I wonder if I hadn't changed the timeout value to `30000` for debugging purpose and made a mistake by commited it.

cedricmessiant: well it seems not according to blame tab

assembl/static2/js/app/components/common/navigationMenu.jsx
```diff
@@ -118,7 +119,8 @@ const mapStateToProps = (state) => {
     isAdmin: connectedUserIsAdmin(),
     debate: state.debate,
     phase: state.phase,
-    i18n: state.i18n
+    i18n: state.i18n,
+    sections: sectionsMock // TODO: will be fetched from a graphql query
```
cedricmessiant: `sectionsMock` !?

assembl/static2/js/app/components/common/sectionsMock.jsx
```diff
@@ -0,0 +1,46 @@
...
+  {
+    id: '6',
+    title: 'My other custom section',
+    url: 'http://www.example.com/2',
+    type: 'CUSTOM'
+    // order: 4
+  }
+];
+
+export default sections;
```
cedricmessiant: this file should be removed

assembl/static2/js/app/pages/administration.jsx
```diff
@@ -68,6 +71,10 @@ class Administration extends React.Component {
       this.putResourcesInStore(nextProps.resources);
     }
 
+    if (nextProps.sections !== this.props.sections) {
```
cedricmessiant: not sure about this one (same for resources above), @vincentfretin as they are deep structures, the 2 arrays will always be different?

assembl/static2/js/app/pages/administration.jsx
```diff
@@ -265,6 +295,14 @@ export default compose(
       };
     }
   }),
+  graphql(SectionsQuery, {
+    props: ({ data }) => {
```
cedricmessiant: you have to handle loading and errors here

assembl/static2/js/app/reducers/adminReducer/adminSections.js
```diff
@@ -0,0 +1,125 @@
...
+  UPDATE_SECTION_URL,
+  TOGGLE_EXTERNAL_PAGE,
+  CREATE_SECTION,
+  DELETE_SECTION,
+  UP_SECTION,
+  DOWN_SECTION
+} from '../../actions/actionTypes';
+import { updateInLangstringEntries } from '../../utils/i18n';
+
+export const sectionsHaveChanged = (state: boolean = false, action: ReduxAction<Action>) => {
```
cedricmessiant: missing type for return value (boolean). You can also create a type for the function, I find it more readable and you can use it before combineReducer to type the whole adminSections reducer. See examples here: https://github.com/assembl/assembl/commit/26b3ca97e4ff343a9867d7d7ed78686b27943fcb

assembl/static2/js/app/reducers/adminReducer/adminSections.js
```diff
@@ -0,0 +1,125 @@
...
+  }
+};
+
+export const sectionsInOrder = (state: List<number> = List(), action: ReduxAction<Action>) => {
+  switch (action.type) {
+  case CREATE_SECTION:
+    return state.push(action.id);
+  case UPDATE_SECTIONS: {
+    const sections = action.sections.sort((a, b) => {
+      const aOrder = a.order;
```
cedricmessiant: you probably don't need the intermediate `const` here

assembl/static2/js/app/reducers/adminReducer/adminSections.js
```diff
@@ -0,0 +1,125 @@
...
+  case DOWN_SECTION: {
+    const idIndex = state.findIndex((item) => {
+      return item === action.id;
+    });
+    const nextId = state.get(idIndex + 1);
+    const newState = state.set(idIndex, nextId);
+    const newState2 = newState.set(idIndex + 1, action.id);
+    return newState2;
+  }
+  default:
```
cedricmessiant: I would manage `DELETE_SECTION` here too to avoid problems with move up/down (see my comment above)

assembl/static2/js/app/pages/administration.jsx
vincentfretin: This is an identity check here, this checks if this is the same array instance, not the content of the array.

assembl/static2/js/app/reducers/adminReducer/adminSections.js
```diff
@@ -0,0 +1,125 @@
...
+      return aOrder - bOrder;
+    });
+    return List(
+      sections.map((s) => {
+        return s.id;
+      })
+    );
+  }
+  case UP_SECTION: {
+    const idIndex = state.findIndex((item) => {
```
cedricmessiant: I find this more clear:

```var x = Immutable.List.of('a', 'b', 'c', 'd', 'e')
var id = 'd'
var idx = x.indexOf(id)
var result = x.delete(idx).insert(idx - 1, id)
```

https://jsfiddle.net/rhnzg6ug/

cedricmessiant: (just added firebug to the js fiddle so you can test it/play with it)

cedricmessiant: (hum my code is for `MOVE_DOWN_SECTION` but you get the point)

assembl/static2/js/app/reducers/adminReducer/adminSections.js
```diff
@@ -0,0 +1,125 @@
...
+  case DELETE_SECTION:
+    return state.setIn([action.id, 'toDelete'], true);
+  case UPDATE_SECTION_URL:
+    return state.setIn([action.id, 'url'], action.value);
+  case TOGGLE_EXTERNAL_PAGE:
+    return state.updateIn([action.id, 'url'], (url) => {
+      if (url !== null) {
+        return null;
+      }
+      return '';
```
cedricmessiant: if `url` was `null` you set it to an empty string?

assembl/static2/js/app/reducers/adminReducer/adminSections.js
```diff
@@ -0,0 +1,59 @@
+// @flow
+import { combineReducers } from 'redux';
+import type ReduxAction from 'redux';
```
swergas: What is ReduxAction? I don't see it in the lib except in a test: https://github.com/reactjs/redux/search?utf8=%E2%9C%93&q=ReduxAction&type=

assembl/static2/css/components/assembl-glyph.scss
vincentfretin: @paolina92 please follow the doc `doc/fontello.rst`, you need to add new icons in the array in `styleguide/icons.jsx`
And be careful, @swergas modified the font too in #246

assembl/static2/js/app/components/administration/discussion/manageSectionsForm.jsx
```diff
@@ -35,4 +43,13 @@ const mapStateToProps = (state) => {
   };
 };
 
-export default connect(mapStateToProps)(ManageSectionsForm);
\ No newline at end of file
+const mapDispatchToProps = (dispatch) => {
+  return {
+    createSection: (nextOrder) => {
+      const newId = Math.round(Math.random() * -1000000).toString();
```
swergas: It feels weird to find this line at this place. It is either not the right place (should be in a utility method), or probably just not needed? When you create a new element you should not have to worry about its id as it is not set yet and will be decided by the backend. For example why just not provide `-1`? It would not work if we try to create several elements?

assembl/static2/fonts/assembl-glyph.svg
```diff
@@ -34,8 +34,12 @@
 
 <glyph glyph-name="minus-circled" unicode="&#xe80d;" d="M420 770q174 0 297-123t123-297-123-297-297-123-297 123-123 297 123 297 297 123z m252-368l-504 0 0-102 504 0 0 102z" horiz-adv-x="840" />
 
+<glyph glyph-name="up-bold" unicode="&#xe80e;" d="M760 336l-192 0 0-350-376 0 0 350-192 0 380 380z" horiz-adv-x="760" />
```
swergas: oops, we have 2 branches (this one and piwik metrics) that add new icons to the fontello font, so there will probably be a problem after merge: we will probably have to re-export the fontello font after the 2 merges

assembl/static2/js/app/components/administration/discussion/manageSectionsForm.jsx
vincentfretin: The id needs to be unique in the frontend, so we generate a random negative number that will be overriden by the postive id that the backend returns.

assembl/static2/js/app/reducers/adminReducer/adminSections.js
```diff
@@ -61,7 +63,30 @@ export const sectionsById = (state: Map<string, Map> = Map(), action: ReduxActio
...
+          isNew: false,
+          order: count,
+          id: s.get('id'),
+          titleEntries: s.get('titleEntries'),
+          url: s.get('url'),
+          type: s.get('type')
+        });
+        newState = newState.set(s.get('id'), sectionInfo);
+      }
+    });
```
swergas: really not sure this is a good way to change the order value of all remaining items

assembl/static2/js/app/reducers/adminReducer/adminSections.js
```diff
@@ -67,32 +67,16 @@ export const sectionsById = (state: Map<string, Map> = Map(), action: ReduxActio
...
-          toDelete: false,
-          isNew: s.get('isNew'),
-          order: count,
-          id: s.get('id'),
-          titleEntries: s.get('titleEntries'),
-          url: s.get('url'),
-          type: s.get('type')
-        });
-        newState = newState.set(s.get('id'), sectionInfo);
+        newState = newState.setIn([s.get('id'), 'order'], count);
```
swergas: much better :) forget my previous comment https://github.com/assembl/assembl/pull/250/commits/9a7e4fd60bffeec6ecd2ce6cc13929dd6937d06e#r154046378

vincentfretin: if you do a review commit by commit, please compare with the latest diff of the PR before posting the comment ;)

assembl/static2/js/app/reducers/adminReducer/adminSections.js
cedricmessiant: it changed a lot since, we don't use `order` in `sectionsById` state anymore (everything related to order is done in `sectionsInOrder`). We just reset the order of each section based on index in `sectionsInOrder` before to send the mutations

assembl/static2/js/app/reducers/adminReducer/adminSections.js
cedricmessiant: From what I remember, I had problems trying to type reducers and found this in code with types on reducers. I saw it as a wrapper that says it is any of the redux "default" actions (`@@INIT` and so) or any type in `Action`
According to flow doc https://flow.org/en/docs/react/redux/ it is not necessary (anymore?) so I will try to remove this after the merges. Thanks for pointing that out

cedricmessiant: https://github.com/reactjs/redux/issues/290

assembl/static2/js/app/components/administration/discussion/manageSectionsForm.jsx
cedricmessiant: yes, we need an id for the redux states, it ease the work. I agree that we could move it somewhere in `utils`

assembl/static2/js/app/components/administration/discussion/editSectionForm.jsx
paolina92: Done

assembl/static2/js/app/components/administration/discussion/editSectionForm.jsx
paolina92: L'oeil de lynx

assembl/static2/js/app/components/administration/discussion/editSectionForm.jsx
paolina92: Allez

assembl/static2/js/app/components/common/navigationMenu.jsx
paolina92: Oui t'avais mis ça mais finalement Norman l'a fait sur une autre branche.

assembl/static2/js/app/pages/administration.jsx
paolina92: yes

assembl/static2/js/app/reducers/adminReducer/adminSections.js
paolina92: Yes

assembl/static2/js/app/reducers/adminReducer/adminSections.js
swergas: good idea, I had tried without success on previous branches but now that you do a bit smaller PRs they load fully in github :p

assembl/models/section.py
ayyazdaniaryan: Agreed. We have consistently used the polymorphic_type on base objects so that we would always have row representation of the type of object instantiated. It's non-critical for now. Can be done ad-hoc down the road.

assembl/static2/js/app/components/administration/saveButton.jsx
ayyazdaniaryan: So the error alert is shown for 30s? It's a bit much.

assembl/graphql/schema.py
```diff
@@ -189,6 +191,12 @@ def resolve_locales(self, args, context, info):
     def resolve_resources_center(self, args, context, info):
         return ResourcesCenter()
 
+    def resolve_sections(self, args, context, info):
+        model = models.Section
+        query = get_query(model, context)
+        discussion_id = context.matchdict['discussion_id']
+        return query.filter(model.discussion_id == discussion_id)
```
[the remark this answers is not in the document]
vincentfretin: Please... do the sort in sql `.order_by(model.order)`. But anyway, I guess the sort is done frontend side @cedricmessiant ?

assembl/graphql/section.py
```diff
@@ -0,0 +1,160 @@
...
+                # Better to have this message than
+                # 'NoneType' object has no attribute 'owner_object'
+                # when creating the saobj below if title=None
+
+            title_langstring = langstring_from_input_entries(title_entries)
+
+            kwargs = {}
+            kwargs['section_type'] = args.get('section_type', 'CUSTOM')
+            kwargs['url'] = args.get('url')
+            kwargs['order'] = args.get('order', 10.0)
```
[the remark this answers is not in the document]
vincentfretin: The 10.0 here is a default value.

assembl/graphql/section.py
```diff
@@ -0,0 +1,160 @@
...
+        discussion_id = context.matchdict['discussion_id']
+        user_id = context.authenticated_userid or Everyone
+
+        section_id = args.get('section_id')
+        section_id = int(Node.from_global_id(section_id)[1])
+        section = models.Section.get(section_id)
+        if section.section_type != SectionTypesEnum.CUSTOM.value:
+            return DeleteSection(success=False)
+
+        permissions = get_permissions(user_id, discussion_id)
```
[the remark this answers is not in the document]
vincentfretin: hum? You need the object to do the permission check.

assembl/static2/css/components/assembl-glyph.scss
```diff
@@ -1,12 +1,12 @@
 @font-face {
   font-family: 'assembl-glyph';
-  src: url('#{$fonts-dir}/assembl-glyph.eot?64503040');
+  src: url('#{$fonts-dir}/assembl-glyph.eot?76240933');
```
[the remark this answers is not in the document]
vincentfretin: It's just the value here from the downloaded zip. I think it's some sort of timestamp. Anyway, it's just a string, doesn't matter what it is, but it just need to be unique. The easiest way is just to increment it.

assembl/static2/js/app/components/administration/discussion/editSectionForm.jsx
```diff
@@ -0,0 +1,135 @@
...
+  index: number,
+  handleTitleChange: Function,
+  handleUrlChange: Function,
+  toggleExternalPageField: Function,
+  handleDeleteClick: Function,
+  handleDownClick: Function,
+  handleUpClick: Function
+};
+
+const DumbEditSectionForm = ({
```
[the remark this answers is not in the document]
vincentfretin: dumb vs smart (container, via connect), this is the terminology used in redux docs and videos. Dumb takes only props.

assembl/static2/css/components/assembl-glyph.scss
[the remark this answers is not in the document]
paolina92: I confirm, it's from the download zip

assembl/static2/js/app/components/administration/discussion/editSectionForm.jsx
paolina92: yes

assembl/static2/js/app/components/administration/saveButton.jsx
paolina92: no I don't think so. This delay is necessary to do a screenshot of the error for example.

assembl/static2/js/app/components/administration/discussion/editSectionForm.jsx
cedricmessiant: from what I remember, we want to be able to change this url (some customers want the home logo to redirect to their site or something like that). Maybe check with PO ?

PR #252, opened 2017-11-28
get_post_url return v2 url if current phase use v2 interface

vincentfretin:


vincentfretin:
@ayyazdaniaryan The same thing for synthesis notification is missing. I leave you to it.

assembl/lib/frontend_urls.py
```diff
@@ -219,8 +244,9 @@ def get_post_url(self, post):
             return self.get_discussion_url() + self.get_relative_post_url(post)
         else:
             route = None
-            phase = get_current_phase_identifier(
-                self.discussion.timeline_events)
+            phase = post.get_created_phase()
+            # The created post must be created within an associated phase
+            assert phase
             first_idea = None
```
vincentfretin: In this case, remove get_current_phase_identifier completly from the code and tests. Don't keep dead code.

ayyazdaniaryan: The get_current_phase_identifier can have its uses as a utility function, no? It's tracking the time *now* to the timelines. When the post is created is unrelated to the reference of now to timeline dates.

vincentfretin: Please don't think future hypothetic use. It's not used right now, just remove it. If one day you want the function, just revert the commit.

swergas: Is get_current_phase_identifier really not used? I see it used below for a route. I'm not sure I understand your comment

vincentfretin: Please look at the latest diff, it's not used anymore.

swergas: Oh yes I saw that after, in these cases please don't comment in this commit but in the general diff or general conversation if it is possible

assembl/lib/frontend_urls.py
```diff
@@ -23,6 +25,27 @@
...
+
+def get_frontend_urls():
+    """Get all V2 routes from source of truth"""
+
+    current = dirname(__file__)
+    route_path = join(current, '..', 'static2/routes.json')
+    if not exists(route_path):
+        raise IOError("Route path could not be found")
+    with open(route_path) as paths:
+        routes = json.loads(paths.read())
```
swergas: If this function is called often then there can be an efficiency problem. For example, reading and parsing a file everytime we want to build an URL or know how to interpret an URl can result in too much I/O on disk. If it is the case, probably read the file once in the app to save routes in memory and then refer to them?

assembl/graphql/synthesis.py
```diff
@@ -27,6 +27,7 @@ class Meta:
     ideas = graphene.List(lambda: IdeaUnion)
     img = graphene.Field(Document)
     creation_date = DateTime()
+    post = graphene.Field("assembl.graphql.post.Post")
```
swergas: why not keep the name "published_in_post" for consistency and clarity?

assembl/static2/tests/unit/pages/synthesis.spec.jsx
```diff
@@ -7,6 +7,8 @@ describe('Synthesis component', () => {
   it('should match Synthesis snapshot', () => {
     const props = {
       routeParams: { slug: 'slugId' },
+      // TODO update this to reflect query change. Changing this
```
swergas: Have you tried again fixing this before merge?

assembl/graphql/synthesis.py
vincentfretin: Why keep? It's `publishes_in_post` in the sqlalchemy relationship. We should maybe rename the relationship to `published_in_post`?

assembl/static2/tests/unit/pages/synthesis.spec.jsx
vincentfretin: @ayyazdaniaryan we can look at it today together if you have time.

assembl/graphql/synthesis.py
ayyazdaniaryan: Yeah, this was an old typo, prior to the era of CRs. Happy to see the relationship updated throughout the codebase. Not in scope of this PR, however.

assembl/static2/tests/unit/pages/synthesis.spec.jsx
vincentfretin: Actually the test is fine, no need to change anything here, we are testing DumbSynthesis. I removed the comment in https://github.com/assembl/assembl/commit/8e5fb260f2978804f85d65f944e96936b1de4155

PR #253, opened 2017-11-28
Fix elastic search plugin errors in docker

n0izn0iz:


configs/docker.rc
```diff
@@ -79,3 +79,5 @@ docker_certbot =
 
 # Should we return a restrictive robots.txt?
 disallow_robots =
+
+docker_elasticsearch_version = 5.6.2
```
vincentfretin: It this possible to move this to `base_env.rc` so we can use it for both docker and fabric with `env.docker_elasticsearch_version` instead of the constant `ELASTICSEARCH_VERSION` in `fabfile.py`? This way we only have one place to change the version.

docker/elasticsearch.Dockerfile
```diff
@@ -0,0 +1,6 @@
+ARG version
+FROM docker.elastic.co/elasticsearch/elasticsearch:${version}
+ARG version
```
vincentfretin: You wrote `ARG version` twice.

ayyazdaniaryan: Baha I dun goofed for not seeing that.

n0izn0iz: On purpose, the first one declares the ARG and allows to use it in the FROM.
The second one allows to use it in the rest of the Dockerfile.
https://docs.docker.com/engine/reference/builder/#understand-how-arg-and-from-interact

vincentfretin: Oh, ok, this is not intuitive at all. :)

vincentfretin:
Ok for me, you can merge.

PR #257, opened 2017-12-01
Chat frame

n0izn0iz:
The `gen_chatbot_css.sh` script is used to generate the style that is used in the iframe (and entered on the external chatbot configuration page)


assembl/scripts/gen_chatbot_css.sh
```diff
@@ -0,0 +1,113 @@
+#!/bin/sh
```
ayyazdaniaryan: What is the script for?

Also, most things in Assembl's backend are bound to Python, and script execution to `fabric`. I'm not against `sh` scripts at all. But, for consistency, and for single execution style, this could be done in fabric quite easily. Convert the scss to `jinja2` template, and use a command similar to `docker_compose` task, to parse the jinja template with the appropriate data. OR, you can create this once in the themes folder. And parse it ad-hoc per discussion and load it via webpack.

assembl/static2/js/app/utils/translations.js
```diff
@@ -362,6 +366,10 @@ const Translations = {
     validate: "Validate",
     introduction: "Introduction",
     conclusion: "Conclusion",
+    chatframe: {
+      title: "Instant discussion window",
+      tooltip: "Assemb'Bot"
```
vincentfretin: The missing l is on purpose?

assembl/scripts/gen_chatbot_css.sh
n0izn0iz: it's to generate the css you have to paste in the chatbot config page at motion.ai

assembl/static2/js/app/components/common/ChatFrame.jsx
```diff
@@ -0,0 +1,63 @@
+import React from 'react';
+import { Translate } from 'react-redux-i18n';
+import { OverlayTrigger, Tooltip } from 'react-bootstrap';
+import { connect } from 'react-redux';
+
+const varPath = (root, path) => {
```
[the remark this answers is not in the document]
n0izn0iz: ty

assembl/static2/js/app/utils/translations.js
```diff
@@ -6,6 +6,10 @@ const Translations = {
     cancel: "Annuler",
     validate: "Valider",
     delete: "Supprimer",
+    chatframe: {
+      title: "Fenêtre de discussion instantanée",
+      tooltip: "Assemb'Bot"
```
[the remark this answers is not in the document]
n0izn0iz: Not my choice

assembl/static2/js/app/utils/translations.js
n0izn0iz: yep

assembl/static2/js/app/components/common/ChatFrame.jsx
```diff
@@ -0,0 +1,63 @@
...
+    return (
+      <div className="chatframe-modal">
+        <div className="chatframe-modal-header">
+          <span className="chatframe-icon chatframe-modal-icon assembl-icon-robot" />
+          <Translate value="chatframe.title" />
+        </div>
+        <iframe className="chatframe-modal-iframe" title="chatframe" src={src} />
+      </div>
+    );
+  };
```
vincentfretin: Why do use static component? Can't you just write them on the module scope?

assembl/scripts/gen_chatbot_css.sh
ayyazdaniaryan: Is there a motionAI API to POST this content to their config?

assembl/static2/js/app/components/common/ChatFrame.jsx
n0izn0iz: it's the chatframe modal so I figured it would be best declared in the chatframe class

assembl/scripts/gen_chatbot_css.sh
n0izn0iz: doesn't look like it https://docs.motion.ai/docs/api https://docs.motion.ai/v1.0/reference

ayyazdaniaryan: Dang, that sucks. Okay, I don't have anything against usage of shell script. But this script is not configurable enough to be used by all developers. I don't know how relevant this is. For now, this should be good enough to start. However, it needs to be documented. So, for that, I would request you add docs to this PR on usage.

n0izn0iz: `botBackgroundColor='blue' botTextColor='white" ./gen_chatbot_css.sh` is too hard to use ?

n0izn0iz: you only need to run it when you create a new bot at motion.ai

assembl/scripts/gen_chatbot_css.sh
```diff
@@ -0,0 +1,113 @@
...
+
+: ${humanBackgroundColor:='#b4f4be'}
+: ${humanTextColor:='black'}
+: ${humanTimeColor:='grey'}
+
+: ${copier:=$(
+    if command -v xclip 2>/dev/null; then :
+    elif command -v pbcopy 2>/dev/null; then :
+    elif command -v less 2>/dev/null; then :
+    else echo cat fi
```
vincentfretin: `:` is the bash built-in for true, right?
Why this complex conditions instead of the following:
`command -v xclip 2>/dev/null || command -v pbcopy 2>/dev/null || command -v less 2>/dev/null || echo cat`
It should be equivalent I think.

n0izn0iz: it's the bash builtin for noop but it's the same as true I think

assembl/scripts/gen_chatbot_css.sh
vincentfretin: For those not familiar with bash scripting, this may not be obvious you need to add variables before the command. For me it is. :)
You should document the usage of this script in a comment at the top of the script. And that for variables that are not defined, default values in this script are used.

assembl/scripts/gen_chatbot_css.sh
n0izn0iz: I don't think I can replace with an or chain, it's not the same behaviour as javascript:
```
(venv)  ✘ Norman@bluenove-44  ~/assembl   motion-chatbot-own-frame  false || true || echo bli && false || echo bla
bla
```

n0izn0iz: your version of the switch seems to work, will replace with it. I wrote it like this because sh && and || is a bit weird some times

assembl/scripts/gen_chatbot_css.sh
n0izn0iz: doc or comment then ?

assembl/static2/js/app/components/common/ChatFrame.jsx
n0izn0iz: you prefer it that way?

assembl/scripts/gen_chatbot_css.sh
vincentfretin: If you have a specific doc to create the chatbot, describe it in the doc then.

assembl/static2/js/app/components/common/ChatFrame.jsx
vincentfretin: I'd prefer in the module scope yes, like we do in other modules. In some other modules, we define several non exported components on the module scope. I don't know why it should be different here. Let's use the same style everywhere.

assembl/static2/js/app/components/common/ChatFrame.jsx
```diff
@@ -2,12 +2,7 @@ import React from 'react';
 import { Translate } from 'react-redux-i18n';
 import { OverlayTrigger, Tooltip } from 'react-bootstrap';
 import { connect } from 'react-redux';
-
-const varPath = (root, path) => {
-  return path.split('.').reduce((value, nextKey) => {
-    return value && value[nextKey];
-  }, root);
-};
+import lodashGet from 'lodash/get';
```
ayyazdaniaryan: Just `get` would do 😀

doc/consultant_notes.rst
```diff
@@ -22,3 +22,34 @@ only use them like described here.
...
+  
+Valid values are css colors. If the output is not printed to the terminal, it should be copied to your clipboard.
+
+Go to the motion.ai website, click on Webchat Settings, go to Customize tab, go to CSS subtab
+ 
+Replace the code in the text area with the generated one
+ 
+Go to deploy tab, copy the link in the 'Frame the webchat url yourself' section
+
+Go to v1 administration of the debate, Discussion preferences section, add a
```
ayyazdaniaryan: Discussion preferences section, under the extra json field, add the following key.

doc/consultant_notes.rst
```diff
@@ -22,3 +22,34 @@ only use them like described here.
...
+::
+  botBackgroundColor
+  botTextColor
+  botTimeColor
+  humanBackgroundColor
+  humanTextColor
+  humanTimeColor
+  
+Valid values are css colors. If the output is not printed to the terminal, it should be copied to your clipboard.
+
```
ayyazdaniaryan: Instead of spaces like this, you can use bullets in rst, since your language is very sparse.

```
* Point 1
* Point 2
* Point 3```

assembl/static2/js/app/components/common/ChatFrame.jsx
n0izn0iz: we already have a `get` function used everywhere, I think this would have caused confusion at some point

PR #261, opened 2017-12-04
Collapsed answer form

paolina92:


assembl/static2/js/app/components/debate/thread/post.jsx
```diff
@@ -49,7 +49,7 @@ export class EmptyPost extends React.PureComponent {
   }
 
   componentDidMount() {
-    this.props.measureTreeHeight(400);
+    this.props.measureTreeHeight(480);
```
ayyazdaniaryan: Could you please make this delay constant with a comment as to why it is the number that it is?

assembl/static2/js/app/components/debate/thread/answerForm.jsx
```diff
@@ -43,9 +47,9 @@ class AnswerForm extends React.PureComponent<*, AnswerFormProps, AnswerFormState
       submitting: false
     };
   }
-
   handleCancel = () => {
-    this.props.hideAnswerForm();
+    const { hideAnswerForm } = this.props;
+    this.setState({ body: null }, hideAnswerForm());
```
ayyazdaniaryan: By doing this, does the previous `draftState` object get garbage collected? Wondering if this could lead to a leak?

assembl/static2/js/app/components/debate/thread/answerForm.jsx
```diff
@@ -43,9 +47,9 @@ class AnswerForm extends React.PureComponent<*, AnswerFormProps, AnswerFormState
       submitting: false
     };
   }
-
```
ayyazdaniaryan: From previous discussion with @cedricmessiant @vincentfretin, the team consensus was to have spaces between methods in JS. We should make this a rule as part of the linter so that it catches it.

assembl/static2/js/app/components/debate/thread/answerForm.jsx
```diff
@@ -55,9 +59,22 @@ class AnswerForm extends React.PureComponent<*, AnswerFormProps, AnswerFormState
   };
 
   handleInputFocus = () => {
-    const isUserConnected = getConnectedUserId(); // TO DO put isUserConnected in the store
-    if (!isUserConnected) {
-      inviteUserToLogin();
+    const { debateData, identifier, handleAnswerClick } = this.props;
```
ayyazdaniaryan: Where is identifier coming from into the props?

assembl/static2/js/app/components/debate/thread/post.jsx
paolina92: Yes sure. I'm not really sure to understand the reason of this value. I asked to @cedricmessiant via Slack.

assembl/static2/js/app/components/debate/thread/answerForm.jsx
paolina92: Sorry, I don't understand what "garbage" means :(

assembl/static2/js/app/components/debate/thread/answerForm.jsx
paolina92: Same here. I'm not sure to understand :(

assembl/static2/js/app/components/debate/thread/post.jsx
paolina92: But for sure, it's not a delay. It's a height value I guess.

assembl/static2/js/app/components/debate/thread/answerForm.jsx
paolina92: Yes

assembl/static2/js/app/components/debate/thread/post.jsx
vincentfretin: ah no, don't modify this, it's not a height, it's a delay of 400ms!

assembl/static2/js/app/components/debate/thread/answerForm.jsx
vincentfretin: @ayyazdaniaryan Why wouldn't it be garbage collected? There is no reference to it, right? The only reference is in state.

cedricmessiant: @paolina92 https://developer.mozilla.org/fr/docs/Web/JavaScript/Gestion_de_la_m%C3%A9moire

assembl/static2/js/app/components/debate/thread/answerForm.jsx
```diff
@@ -55,9 +59,22 @@ class AnswerForm extends React.PureComponent<*, AnswerFormProps, AnswerFormState
...
+    const isPhaseCompleted = getIfPhaseCompletedByIdentifier(debateData.timeline, identifier);
+    if (isPhaseCompleted) {
+      const body = (
+        <div>
+          <Translate value="debate.noAnswer" />
+        </div>
+      );
+      displayModal(null, body, true, null, null, true);
+    } else {
+      const isUserConnected = getConnectedUserId(); // TO DO put isUserConnected in the store
```
cedricmessiant: you should use the `promptForLoginOr` util here so we wont have to search this code everywhere if the TODO is adressed one day

assembl/static2/js/app/components/debate/thread/post.jsx
```diff
@@ -301,17 +300,17 @@ export class EmptyPost extends React.PureComponent {
...
-            />
-          </div>
-        ) : null}
+        <div className={this.state.showAnswerForm ? 'answer-form' : 'collapsed-answer-form'}>
+          <AnswerForm
+            parentId={id}
+            ideaId={ideaId}
+            refetchIdea={refetchIdea}
+            textareaRef={answerTextareaRef}
+            hideAnswerForm={this.hideAnswerForm}
```
cedricmessiant: are you sure `hideAnswerForm` is still used ?

paolina92: Yes It's still used

PR #264, opened 2017-12-05
Discussion logo url

ayyazdaniaryan:
This is a quick fix required for EDF, there is no user story associated to it. However, it used to be the expected behaviour in Assembl V1 that is now added to Assembl V2.

### DIFF
--- a/assembl/graphql/discussion.py
+++ b/assembl/graphql/discussion.py
@@ -1,6 +1,7 @@
 import os.path
 
 import graphene
+from graphene_sqlalchemy import SQLAlchemyObjectType
 from pyramid.httpexceptions import HTTPUnauthorized
 from pyramid.security import Everyone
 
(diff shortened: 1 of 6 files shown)


ayyazdaniaryan:
@cedricmessiant @vincentfretin I was thinking of putting the Query request into the `TabConditions` query at first, however, that query occurs in a child component to `NavBar`. The scope of this fix was extremely minimal. I wanted to avoid having to edit the tabs code to hoist the query to NavBar and pass it as a child. However, I am in favor of having a single query to fetch all navbar based resources.

vincentfretin:
You don't need to change the code, just do the same query. It's dedup automatically by apollo and the query is only done once.

maparent:
Note: I think this is an improvement over previous state. However, to discuss with UX team: I think we may want two distinct tabs for our home and the external home.

PR #266, opened 2017-12-06
Mediactive login

maparent:
New social-auth option to work with mediactive. Tests TBD.
Many changes to social login pipeline, because we cannot assume the user came to the social login through assembl; so discussion or user may no be set, or may not be correct.

### DIFF
--- a/assembl/auth/encryption.py
+++ b/assembl/auth/encryption.py
@@ -1,0 +1,78 @@
+import base64
+
+from cryptography.hazmat.backends import default_backend
+from cryptography.hazmat.primitives import padding, hashes, serialization
+from cryptography.hazmat.primitives.ciphers import Cipher, algorithms, modes
+
+from .make_saml import private_key_from_cleaned_text
+
+
+class Encrypter(object):
+    def encrypt(self, message):
+        raise NotImplementedError()
+
+
+class Decrypter(object):
+    def decrypt(self, message):
+        raise NotImplementedError()
+
+
+class AESDecrypter(Decrypter):
+    def __init__(self, password, backend=None):
+        backend = backend or default_backend()
+        self.backend = backend
+        self.password = password
+
+    def decrypt(self, message):
+        iv = base64.b64decode(message[:24])
+        cipher = Cipher(
+            algorithms.AES(self.password),
+            modes.CBC(iv), backend=self.backend)
+        message = base64.b64decode(message[24:])
+        decryptor = cipher.decryptor()
+        padded = decryptor.update(message)
+        padded += decryptor.finalize()
+        unpadder = padding.PKCS7(128).unpadder()
+        decrypted = unpadder.update(padded)
+        decrypted += unpadder.finalize()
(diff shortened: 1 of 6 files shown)


assembl/auth/encryption.py
```diff
@@ -0,0 +1,77 @@
...
+
+class AESDecryptor(Decryptor):
+    """A decryptor that uses AES symmetric keys"""
+    def __init__(self, password, backend=None):
+        backend = backend or default_backend()
+        self.backend = backend
+        self.password = password
+
+    def decrypt(self, message):
+        iv = base64.b64decode(message[:24])
```
ayyazdaniaryan: `24` should be a constant with a comment for maintainability.

assembl/auth/encryption.py
```diff
@@ -0,0 +1,77 @@
...
+        cipher = Cipher(
+            algorithms.AES(self.password),
+            modes.CBC(iv), backend=self.backend)
+        message = base64.b64decode(message[24:])
+        decryptor = cipher.decryptor()
+        padded = decryptor.update(message)
+        padded += decryptor.finalize()
+        unpadder = padding.PKCS7(128).unpadder()
```
ayyazdaniaryan: So should the padder int.

assembl/auth/encryption.py
```diff
@@ -0,0 +1,77 @@
...
+
+class AESEncryptor(Encryptor):
+    """An encryptor that uses AES symmetric keys"""
+    def __init__(self, password, backend=None):
+        backend = backend or default_backend()
+        self.backend = backend
+        self.password = password
+
+    def encrypt(self, message):
+        iv = urandom(16)
```
ayyazdaniaryan: The bit counts should also be const. Likely as a member variable on the `Encryptor` root class.

assembl/auth/mediactive.py
```diff
@@ -0,0 +1,43 @@
...
+    name = 'mediactive'
+    USERNAME_KEY = 'username'
+    html_parser = HTMLParser()
+
+    def auth_url(self):
+        """Must return redirect URL to auth provider"""
+        next_url = self.data.get('next', None)
+        base = self.setting('LOGIN_URL')
+        if next_url:
+            base += '?' + urlencode({'next': next_url})
```
ayyazdaniaryan: Use `urljoin` and `urlencode` to ensure flexible usage of the config file under the `LOGIN_URL` key.

maparent: urlencode I'm using... not sure what urljoin is adding anything to this. 
Unless you mean urlparse/urldefrag?

maparent: I wrote it for the heck of it, but I think it's overkill. Tell me if you want it.
```python
        if next_url:
            pr = urlparse(base)
            qs = parse_qs(pr.query)
            qs['next'] = [next_url]
            base = urlunparse(pr._replace(query=urlencode(qs, True)))
```

PR #268, opened 2017-12-11
Fix notification mails links when debate is using v2

n0izn0iz:
https://bluenove.atlassian.net/browse/DEVAS-1040

I'm not sure this will fix it

### DIFF
--- a/assembl/lib/frontend_urls.py
+++ b/assembl/lib/frontend_urls.py
@@ -220,7 +220,10 @@ class FrontendUrls(object):
         # this method is kept mostly for legacy routes that do not exist in
         # new front-end yet.
         if request is None:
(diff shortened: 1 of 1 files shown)


vincentfretin:
This should fix it indeed.

n0izn0iz:
I can't find a way to check the content of outbound emails locally :(

ayyazdaniaryan:
@n0izn0iz Here's how you can do it:
- You'll have to do a bit of database inspection to find the Notification ID object associated with a post created, then use the `/data/Notification/<id>/mail_html_preview` to view the notification message.

n0izn0iz:
nofix

n0izn0iz:
Actually, request is not None when mails are rendered in the browser views but it's None in other cases so we need both

PR #271, opened 2017-12-12
add a sentiment vote and share for column synthesis

vincentfretin:


assembl/alembic/versions/c3f8bc9c75d5_column_synthesis_post.py
```diff
@@ -0,0 +1,73 @@
...
+from assembl.lib.sqla import mark_changed
+
+
+def upgrade(pyramid_env):
+    # Do stuff with the app's models here.
+    from assembl import models as m
+    db = m.get_session_maker()()
+    with transaction.manager:
+        # take the first sysadmin as creator
+        creator_id = m.User.default_db.query(m.User).join(m.User.roles).filter(m.Role.id == 7)[0:1][0].id
```
cedricmessiant: `[0:1][0]` !?

assembl/static2/js/app/components/debate/multiColumns/tabbedColumns.jsx
```diff
@@ -28,8 +29,8 @@ export default class TabbedColumns extends React.Component {
     const col = messageColumns[index];
     const synthesisProps = showSynthesis && {
       classifier: activeKey,
-      synthesisTitle: I18n.t('multiColumns.synthesis.title', { colName: col.name }),
-      synthesisBody: col.header || I18n.t('multiColumns.synthesis.noSynthesisYet'),
+      synthesisTitle: get(col, 'columnSynthesis.subject') || I18n.t('multiColumns.synthesis.title', { colName: col.name }),
```
cedricmessiant: why don't you use the third param of `lodash.get` (default value) ?

assembl/alembic/versions/c3f8bc9c75d5_column_synthesis_post.py
vincentfretin: [0:1] will do a offset 0 limit 1 in the sql query.

assembl/static2/js/app/components/debate/multiColumns/tabbedColumns.jsx
vincentfretin: Oh right, I forgot about that.

vincentfretin: done in https://github.com/assembl/assembl/commit/af5de5f637e848283f0ea55af700483a5094ddb6

assembl/alembic/versions/c3f8bc9c75d5_column_synthesis_post.py
ayyazdaniaryan: Using `7` as the role.id also feels wrong. I suspect it won't be an issue across servers due to how the servers are bootstrapped, but the query should still be stable. Find the ID of of the `r:sysadmin` and apply here. And after your comment, I still do not understand the `[0:1][0]`.

assembl/models/idea_content_link.py
```diff
@@ -50,7 +50,7 @@ class IdeaContentLink(DiscussionBoundBase):
 
     # This is nullable, because in the case of extracts, the idea can be
     # attached later.
-    idea_id = Column(Integer, ForeignKey(Idea.id),
+    idea_id = Column(Integer, ForeignKey(Idea.id, ondelete="CASCADE", onupdate="CASCADE"),
```
ayyazdaniaryan: Doesn't this have to be defined in a migration as well to be a part of the sql schema?

assembl/models/idea_msg_columns.py
```diff
@@ -129,5 +124,52 @@ def ensure_ordering(cls):
...
             cls.ensure_ordering_for_idea(idea)
 
+    def get_column_synthesis(self):
+        synthesis = self.db.query(ColumnSynthesisPost
+            ).join(ColumnSynthesisPost.idea_links_of_content
+            ).filter(
+                IdeaContentLink.idea_id == self.idea_id,
+                ColumnSynthesisPost.tombstone_date == None,
+                Content.message_classifier == self.message_classifier
+            ).all()
```
ayyazdaniaryan: Why not use `.first()`, which would return the synthesis or `None` ?

assembl/models/idea_msg_columns.py
```diff
@@ -129,5 +124,52 @@ def ensure_ordering(cls):
...
+    def set_column_synthesis(self, subject=None, body=None, creator_id=None):
+        synthesis = self.get_column_synthesis()
+        if synthesis is None:
+            user_id = creator_id or get_current_request().authenticated_userid
+            synthesis = ColumnSynthesisPost(
+                message_classifier=self.message_classifier,
+                discussion_id=self.idea.discussion_id
+            )
+            self.db.add(synthesis)
+            idea_post_link = IdeaRelatedPostLink(
```
ayyazdaniaryan: Ooh, I'm not so sure about this. SynthesisPosts don't have `IdeaContentLinks`. Adding this will also cause the synthesis post of a column to be counted as a post (should it?).

assembl/static2/js/app/components/debate/multiColumns/columnsView.jsx
```diff
@@ -14,21 +16,15 @@ class ColumnsView extends React.Component {
...
+          <TabbedColumns {...this.props} showSynthesis={showSynthesis} />
         ) : (
-          <MultiColumns
-            {...this.props}
-            width={`${100 / columns.length}%`}
-            showSynthesis={showSynthesis}
-            identifier={identifier}
-            debateData={debateData}
-          />
+          <MultiColumns {...this.props} width={`${100 / columns.length}%`} showSynthesis={showSynthesis} />
```
ayyazdaniaryan: `width` as a string - can't it be an integer and then typecasted at the lowest-level `style` parameter?

assembl/models/idea_msg_columns.py
ayyazdaniaryan: @maparent tagging you because this one needs attention. We are replaceing the langstring `header_id` in idea message columns with a foreign key to a post, which will be have a subject, body, sentiments, etc. But it is a representation of a Synthesis. Should this post have an idea content link, or should it behave like a `SynthesisPost` and avoid having links?

vincentfretin: The new ColumnSynthesisPost inherits from Content, not Post, to avoid to be counted and be retrieved as a top post. IdeaRelatedPostLink inherits from IdeaContentPositiveLink, this can be used on a Content object.

assembl/models/idea_msg_columns.py
vincentfretin: I forgot about `.first()`, yeah you can probably use it here.

assembl/models/idea_content_link.py
vincentfretin: There may be a migration missing yes. Can you add it? The change here was mainly to fix the tests. In reality ideas are not removed from the database.

assembl/alembic/versions/c3f8bc9c75d5_column_synthesis_post.py
vincentfretin: I think `[0:1][0]` is exactly the same as `.first()` actually, does `first()` add a limit 1 in the query ?

assembl/static2/js/app/components/debate/multiColumns/columnsView.jsx
cedricmessiant: this one wasn't part of this commit but I think it is ok to pass a specific `width` prop since we are managing columns here (and it's a percentage).

assembl/models/idea_msg_columns.py
cedricmessiant: done

assembl/alembic/versions/c3f8bc9c75d5_column_synthesis_post.py
cedricmessiant: I did the change about the role id (on my instance, sysadmin id was 4)

cedricmessiant:
sorry @ayyazdaniaryan I merged before I saw your comments but I did the changes afterwards

assembl/static2/js/app/components/debate/common/postActions.jsx
```diff
@@ -147,4 +167,5 @@ class PostActions extends React.Component {
   }
 }
 
+// $FlowFixMe
```
swergas: what is this?

assembl/static2/js/app/components/debate/multiColumns/postColumn.jsx
```diff
@@ -3,33 +3,12 @@ import React from 'react';
 
 import Tree from '../../common/tree';
 import ColumnHeader from './columnHeader';
-import ColumnsPost from '../../../components/debate/multiColumns/columnsPost';
 import FoldedPost from '../common/post/foldedPost';
-import BoxWithHyphen from '../../common/boxWithHyphen';
+import ColumnsPost from '../../../components/debate/multiColumns/columnsPost';
+import ColumnSynthesis, { type Props as ColumnSynthesisProps } from './columnSynthesis';
```
swergas: why not name `Props` directly `ColumnSynthesisProps` in `columnSynthesis.jsx`?

assembl/models/idea_content_link.py
ayyazdaniaryan: Yeah, I'll add a migration for this.

assembl/static2/js/app/components/debate/common/postActions.jsx
cedricmessiant: it is the ignore flag for flow. Apparently, flow types are not handled correctly in this case (it is related to the types def in react-apollo)

assembl/static2/js/app/components/debate/multiColumns/postColumn.jsx
cedricmessiant: In the end, I prefer to use the shorter `Props` and `State` names next to the component (the short version is "light" and we rarely need to import types from other components).

PR #284, opened 2017-12-18
EDF - Home video no autoplay

n0izn0iz:
You need to add a `video.posterUrl` value in the extra_json
for edf it would be:
`"posterUrl": "https://edf-content.cdn.mediactive-network.net/images/poster-player-EDF.png"`

### DIFF
--- a/assembl/static2/js/app/components/home/video.jsx
+++ b/assembl/static2/js/app/components/home/video.jsx
@@ -82,6 +82,8 @@ class Video extends React.Component {
     const videoUrl = get(debateData, 'video.videoUrl', '');
     const posterUrl = get(debateData, 'video.posterUrl', '');
     const isVideoFile = EDFHacks.srcIsVideoFile(videoUrl);
+    const posterProps = {};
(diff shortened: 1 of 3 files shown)


assembl/static2/js/app/components/home/video.jsx
```diff
@@ -2,6 +2,60 @@ import React from 'react';
...
+  'mxf',
+  'roq',
+  'nsv',
+  'f4v',
+  'f4p',
+  'f4a',
+  'f4b'
+];
+
+const EDFHacks = {
```
maparent: This part is not EDF-specific, though it was done for them. So EDF should not be part of the name.
We can and should do better to check file vs frame, but this is still a vast improvement from the previous code; so I'm not even sure it should be called a hack as such. (So how about NaiveVideoFileFrameDetector?)
In other words, this should be written in a style that can go to develop.

I had to redo the work for media; ideally there should have been a Media component which can receive the video url and poster as props, and the nature of the url (video/frame/autodetect, with autodetect as default?) and is shared between home and thematic.
But that needs to be scheduled properly.

PR #285, opened 2017-12-19
Do not display non-important extracts

n0izn0iz:


assembl/static2/js/app/components/debate/thread/nuggets.jsx
```diff
@@ -47,8 +47,8 @@ class Nuggets extends React.Component {
   render() {
     const { extracts } = this.props;
     const { top } = this.state;
-
-    return Array.isArray(extracts) && extracts.length > 0 ? (
+    const importantExtracts = Array.isArray(extracts) && extracts.filter(({ important }) => important);
+    return importantExtracts && importantExtracts.length > 0 ? (
```
cedricmessiant: I think that ` importantExtracts.length > 0` is useless. I would probably return null if `!importantExtracts` (here I guess that your render method returns `undefined` but maybe it's not an issue)

n0izn0iz: I don't think I can remove the `length > 0` check because it would render a diamond with no extracts.

The render method shouldn't return undefined since I use a ternary to return the component or null, I should maybe change this since it's not super clear.

n0izn0iz: An empty array is not truthy, right ?

cedricmessiant:
```[] ? 'true' : 'false'
"true"```

oh, sorry, I still live in python world ;-) :confused:

PR #286, opened 2017-12-19
Youtube theater mode

n0izn0iz:
This adds 'react-active-html ' to the depencies to convert posts html into react components and replace youtube iframes with youtube components (also remove the use of dangerouslySetInnerHtml in PostBody)


assembl/static2/js/app/components/common/youtubeTheater.jsx
```diff
@@ -0,0 +1,52 @@
...
+
+  timeout: number;
+
+  closeTheater = () => this.setState({ theaterMode: false });
+
+  openTheater = () => this.setState({ theaterMode: true });
+
+  render = () => {
+    const { videoId } = this.props;
+    const theaterMode = get(this, 'state.theaterMode', false);
```
vincentfretin: I'd prefer you initialize `this.state = { theaterMode: false }` in constructor instead of doing that.

assembl/static2/js/app/components/common/youtubeTheater.jsx
```diff
@@ -0,0 +1,52 @@
...
+        frameBorder="0"
+      />
+    );
+    return (
+      <div className={`youtube-theater ${theaterMode ? 'open' : ''}`}>
+        {theaterMode ? (
+          <div className="theater-content">
+            <div className="youtube-video">
+              {video}
+              {theaterMode && <button onClick={this.closeTheater} className="close-theater-button assembl-icon-cancel" />}
```
vincentfretin: There is no need to check theaterMode again here.

PR #289, opened 2017-12-20
Display a message instead of landing page if timeline is empty

cedricmessiant:
Fixes DEVAS-871

### DIFF
--- a/assembl/static2/js/app/components/common/messagePage.jsx
+++ b/assembl/static2/js/app/components/common/messagePage.jsx
@@ -1,0 +1,21 @@
+// @flow
+import React from 'react';
+import { Grid, Jumbotron } from 'react-bootstrap';
+
+type Props = {
+  title: ?string,
+  text: string
+};
+
+const MessagePage = ({ title, text }: Props) => (
+  <Grid fluid>
+    <div className="max-container">
+      <Jumbotron>
+        {title && <h3>{title}</h3>}
+        <div>{text}</div>
+      </Jumbotron>
+    </div>
+  </Grid>
+);
+
+export default MessagePage;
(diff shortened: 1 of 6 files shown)


assembl/static2/js/app/components/common/messagePage.jsx
```diff
@@ -0,0 +1,21 @@
...
+type Props = {
+  title: ?string,
+  text: string
+};
+
+const MessagePage = ({ title, text }: Props) => (
+  <Grid fluid>
+    <div className="max-container">
+      <Jumbotron>
+        {title && <h3>{title}</h3>}
```
MilanBarande: Just an h3, no styleguide class ?

assembl/static2/js/app/utils/translations.js
```diff
@@ -560,7 +561,8 @@ const Translations = {
       contact: "Do you wish to know more about this platform?",
       contactUs: "Contact us",
       from_start_to_end: "from %{start} to %{end}",
-      chatbot: "Chat with %{chatbotName}"
+      chatbot: "Chat with %{chatbotName}",
+      assemblNotConfigured: "Assembl is not configured properly"
```
MilanBarande: I wouldn't say "not configured properly" but rather "has not been fully configured yet"

assembl/static2/js/app/components/common/messagePage.jsx
cedricmessiant: done, thanks

assembl/static2/js/app/utils/translations.js
cedricmessiant: done

PR #291, opened 2017-12-27
Antivirus

maparent:


assembl/models/attachment.py
```diff
@@ -172,6 +183,33 @@ def __init__(self, *args, **kwargs):
...
+        if extensions:
+            # somewhat random
+            return extensions[0]
+
+    def check_for_viruses(self, antivirus=None):
+        "Check if the file has viruses"
+        antivirus = antivirus or get_antivirus()
+        safe = antivirus.check(self.data, self.guess_extension())
+        status = AntiVirusStatus.passed.name if safe else AntiVirusStatus.failed.name
+        self.av_checked = status
```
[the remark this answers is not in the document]
ayyazdaniaryan: I don't think this function should be the end-all, be-all of virus checking needs. This should just make a check. Another 1-2 functions may be needed. One for setting the `av_checked` flag on an object. Another, to combine the two functions to be the `make_and_set_virus_check_status`, or something of that nature.

assembl/scripts/check_antivirus.py
```diff
@@ -0,0 +1,44 @@
...
+    if args.discussion:
+        (discussion_id,) = session.query(Discussion.id).filter_by(slug=args.discussion).first()
+        files = files.filter_by(discussion_id=discussion_id)
+    if args.reset:
+        with transaction.manager:
+            files.update({
+                File.av_checked: AntiVirusStatus.unchecked.name})
+            mark_changed(session)
+    with transaction.manager:
+        file_ids = files.with_entities(File.id).all()
```
ayyazdaniaryan: You could skip the `with_entities` step and just go straight to `files.all()`, and renaming `file_ids` to `file_objects`.

assembl/scripts/check_antivirus.py
```diff
@@ -0,0 +1,44 @@
...
+    for file_id in file_ids:
+        with transaction.manager:
+            file_obj = session.query(File).options(undefer(File.data)).get(file_id)
```
ayyazdaniaryan: Further simplification, based on comment above:
```
for file in file_objects:
   with transaction.manager:
      data = file.data  # fetches the lazy column
      file.check_for_viruses(antivirus)
```

assembl/models/attachment.py
```diff
@@ -172,6 +183,33 @@ def __init__(self, *args, **kwargs):
...
+        return status
+
+    def safe_data(self):
+        if self.av_checked == AntiVirusStatus.unchecked.name:
+            needs_check = self.discussion.preferences['requires_virus_check']
+            if needs_check:
+                self.av_checked = self.check_for_viruses()
+        if self.av_checked == AntiVirusStatus.failed.name:
+            return ''
```
ayyazdaniaryan: I think as a truly base implementation, this may suffice, but it can certainly improved upon. We could have an image placeholder for AV failed files, and instead of returning `''`, you can return a document object representing this placeholder. @paolina92 Can ask Roxana to make a quick image for AV failed if needed quickly.

assembl/lib/antivirus.py
```diff
@@ -0,0 +1,69 @@
...
+
+resolver = DottedNameResolver(__package__)
+
+
+class AntiVirus(object):
+    def check(self, data, suffix=None):
+        return NotImplementedError()
+
+
+class MockPositiveAntiVirus(AntiVirus):
```
ayyazdaniaryan: TODO: @mohamedzayed92 Should add these to the list of entities to be tested as there are no unit tests in this PR.

assembl/lib/antivirus.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+class MockRandomAntiVirus(AntiVirus):
+    "A Mock antivirus for testing. Reports 10%% files infected."
+    def check(self, data, suffix=None):
+        return random() > 0.1
+
+
+class FileAntiVirus(AntiVirus):
+    def check(self, data, suffix=None):
```
ayyazdaniaryan: Ugh, so much heavy lifting per file. This is going to crush CPU under moderate load. For first pass, though, it'll do.

assembl/views/api2/attachments.py
```diff
@@ -49,7 +49,7 @@ def get_file(request):
         .encode('iso-8859-1', 'replace'))
     url_quoted_utf8_filename = url_quote(f.title.encode('utf-8'))
     return Response(
-        body=f.data,
+        body=f.safe_data(),
```
ayyazdaniaryan: And if the resource is being called by multiple clients simultaneously when resource status is unchecked? There will be duplicate checks, and if antivirus machinery is unpredictable, the response of multiple checks could result in different status => headaches :(.

assembl/lib/antivirus.py
mohamedzayed92: no problem @ayyazdaniaryan ... As soon as it gets merged, i will create a fixture for MockPositiveAntivirus and unit tests

assembl/alembic/versions/c14c885ecdc2_document_av_checked.py
```diff
@@ -0,0 +1,35 @@
...
+
+from assembl.lib import config
+
+
+def upgrade(pyramid_env):
+    with context.begin_transaction():
+        avs = sa.Enum(
+            "unchecked", "passed" "failed", name="anti_virus_status")
+        # No create_type in alembic?
+        op.execute("CREATE TYPE anti_virus_status AS ENUM ('failed', 'passed', 'unchecked')")
```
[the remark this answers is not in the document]
maparent: The table creation implicitly declares types, which is what happened in the revision you pointed, but column addition does not. So I have to be explicit.

assembl/alembic/versions/c14c885ecdc2_document_av_checked.py
```diff
@@ -0,0 +1,35 @@
...
+import transaction
+
+
+from assembl.lib import config
+
+
+def upgrade(pyramid_env):
+    with context.begin_transaction():
+        avs = sa.Enum(
+            "unchecked", "passed" "failed", name="anti_virus_status")
```
swergas: Is there a comma missing after `"passed"`?

assembl/models/attachment.py
```diff
@@ -167,6 +175,10 @@ def __init__(self, *args, **kwargs):
 
     # Should we defer this?
     data = Column(LargeBinary, nullable=False)
+    # Note: use SQLA 1.1 for a better Enum behaviour
+    av_checked = Column(Enum(*AntiVirusStatus.__members__.keys(),
+                             name="anti_virus_status"),
+                        server_default='unchecked')
```
swergas: instead of `server_default='unchecked'`, you could probably use something like `server_default=AntiVirusStatus.unchecked` for safety

assembl/models/attachment.py
```diff
@@ -180,6 +183,29 @@ def __init__(self, *args, **kwargs):
...
+    def guess_extension(self):
+        extensions = guess_all_extensions(self.mime_type)
+        if extensions:
+            # somewhat random
+            return extensions[0]
+
+    def check_for_viruses(self, antivirus=None):
+        "Check if the file has viruses"
+        antivirus = antivirus or get_antivirus()
+        safe = antivirus.check(self.data, self.guess_extension())
```
swergas: What is the reason to do work about file extension? Is it really valuable input for the antivirus? (does it not already do this work?)

assembl/models/attachment.py
swergas: yes as well as this function probably needs to return immediately instead of wait until av check is done

assembl/scripts/check_antivirus.py
swergas: it may be intentional so that query does not get file contents (the total can be huge)

assembl/models/attachment.py
swergas: So I think we need to add a "check_in_progress" state or something, so that on first access to the file, status is "unchecked", and we add the check in a job and set status to "check_in_progress" and return a placeholder. (See how other web services like gmail do it)

assembl/views/api2/attachments.py
swergas: Related to this, see my comment in safe_data(): https://github.com/assembl/assembl/pull/291/commits/901c07adef4b31aaeb0ea7979ae5dd5b472b85af#r159834353

assembl/models/attachment.py
```diff
@@ -167,6 +177,36 @@ def __init__(self, *args, **kwargs):
...
+            # somewhat random
+            return extensions[0]
+
+    def ensure_virus_checked(self, antivirus=None):
+        "Check if the file has viruses"
+        antivirus = antivirus or get_antivirus()
+        # Lock row to avoid multiple antivirus processes
+        (status,) = self.db.query(File.av_checked).filter_by(id=self.id).with_for_update().first()
+        if status == AntiVirusStatus.failed.unchecked.name:
+            safe = antivirus.check(self.data, self.guess_extension())
```
swergas: What is the reason to do work about file extension before giving it to the antivirus? Is it really valuable input for the antivirus? (does it not already do this work?)

swergas:
ok for me, i added comments for improvements

PR #292, opened 2017-12-29
Vote session model for token vote admin branch (WIP)

n0izn0iz:


assembl/alembic/versions/7ea03fbce9a8_create_vote_session_table.py
```diff
@@ -0,0 +1,69 @@
...
+
+def LangStringId(column_name):
+    return Column(
+        column_name + '_id',
+        Integer(),
+        ForeignKey('langstring.id')
+    )
+
+
+def lang_strings_args(lang_strings_names):
```
swergas: bad function name

assembl/alembic/versions/7ea03fbce9a8_create_vote_session_table.py
```diff
@@ -0,0 +1,69 @@
...
+    args = [LangStringId(lang_string_name)
+        for lang_string_name in lang_strings_names]
+    
+    lang_strings_id_names = [name + "_id" for name in lang_strings_names]
+    
+    args.append(UniqueConstraint(*lang_strings_id_names))
```
swergas: why is a unique constraint needed for langstrings?

assembl/alembic/versions/7ea03fbce9a8_create_vote_session_table.py
```diff
@@ -0,0 +1,69 @@
...
+def upgrade(pyramid_env):
+    with context.begin_transaction():
+        op.create_table(
+            'vote_session',
+            IdColumn(),
+            ForeignIdColumn('discussion_phase',
+                nullable = False,
+                unique = True,
+            ),
+            *lang_strings_args([
```
swergas: lack of clarity about what is happening here, partly caused by bad function name

assembl/alembic/versions/7ea03fbce9a8_create_vote_session_table.py
n0izn0iz: Good question, it was declared like that in the example I followed.
I can't think of a situation where the same langstring object would be used for two different fields tho

assembl/alembic/versions/7ea03fbce9a8_create_vote_session_table.py
n0izn0iz: what would you recommend ? it's named like that because it's a function generating the langstrings arguments for the create_table function

PR #293, opened 2017-12-29
Docker dev env

n0izn0iz:


docker/dev.docker-compose.yml
```diff
@@ -0,0 +1,141 @@
...
+      - ..:/assembl
+    command: "sh -c '. venv/bin/activate && supervisord -n'"
+    depends_on:
+      - database
+      - redis
+      - memcached
+      - elasticsearch
+    ports:
+      - "6543:6543"
+      - "8080:8080"
```
vincentfretin: Be careful, you are exposing those ports to your network, this is equivalent to `"0.0.0.0:6543:6543"`. Docker add an iptables rule to open those ports on eth0. You probably want `"127.0.0.1:6543:6543"` here.

docker/dev.docker-compose.yml
```diff
@@ -0,0 +1,141 @@
...
+  dev:
+    image: assembl_supervisor
+    volumes:
+      - ..:/assembl
+    command: "sh -c '. venv/bin/activate && supervisorctl start dev:'"
+    depends_on:
+      - supervisord
+    restart: always
```
vincentfretin: You probably should add
```
supervisor__autostart_webpack = true
supervisor__autostart_pserve = true
supervisor__autostart_gulp = true
```
in your `docker_develop.rc` instead of starting a container to `start dev:`

docker/dev.docker-compose.yml
```diff
@@ -0,0 +1,141 @@
...
+      - ..:/assembl
+    hostname: cli-ssh
+    user: root
+    command: /bin/sh -c '/etc/init.d/ssh start && /usr/bin/ssh-keyscan localhost && while :; do sleep 1; done'
+  database:
+    image: postgres:9.5.5-alpine
+    volumes:
+      - ../pgdata:/var/lib/postgresql/data
+    expose:
+      - "5432"
```
vincentfretin: please use `127.0.0.1:5432:5432`

docker/dev.docker-compose.yml
n0izn0iz: I want access on lan, but I'll restrict to loopback by default

docker/dev.docker-compose.yml
n0izn0iz: pgadmin should not be on lan tho you're right

docker/dev.docker-compose.yml
n0izn0iz: will do

n0izn0iz:
I need to write some doc for this

paolina92:

PR #294, opened 2018-01-03
Flake8

cedricmessiant:
Use flake8 on all python files. Some modules are still excluded for now, we should fix them (see `.flake8` file). It also fixes some bugs that I discovered thanks to flake8
I ignored some rules, we can of course discuss the rules and params that we use.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -37,6 +37,6 @@
     files: ".*/static2/.*graphql$"
   - id: flake8
     name: flake8
-    entry: bash -c "PATH=./venv/bin/:$PATH flake8 `git diff --name-only --diff-filter=d --cached -- '*.py'|grep assembl/graphql`"
+    entry: bash -c "PATH=./venv/bin/:$PATH flake8 `git diff --name-only --diff-filter=d --cached -- '*.py'|grep assembl/`"
     language: system
-    files: "assembl/graphql/.*.py"
+    files: "assembl/"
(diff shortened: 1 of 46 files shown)


.pre-commit-config.yaml
```diff
@@ -37,6 +37,6 @@
     files: ".*/static2/.*graphql$"
   - id: flake8
     name: flake8
-    entry: bash -c "PATH=./venv/bin/:$PATH flake8 `git diff --name-only --diff-filter=d --cached -- '*.py'|grep assembl/graphql`"
+    entry: bash -c "PATH=./venv/bin/:$PATH flake8 `git diff --name-only --diff-filter=d --cached -- '*.py'|grep assembl/`"
     language: system
-    files: "assembl/graphql/.*.py"
+    files: "assembl/"
```
vincentfretin: This is not a valid expression for `files`. You should remove files here and add
use `types: [python]` instead.

vincentfretin:
Tests are failing with `[localhost] out: ImportError: cannot import name PrivateObjectMixin`

assembl/models/auth.py
```diff
@@ -42,6 +42,7 @@
     URLString, EmailString, EmailUnicode, CaseInsensitiveWord, CoerceUnicode)
 from ..lib.raven_client import capture_exception
 from . import Base, DiscussionBoundBase, PrivateObjectMixin
+from .generic import ContentSource
```
vincentfretin: Failing tests may be related to this.

assembl/models/auth.py
```diff
@@ -1510,7 +1507,6 @@ def get_discussion_id(self):
 
     @classmethod
     def get_discussion_conditions(cls, discussion_id, alias_maker=None):
-        from .generic import ContentSource
```
vincentfretin: Failing tests may be related to this.

.pre-commit-config.yaml
cedricmessiant: done

PR #295, opened 2018-01-03
Fix i18n in admin

cedricmessiant:
Fixes DEVAS-1083
I think that we should rename `selectedLocale` and `locale` variables. These names are confusing : 
- `selectedLocale` is the locale selected on the flags in the right column (it allows users to choose the language in which they edit the content)
- `locale` (`state.i18n.locale`) is the locale chosen in the navigation bar to specify the language of the interface

What do you think about it ? Any ideas for better names ?

### DIFF
--- a/assembl/static2/js/app/actions/adminActions/index.js
+++ b/assembl/static2/js/app/actions/adminActions/index.js
@@ -1,6 +1,6 @@
-export const updateSelectedLocale = newLocale => ({
+export const updateEditLocale = newLocale => ({
   newLocale: newLocale,
-  type: 'UPDATE_SELECTED_LOCALE'
+  type: 'UPDATE_EDIT_LOCALE'
 });
 
 export const updateThematics = thematics => ({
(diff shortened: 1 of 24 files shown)


cedricmessiant:
Name ideas : `dataLocale` and `interfaceLocale` (or stick with `locale` as it is used only in `administration.jsx`) ?

cedricmessiant:
@vincentfretin @paolina92 ?

vincentfretin:
locale stays, we use this name for state.i18n.locale in other part of the application. I think you meant to rename selectedLocale? Maybe editLocale?

cedricmessiant:
I'm fine with `editLocale` (it is in fact the lang param that we pass to graphql queries and mutations but `lang` is not a good name here)

cedricmessiant:
Done. There are still `selectedLocale` occurences in the code but it is for the post translation modal
I'm waiting for the travis builds and then I merge

PR #297, opened 2018-01-04
Unittest utils

mohamedzayed92:


assembl/models/action.py
```diff
@@ -293,8 +298,8 @@ def send_post_to_socket_ts(mapper, connection, target):
         & (_lpt.c.post_id == Content.__table__.c.id)
         & (_actt.c.type ==
            LikeSentimentOfPost.__mapper_args__['polymorphic_identity'])
-        & (_actt.c.tombstone_date == None)
-        ).correlate_except(_actt, _lpt), deferred=True)
+        & (_actt.c.tombstone_date is None)
```
vincentfretin: revert this change, in sqlalchemy `is None` is not supported

assembl/models/action.py
```diff
@@ -305,8 +310,8 @@ def send_post_to_socket_ts(mapper, connection, target):
         & (_dpt.c.post_id == Content.__table__.c.id)
         & (_actt2.c.type ==
            DisagreeSentimentOfPost.__mapper_args__['polymorphic_identity'])
-        & (_actt2.c.tombstone_date == None)
-        ).correlate_except(_actt2, _dpt), deferred=True)
+        & (_actt2.c.tombstone_date is None)
```
vincentfretin: revert this change
really don't modify this file, flake8 branch already fixes the errors.

assembl/models/auth.py
```diff
@@ -697,7 +697,8 @@ class User(AgentProfile):
     __tablename__ = "user"
 
     __mapper_args__ = {
-        'polymorphic_identity': 'user'
+        'polymorphic_identity':
+         'user'
```
vincentfretin: revert this change

assembl/tests/fixtures/ideas.py
```diff
@@ -95,9 +96,9 @@ def fin():
 
 
 @pytest.fixture(scope="function")
-def subidea_1(request, discussion, root_idea, test_session, ):
-    """An Idea fixture with a idealink to root idea fixture -
-    root_idea
+def subidea_1(request, discussion, root_idea, test_session):
+    """A subidea fixture with a idealink to root idea fixture -
+    top_idea
```
vincentfretin: Why did you change root_idea to top_idea here?

assembl/tests/model_tests/test_discussion.py
```diff
@@ -16,3 +16,107 @@ def test_add_discussion(test_session):
...
+
+def test_count_post_viewers2(
+        discussion, root_post_1, post_viewed, post_viewed2):
+    number_of_post_viewers = discussion.count_post_viewers()
+    assert number_of_post_viewers == 2
+
+
+def test_get_discussion_urls(discussion):
+    discussion_urls = discussion.get_discussion_urls()
+    assert 'http://localhost:6546/jacklayton2' == discussion_urls[0]
```
vincentfretin: You shouldn't hard code the domain here, it may change. You can use:
`assert discussion_urls[0].endswith('/jacklayton2')`

assembl/tests/fixtures/ideas.py
mohamedzayed92: While discussing with @ayyazdaniaryan what I understood is that there is just one root idea in a discussion, from which goes out top ideas and from each each top idea goes subideas. In the code comments the terminology is a bit different ... the root idea is actually treated as a top idea ... What do you think should I revert it ?

assembl/models/action.py
mohamedzayed92: it's weird ... I actually never did this change !!

mohamedzayed92: sorry my bad, I actually did "autopep8 --in-place --aggressive on that file ... this is the plugin I found to put everything on pep8

assembl/tests/fixtures/ideas.py
vincentfretin: root_idea is really root_idea here, it's discussion.root_idea. subidea_1 is actually your top idea, level 1. Just revert your docstring change here.

assembl/tests/model_tests/test_discussion.py
```diff
@@ -1,21 +1,22 @@
...
-    d.terms_and_conditions = LangString.create(
-        u"Use the haptic JSON system, then you can quantify the cross-platform capacitor!", "en")
-    d.legal_notice = LangString.create(
-        u"Use the optical SCSI microchip, then you can generate the cross-platform pixel!", "en")
-    assert d.topic == u"Education"
-    assert d.discussion_locales == ['en', 'fr', 'de']
-    assert d.terms_and_conditions.entries[0].value == u"Use the haptic JSON system, then you can quantify the cross-platform capacitor!"  # noqa: E501
-    assert d.legal_notice.entries[0].value == u"Use the optical SCSI microchip, then you can generate the cross-platform pixel!"  # noqa: E501
+def test_add_discussion(discussion):
+    # d = Discussion(
```
cedricmessiant: I don't like to test a content created by the fixture because the behavior you want to test here is the discussion creation. I think you could have just add the code that deletes the discussion at the end of this test to fix the test isolation problem. I guess it is not a big deal and it is just a personal opinion @vincentfretin ? @ayyazdaniaryan ?
(Also, if in the end you merge it like this, please remove the commented lines here)

assembl/tests/model_tests/test_discussion.py
```diff
@@ -119,4 +120,3 @@ def test_get_roles_by_permission(
     roles_by_permission = discussion_with_permissions.get_roles_by_permission()
     assert simple_permission.name in roles_by_permission
     assert simple_role.name in roles_by_permission[simple_permission.name]
-
```
cedricmessiant: please don't do this, it is not ok with this `.editorconfig` option : 
```insert_final_newline = true```

(use a editorconfig plugin for your editor to avoid this)

**Comment by cedricmessiant** [comment]
I just read the last commit but I'm OK to merge the tests and move on.
I guess you should rebase and push --force to remove all these merge commits ? Maybe @ayyazdaniaryan can help you with this ?

File: assembl/tests/model_tests/test_discussion.py (hunk 6 above)
Comment (vincentfretin) [reply assembl/tests/model_tests/test_discussion.py]: Indeed, you could have fixed it by removing all created objects I think. You should do a flush before the asserts otherwise you don't really test anything, you will miss the required fields errors for example.

File: assembl/tests/fixtures/auth.py [hunk 8]
```diff
@@ -47,4 +190,4 @@ def fin():
         test_session.delete(sap)
         test_session.flush()
     request.addfinalizer(fin)
-    return sap
+    return sap
```
vincentfretin: Like @cedricmessiant said, please respect .editorconfig configuration.

assembl/tests/fixtures/ideas.py
```diff
@@ -143,6 +144,27 @@ def fin():
...
+             discussion=discussion,
+             description=LangString.create("The definition of the subidea_1_2", 'en'))
+    test_session.add(i)
+    l_1_12 = IdeaLink(source=subidea_1, target=i)
+    test_session.add(l_1_12)
+    test_session.flush()
+
+    def fin():
+        print "finalizer subidea_1_1"
+        test_session.delete(l_1_11)
```
vincentfretin: Is this fixture used? l_1_11 is undefined, the print statement is wrong.

assembl/tests/lib_tests/test_utilities.py
```diff
@@ -0,0 +1,95 @@
...
+from assembl.tests.fixtures.base import test_webrequest
+#from assembl.models.discussion import read_post_ids
+
+
+# def test_get_localizer(test_dummy_web_request):
+# 	request = test_dummy_web_request
+# 	localizer = request.localizer
+# 	assert get_localizer(request) == localizer
+# 	assert get_localizer(None) == localizer 
+
```
vincentfretin: please remove commented code

assembl/tests/model_tests/test_discussion.py
```diff
@@ -16,3 +16,107 @@ def test_add_discussion(test_session):
     assert d.discussion_locales == ['en', 'fr', 'de']
     assert d.terms_and_conditions.entries[0].value == u"Use the haptic JSON system, then you can quantify the cross-platform capacitor!"  # noqa: E501
     assert d.legal_notice.entries[0].value == u"Use the optical SCSI microchip, then you can generate the cross-platform pixel!"  # noqa: E501
+    test_session.delete(d)
```
vincentfretin: You need test_session.flush() before the assert, otherwise you don't test the sql part, only the python part.

assembl/tests/model_tests/test_discussion.py
```diff
@@ -16,3 +16,107 @@ def test_add_discussion(test_session):
...
+
+
+def test_post_is_read(discussion, post_viewed, post_viewed2,
+                      participant1_user, participant2_user, root_post_1):
+    user_id = participant1_user.id
+    user_id2 = participant2_user.id
+    read_post_ids = discussion.read_post_ids(user_id)
+    read_post_ids2 = discussion.read_post_ids(user_id2)
+    ids = list(read_post_ids)
+    ids2 = list(read_post_ids2)
```
vincentfretin: please add a flush

vincentfretin:
Please do a rebase so we have a cleaner history once merged.

vincentfretin:
@mohamedzayed92 you just need to revert completly the changes done in assembl/models/action.py for the tests to pass.

PR #300, opened 2018-01-04
Prevent crash if a discussion phase has no .start or .end date

n0izn0iz:
I found it because if a phase has no date it crashes the notifications tasks

The new behavior is rough but it prevents the crash in this case

### DIFF
--- a/assembl/lib/frontend_urls.py
+++ b/assembl/lib/frontend_urls.py
@@ -59,10 +59,8 @@ def get_timeline_for_date(discussion, date):
     """
     if isinstance(date, basestring):
         date = dateutil.parser.parse(date)
-    mindate = datetime.date(datetime.MINYEAR, 1, 1)
-    mindate = datetime.datetime.combine(mindate, datetime.datetime.min.time())
-    maxdate = datetime.date(datetime.MAXYEAR, 1, 1)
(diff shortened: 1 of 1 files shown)


assembl/lib/frontend_urls.py
```diff
@@ -59,12 +59,15 @@ def get_timeline_for_date(discussion, date):
     """
     if isinstance(date, basestring):
         date = dateutil.parser.parse(date)
-
-    phases = sorted(discussion.timeline_phases, key=lambda p: p.start)
+    mindate = datetime.date(datetime.MINYEAR, 1, 1)
+    mindate = datetime.datetime.combine(mindate, datetime.datetime.min.time())
```
cedricmessiant: can't you just use `mindate = datetime.datetime(datetime.MINYEAR, 1, 1)` or `mindate = datetime.datetime(datetime.MINYEAR, 1, 1, min_time.hour, min_time.minute)` ?

vincentfretin: `mindate = datetime.datetime(datetime.MINYEAR, 1, 1)` is exactly what you want indeed. This gives you `datetime.datetime(1, 1, 1, 0, 0)`, exactly the same thing as the two lines.

cedricmessiant:
done / merged !

PR #302, opened 2018-01-04
can edit synthesis title from v1

vincentfretin:


assembl/static/js/app/views/messageColumnsPanel.js
```diff
@@ -346,10 +348,16 @@ var MessageColumnView = BaseMessageColumnView.extend({
     BaseMessageColumnView.prototype.onRender.apply(this, arguments);
     var that = this,
         header = this.model.get('header'),
+        synthesis_title = this.model.get('synthesis_title'),
         canEdit = Ctx.getCurrentUser().can(Permissions.ADMIN_DISCUSSION),
         renderId = _.clone(this._renderId);
 
     if (this.processIsEnded() || canEdit) {
+      this.messageColumnSynthesisTitle.show(new CKEditorLSField({
```
swergas: A synthesis title can contain just text and not rich text, so why do you use a rich text editor and then remove rich text in backend using `sanitize_text`, instead of just using a non-rich text field like `editableLSField`? I think for non-rich text inputs, frontend should not submit rich text to backend.

assembl/static/js/app/views/messageColumnsPanel.js
```diff
@@ -395,6 +403,11 @@ var MessageColumnView = BaseMessageColumnView.extend({
       }else{
         $('.js_messageColumnDescription').removeClass('message-column-description');
       }
+      if(!synthesis_title || synthesis_title.isEmptyStripped(translationData)){
+        $('.js_messageColumnSynthesisTitle').addClass('message-column-description');
+      }else{
+        $('.js_messageColumnSynthesisTitle').removeClass('message-column-description');
+      }
```
swergas: I'm not sure I understand the logic/goal behind this block.

assembl/static/js/app/views/messageColumnsPanel.js
vincentfretin: I tried to use editableLSField, but this is not the same behavior as the CKEditorLSField. When I click on the field, it was not the correct language I wanted to edit.
I don't well understand this stuff, so if you want to give it a try, please commit in the branch.

assembl/static/js/app/views/messageColumnsPanel.js
vincentfretin: It's just for style, it adds or removes a border and margin.

assembl/static/js/app/views/messageColumnsPanel.js
swergas: OK I'm starting to investigate. As is, it cannot work either because CKEditorLSField has no translationData parameter given.

PR #308, opened 2018-01-10
Vote session permissions

n0izn0iz:


assembl/graphql/schema.py
```diff
@@ -115,7 +115,6 @@ def resolve_root_idea(self, args, context, info):
         return root_thematic
 
     def resolve_vote_session(self, args, context, info):
-
         discussion_phase_id = args.get('discussion_phase_id')
         discussion_phase = models.DiscussionPhase.get(discussion_phase_id)
         # TODO: see if we can avoid this next(iter( thing with a one-to-one relationship
```
vincentfretin: About this TODO, you may simply want uselist=False on the relationship, no?

assembl/graphql/vote_session.py
```diff
@@ -78,7 +98,7 @@ def mutate(root, args, context, info):
             uploaded_file = context.POST[image].file
             uploaded_file.seek(0)
             data = uploaded_file.read()
-            discussion_id = context.matchdict['discussion_id']
+            discussion_id = context.matchdict("discussion_id")
```
vincentfretin: Are you sure about this change?

assembl/graphql/vote_session.py
```diff
@@ -56,16 +72,20 @@ def mutate(root, args, context, info):
...
 
         # TODO: see if we can avoid this next(iter( thing with a one-to-one relationship
         vote_session = next(iter(discussion_phase.vote_session or []), None)
         if vote_session is None:
-            vote_session = models.VoteSession(
-                discussion_phase=discussion_phase)
+            require_permission(CrudPermissions.CREATE)
+            vote_session = models.VoteSession(discussion_phase=discussion_phase)
+        else:
+            require_permission(CrudPermissions.UPDATE)
```
vincentfretin: This is wrong, CrudPermissions.UPDATE needs to be checked on the object you update, so with `obj.user_can` api. `cls.user_can_cls` is only used for `CrudPermissions.CREATE`

assembl/graphql/vote_session.py
```diff
@@ -12,6 +14,20 @@
     add_langstrings_input_attrs
 )
 from .document import Document
+from assembl.auth import CrudPermissions
+from assembl.auth.util import get_permissions
+
+
+def make_permissions_querier(cls, context):
```
vincentfretin: You can rename context to request here.

assembl/graphql/vote_session.py
n0izn0iz: woops, ty ^^

assembl/graphql/vote_session.py
n0izn0iz: I was wondering about that, thanks

n0izn0iz: After writing tests it seems that the class version is fine for this case.

assembl/graphql/schema.py
n0izn0iz: Perfect, thank you !

assembl/graphql/vote_session.py
vincentfretin: Yes because we don't really care here about owner (is_owner method), this is the difference between user_can_cls and user_can which check is_owner.

vincentfretin: But you really should support the right api now, this is important if another developer uses your require_permission api.

n0izn0iz: On it

vincentfretin:
ok for me

vincentfretin:
but you have a failing test, so not ok

assembl/tests/views_tests/test_graphql_vote_session.py
```diff
@@ -0,0 +1,254 @@
...
+        variable_values={
+            "discussionPhaseId": discussion_phase_id,
+            "titleEntries": [{"localeCode": "en", "value": new_title}]
+        }
+    )
+
+    assert ("wrong credentials" in response.errors[0].message)
+
+
+def test_graphql_update_vote_session_unauthenticated(graphql_unauthenticated_request, vote_session):
```
vincentfretin: For permissions tests, you should add discussion_with_data fixture to have the permissions associated to the discussion.

n0izn0iz:
Wow, tests are not failing locally but are on travis...

n0izn0iz:
Doing it now

n0izn0iz:
Still, why one travis build is passing and the other breaking?

vincentfretin:
It may depends on which tests you are executing and in which order. I did https://github.com/assembl/assembl/commit/7d74733eeb980edbba55c260240ec8f6083bfb68 to be sure the fixtures are properly isolated. Please rebase and use `discussion_with_data` fixture to have the permissions associated to the discussion for the tests where you test permissions.

vincentfretin:
I don't see where test_participant_webrequest fixture is defined. I did a search, I didn't find it.

vincentfretin:
You don't seem to have rebased. Travis: fixture 'discussion_with_default_data' not found

n0izn0iz:
Nice catch, it's named differently but this fixture is not used, I will remove it instead of fixing I think

n0izn0iz:
woops, I cherry-picked the commit of the rebased branch

n0izn0iz:
but as you see the pr test is passing which shows that the github merge is the same as the rebase

assembl/tests/fixtures/graphql.py
```diff
@@ -4,12 +4,29 @@
...
     req = test_adminuser_webrequest
     req.matchdict = {"discussion_id": discussion.id}
     req.method = 'POST'
     return req
 
 
+@pytest.fixture(scope="function")
+def graphql_unauthenticated_request(request, test_webrequest, discussion_with_default_data):
+    req = test_webrequest
+    req.matchdict = {"discussion_id": discussion_with_default_data.id}
```
cedricmessiant: you really need `discussion_with_default_data` here? from what I understood, it creates more stuff and thus takes more time. If it is just to test unauthenticated, `discussion` may suffice, isn't it?

vincentfretin: In this particular case yes, `discussion` suffice. `discussion_with_default_data` is needed in the case where you test a permission with a non admin account.

n0izn0iz: I think I do. It's hat fixed the failing tests

n0izn0iz: https://travis-ci.org/assembl/assembl/jobs/327716230

assembl/graphql/schema.py
```diff
@@ -115,11 +116,14 @@ def resolve_root_idea(self, args, context, info):
         return root_thematic
 
     def resolve_vote_session(self, args, context, info):
-
         discussion_phase_id = args.get('discussion_phase_id')
         discussion_phase = models.DiscussionPhase.get(discussion_phase_id)
-        # TODO: see if we can avoid this next(iter( thing with a one-to-one relationship
-        vote_session = next(iter(discussion_phase.vote_session or []), None)
+        require_discussion_phase_permission = make_instance_permissions_querier(discussion_phase, context)
```
cedricmessiant: you really need to check the permissions like this? I thought that the `SecureObjectType` class took care of it @ayyazdaniaryan @vincentfretin ?

n0izn0iz: I was wondering about that

cedricmessiant: (but maybe I miss something, I read the code quickly)

assembl/graphql/vote_session.py
```diff
@@ -56,16 +57,19 @@ def mutate(root, args, context, info):
...
-                "A vote session can only be created or edited with a '{}' discussion phase, check discussionPhaseId value".format(phase_slug))
+                "A vote session can only be created or edited with a '{}' discussion phase, check discussionPhaseId value".format(phase_identifier))
 
-        # TODO: see if we can avoid this next(iter( thing with a one-to-one relationship
-        vote_session = next(iter(discussion_phase.vote_session or []), None)
+        vote_session = discussion_phase.vote_session
         if vote_session is None:
-            vote_session = models.VoteSession(
-                discussion_phase=discussion_phase)
+            require_cls_permission = make_cls_permissions_querier(models.VoteSession, context)
```
cedricmessiant: maybe name it `check_vote_session_permission` ?

assembl/tests/fixtures/graphql.py
vincentfretin: You're right, the unauthencited user Everyone is not a special case, only the admin user having R_SYSADMIN is, see auth/util.py:get_permissions. For a user not having R_SYSADMIN, you need all DiscussionPermission created by `discussion_with_default_data`.

assembl/graphql/permissions_helpers.py
```diff
@@ -0,0 +1,27 @@
+from pyramid.security import Everyone
+from pyramid.httpexceptions import HTTPUnauthorized
+from assembl.auth.util import get_permissions
+
+
+def make_cls_permissions_querier(cls, request):
```
cedricmessiant: I agree that a helper is useful but maybe a simple `check_permission(cls, permission, request)` function would have been simpler here. You don't "gain" much since you are just accessing attributes/dict values in the outer function.

assembl/models/attachment.py
```diff
@@ -482,4 +482,8 @@ class VoteSessionAttachment(Attachment):
         'with_polymorphic': '*'
     }
 
-    # TODO: permissions !!!
+    crud_permissions = CrudPermissions(
+        P_ADMIN_DISC,   # create
```
cedricmessiant: I think that you can use named attributes instead of comments if you want it to be explicit (and I think it is a good idea to be explicit (even more with the 7 arguments that we have elsewhere in the code)

assembl/graphql/schema.py
vincentfretin: `SecureObjectType` only check the READ permission if you do a node query only.

assembl/models/vote_session.py
```diff
@@ -46,5 +48,11 @@ def get_discussion_id(self):
     def get_discussion_conditions(cls, discussion_id, alias_maker=None):
         return (cls.discussion_phase.discussion_id == discussion_id,)
 
+    crud_permissions = CrudPermissions(
+        P_ADMIN_DISC,   # create
```
cedricmessiant: (same here)

assembl/tests/fixtures/graphql.py
cedricmessiant: ok, good to know

assembl/models/attachment.py
n0izn0iz: agree, thanks

assembl/graphql/permissions_helpers.py
n0izn0iz: even more since the produced functions will probably only be called once, will change

assembl/tests/views_tests/test_graphql_vote_session.py
```diff
@@ -0,0 +1,254 @@
...
+    return components[0] + "".join(x.title() for x in components[1:])
+
+
+def assert_langstring_is_equal(langstring_name, fetched_model, source_model):
+    source_value = getattr(source_model, langstring_name).entries[0].value
+    fetched_value = fetched_model[snake_to_camel(langstring_name) + 'Entries'][0]['value']
+    assert source_value == fetched_value
+
+
+def assert_langstrings_are_equal(langstrings_names, fetched_model, source_model):
```
cedricmessiant: maybe you could move these 3 helpers to `tests/utils` ?

n0izn0iz: yes but I would need to modify them to be more generic instead of assuming there is only one entry

cedricmessiant: ok, so they can stay here in the meanwhile and if someone needs something similar he will refactor/move the helpers

assembl/graphql/schema.py
vincentfretin: For other queries, we check at least the READ permission on the discussion in the pyramid view.
In the current code afaict, we always use the 'read' permission for the `CrudPermissions.READ` permission type, and permissions-roles mapping is only on discussion, so checking READ permission type on both discussion and on an object is redundant. But you can keep the READ check here, to be safe.

assembl/graphql/vote_session.py
ayyazdaniaryan: Agreed, could be renamed to this instead, and maybe not even be a closure.

assembl/models/langstrings_helpers.py
```diff
@@ -43,9 +43,9 @@ def langstrings_attrs_dict(langstrings_names):
     return d
 
 
-def LangstringsBase(langstrings_names):
+def LangstringsBase(langstrings_names, model_name):
```
ayyazdaniaryan: As this is a method, can you please make it pythonic snake_case? `langstring_base`.

ayyazdaniaryan: The input here should have some sanitization, at min, a `string.trim`. Perhaps comment the `:param model_name` so that the format is clear for future usage.

assembl/models/vote_session.py
```diff
@@ -15,7 +16,7 @@
 
 
 class VoteSession(
-    LangstringsBase(langstrings_names),
+    LangstringsBase(langstrings_names, "VoteSession"),
```
ayyazdaniaryan: Better approach, `models.VoteSession.__name__`, keeps things DRY

assembl/tests/views_tests/test_graphql_vote_session.py
ayyazdaniaryan: The `snake_to_camel` is generic enough to be moved, tbf. But, keep these here for now. We'll make this more generic as more tests are being written.

assembl/tests/views_tests/test_graphql_vote_session.py
```diff
@@ -0,0 +1,254 @@
...
+def snake_to_camel(string):
+    # https://stackoverflow.com/questions/19053707/converting-snake-case-to-lower-camel-case-lowercamelcase
+    components = string.split('_')
+    # We capitalize the first letter of each component except the first one
+    # with the 'title' method and join them together.
+    return components[0] + "".join(x.title() for x in components[1:])
+
+
+def assert_langstring_is_equal(langstring_name, fetched_model, source_model):
+    source_value = getattr(source_model, langstring_name).entries[0].value
```
ayyazdaniaryan: Matching the `0`th value between `langstring` entries is extremely unsafe! You should make a check on the the specific locale, OR, check the `first_original` of the langstrings to check both versions.

ayyazdaniaryan: I understand in testing this might not be a real issue actually, so perhaps make a comment about this.

assembl/tests/views_tests/test_graphql_vote_session.py
```diff
@@ -0,0 +1,254 @@
...
+    fetched_value = fetched_model[snake_to_camel(langstring_name) + 'Entries'][0]['value']
```
ayyazdaniaryan: Please make a comment that the `fetched_model` is NOT an SQLA model, but rather a GraphQL model.

assembl/tests/views_tests/test_graphql_vote_session.py
```diff
@@ -0,0 +1,254 @@
...
+      }
+      propositionsSectionTitleEntries {
+        ...langStringEntry
+      }
+    }
+"""
+
+
+def en_value(entry):
+    return entry[0]['value']
```
ayyazdaniaryan: lolz, maybe only true for test case. Should probably check the locale for en. Again, it's testing, so the code will pass, but this is semantically not correct.

assembl/tests/views_tests/test_graphql_vote_session.py
```diff
@@ -0,0 +1,254 @@
...
+        vote_session)
+
+    fetched_image = fetched_vote_session['headerImage']
+    source_image = vote_session.attachments[0].document
+    assert fetched_image['title'] == source_image.title
+    assert fetched_image['mimeType'] == source_image.mime_type
+    assert fetched_image['externalUrl'] == source_image.external_url
+
+
+def en_entry(value):
```
ayyazdaniaryan: Move this to where the other helpers are, please :)

assembl/tests/views_tests/test_graphql_vote_session.py
```diff
@@ -0,0 +1,254 @@
...
+    assert source_value == fetched_value
+
+
+def assert_langstrings_are_equal(langstrings_names, fetched_model, source_model):
+    for langstring_name in langstrings_names:
+        assert_langstring_is_equal(langstring_name, fetched_model, source_model)
+
+
+fragments = u"""
+    fragment voteSessionGlobals on VoteSession {
```
ayyazdaniaryan: This isn't bad for a first implementation. Better approach, as you suggested yourself, is to use the fragments in the graphql folder itself. Read them as files, ensure they are in unicode, and assign them to the fragments variable. This *should* be a fixture

assembl/tests/views_tests/test_graphql_vote_session.py
```diff
@@ -0,0 +1,254 @@
...
+    assert en_value(fetched_vote_session['subTitleEntries']) == new_sub_title
+    assert en_value(fetched_vote_session['instructionsSectionTitleEntries']) == new_instructions_section_title
+    assert en_value(fetched_vote_session['instructionsSectionContentEntries']) == new_instructions_section_content
+    assert en_value(fetched_vote_session['propositionsSectionTitleEntries']) == new_propositions_section_title
+
+    fetched_image = fetched_vote_session['headerImage']
+    assert fetched_image['title'] == new_image_name
+    assert fetched_image['mimeType'] == new_image_mime_type
+    new_image_data = new_image.file.getvalue()
+    fetched_image_data = test_app.get(fetched_image['externalUrl']).body
```
ayyazdaniaryan: Ah cool, so you're committing a LocalStorage object in here. So, as it stands, this object will NOT be deleted from the db when the test is complete. This breaks test isolation principles. Let's make a cleanup function to remove these excess objects from the DB after the test (like the attachment)

assembl/tests/views_tests/test_graphql_vote_session.py
```diff
@@ -0,0 +1,254 @@
...
+    assert fetched_image_data == new_image_data
+
+
+def test_graphql_update_vote_session(graphql_request, vote_session, test_app):
+    mutate_and_assert(graphql_request, vote_session.discussion_phase_id, test_app)
+
+
+simple_get_query = u"""
```
ayyazdaniaryan: Maybe move this to the top of the file with other helpers?

assembl/tests/views_tests/test_graphql_vote_session.py
```diff
@@ -0,0 +1,254 @@
...
+    query VoteSesion($discussionPhaseId: Int!) {
+        voteSession(discussionPhaseId: $discussionPhaseId) {
+            id
+        }
+    }
+"""
+
+
+def vote_session_not_created(discussion_phase_id, graphql_request):
```
ayyazdaniaryan: same thing

assembl/tests/views_tests/test_graphql_vote_session.py
```diff
@@ -0,0 +1,254 @@
...
+    response = schema.execute(
+        simple_get_query,
+        context_value=graphql_request,
+        variable_values={"discussionPhaseId": discussion_phase_id}
+    )
+    return (response.errors is None) and (response.data['voteSession'] is None)
```
ayyazdaniaryan: Shouldn't an exception be thrown?

assembl/models/vote_session.py
n0izn0iz: Can't access VoteSession name in it's definition without doing dark magic :(
```
class foo(object):
    _name = traceback.extract_stack()[-1][2]
```

ayyazdaniaryan: Woah, that's surprising.

None of these work? 
https://stackoverflow.com/questions/510972/getting-the-class-name-of-an-instance-in-python

PR #309, opened 2018-01-10
Decrease time for running tests

vincentfretin:
We did some regressions in the past two months on the time we need to run all tests.
I fixed them.
On my machine `py.test assembl`
`256 passed, 3 skipped, 6 xfailed in 861.87 seconds so 14.33 minutes`
It was before 35 minutes.

On travis, we go from 38 to 24 or 28 minutes.

### DIFF
--- a/assembl/tests/utils/__init__.py
+++ b/assembl/tests/utils/__init__.py
@@ -122,15 +122,7 @@ def clear_rows(app_settings, session):
 def drop_tables(app_settings, session):
     log.info('Dropping all tables.')
     if not using_virtuoso():
-        # postgres. Thank you to
-        # http://stackoverflow.com/questions/5408156/how-to-drop-a-postgresql-database-if-there-are-active-connections-to-it
         session.close()
-        session.execute(
-            """SELECT pg_terminate_backend(pg_stat_activity.pid)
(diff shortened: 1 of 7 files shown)


assembl/tests/utils/__init__.py
```diff
@@ -122,15 +122,7 @@ def clear_rows(app_settings, session):
 def drop_tables(app_settings, session):
     log.info('Dropping all tables.')
     if not using_virtuoso():
-        # postgres. Thank you to
-        # http://stackoverflow.com/questions/5408156/how-to-drop-a-postgresql-database-if-there-are-active-connections-to-it
         session.close()
-        session.execute(
-            """SELECT pg_terminate_backend(pg_stat_activity.pid)
```
swergas: this needed review before merge

vincentfretin: The drop_tables function here is only used in tests, so I think it's safe to remove this. I'm not sure why it was historically needed, maybe to fix an issue between jenkins jobs because it was stateful? In travis, it's stateless, the env is recreated each time. This code only works if the user running the tests is the assembl_test user I guess. It gave sometime an error on travis and on our local machine.

PR #310, opened 2018-01-12
Avoid downgrading setuptools

vincentfretin:


fabfile.py
```diff
@@ -556,7 +554,6 @@ def build_virtualenv():
                     venv_config.set(sec, option, val)
                 with open(vefile, 'w') as f:
                     venv_config.write(f)
-    run('rm /tmp/distribute* || echo "ok"')  # clean after myself
```
ayyazdaniaryan: With the `--no-setuptools`, does python not create mess in the `/tmp` folder?

requirements.txt
```diff
@@ -29,7 +29,6 @@ coverage==4.1
 cryptography==2.0.3
 DateTime==4.1.1
 defusedxml==0.4.1
-# distribute==0.7.3 # This fails with latest pip. Installed properly as a dependency.
```
ayyazdaniaryan: deprecated, yes? 
https://pypi.python.org/pypi/distribute/0.6.49

setup.py
```diff
@@ -1,14 +1,13 @@
 import os
 
+from pip.download import PipSession
```
ayyazdaniaryan: Any reason for the move?

fabfile.py
vincentfretin: distribute was the fork of setuptools several years ago, that was since merged back into setuptools. We don't use distrubute anymore.

setup.py
vincentfretin: Yes, all imports needs to be before variables.

.pre-commit-config.yaml
```diff
@@ -37,6 +37,6 @@
     files: ".*/static2/.*graphql$"
   - id: flake8
     name: flake8
-    entry: bash -c "PATH=./venv/bin/:$PATH flake8 `git diff --name-only --diff-filter=d --cached -- '*.py'|grep assembl/`"
+    entry: bash -c "PATH=./venv/bin/:$PATH flake8 `git diff --name-only --diff-filter=d --cached -- '*.py'`"
```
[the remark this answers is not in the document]
vincentfretin: The grep here was previously "grep assembl/graphql/" and has been transformed to just "grep assembl/" when we did the flake8 branch with the exclude files. So it is not necessary anymore, and the issue here was that because of the grep, modifing setup.py or fabfile.py was executing just `fake8` in the pre-commit, without any args, so it tried to check the working directory including venv and taking 30 minutes or more to execute.

fabfile.py
vincentfretin: With or without the --no-setuptools, it doesn't create this /tmp/distribute files. This was some old code when we used distribute instead of setuptools on this project.

requirements.txt
vincentfretin: yes, like I explained in the previous comment

.flake8
```diff
@@ -1,4 +1,4 @@
 [flake8]
 max-line-length=160
 ignore = E501,F403,E722,E123,E124
-exclude = assembl/static/*,assembl/static2/*,assembl/alembic/*,assembl/indexing/*,assembl/lib/*,assembl/locale/*,assembl/nlp/*,assembl/scripts/*,assembl/semantic/*,assembl/static/*,assembl/static2/*,assembl/tasks/*,assembl/templates/*,assembl/tests/*,assembl/tweens/*,assembl/view_def/*,assembl/views/*
\ No newline at end of file
+exclude = assembl/static/*,assembl/static2/*,assembl/alembic/*,assembl/indexing/*,assembl/lib/*,assembl/locale/*,assembl/nlp/*,assembl/scripts/*,assembl/semantic/*,assembl/static/*,assembl/static2/*,assembl/tasks/*,assembl/templates/*,assembl/tests/*,assembl/tweens/*,assembl/view_def/*,assembl/views/*
```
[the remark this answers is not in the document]
vincentfretin: Good catch, I'll removethe duplicate entries.
flake8 is operating on all python files that are not excluded by this list, so currently setup.py, fabfile.py, models, graphql

PR #311, opened 2018-01-16
fix burger language menu on small height screens (DEVAS-1075)

n0izn0iz:
https://bluenove.atlassian.net/browse/DEVAS-1075


vincentfretin:
If this fixes the issue, I guess it's ok.

assembl/static2/js/app/components/navbar/languageMenu.jsx
```diff
@@ -73,7 +75,7 @@ class LanguageMenu extends React.Component {
   };
 
   render() {
-    const { size, i18n, style, className, setWidth } = this.props;
+    const { size, i18n, style, className, setWidth = doNothing } = this.props;
```
vincentfretin: Did you have an error? Why do you need a default value now?

n0izn0iz: yes. of the two uses of it, one doesn't provide this callback. I think it worked before but it was spewing red in the console.

PR #312, opened 2018-01-16
Modify password

vincentfretin:


assembl/static2/js/app/components/common/modifyPasswordForm.jsx
```diff
@@ -0,0 +1,122 @@
...
+    disabled: false,
+    oldPassword: '',
+    newPassword: '',
+    newPassword2: ''
+  };
+
+  handleOldPasswordChange = (e: SyntheticEvent) => {
+    if (e.target instanceof HTMLInputElement) {
+      const value = e.target.value;
+      const disabled = value.length === 0 || this.state.newPassword.length === 0 || this.state.newPassword2.length === 0;
```
cedricmessiant: maybe use a function or method for this to avoid repetition / for maintainability

vincentfretin: This is different each time here.

vincentfretin: or you mean to create a function like isEmpty?

assembl/graphql/view.py
```diff
@@ -13,9 +13,15 @@
 class LoggingMiddleware(object):
     def resolve(self, next, source, gargs, context, info, *args, **kwargs):
         if source is None:
+            variables = {}
+            for key, value in info.variable_values.items():
+                if 'password' in key or 'Password' in key:
+                    variables[key] = 'xxxxxxxxxxxxx'
```
ayyazdaniaryan: Is it safe to override the info key in the middleware? Would it be possible for logging middleware to affect the info fields downstream for the mutation?

assembl/models/auth.py
```diff
@@ -759,6 +759,15 @@ def password_p(self):
     def password_p(self, password):
         from ..auth.password import hash_password
         if password:
+            # keep the current password to avoid the user to reuse it later
+            self.old_passwords.append(
```
ayyazdaniaryan: Do you not make any checks before appending?

assembl/graphql/user.py
```diff
@@ -96,12 +99,34 @@ def mutate(root, args, context, info):
...
+            if name is not None:
+                user.real_name_p = name
+
+            old_password = args.get('old_password')
+            new_password = args.get('new_password')
+            new_password2 = args.get('new_password2')
+            # only modify the password if it was given in parameter
+            if old_password is not None and new_password is not None and new_password2 is not None:
+                if not user.check_password(old_password):
+                    raise Exception(u"002: You entered a wrong password.")
```
ayyazdaniaryan: What pattern are you using for the `00N` errors? I'm guessing you use these in the front-end to render the appropriately translated response for errors? If so, you can opt to translate in the backend using the `TranslationFactory` and pass the message along. Furthermore, just during the transition period of moving to graphql, @maparent and I created a sort of "contract" for errors in API format for the V2 interface. I don't know if you wish to expand on that contract or utilize this instead.

The contract is described best in `views/__init__.py` as `JSONError`. You can see multiple uses of the JSONError in `assembl/views/api2/auth.py`. And the consumer of these errors are often in the actions/services in the V2 front-end. This way the front-end just passes the error along to the banner.

assembl/alembic/versions/27179de32822_add_old_password_table.py
```diff
@@ -0,0 +1,46 @@
...
+                    'user.id',
+                    ondelete="CASCADE",
+                    onupdate="CASCADE"
+                ),
+                nullable=False,
+                index=True
+            ),
+            sa.Column(
+                'password',
+                sa.Binary(115),
```
ayyazdaniaryan: Hmm, I'm not so sure, something tells me this is wrong, old passwords are still sensitive data, even if outdated, and should be encrypted. Perhaps use the Psql column encryption on this one? I understand this will require key management on the server, more orchestration for fabric/docker tooling. Thoughts?

assembl/models/auth.py
```diff
@@ -759,6 +759,15 @@ def password_p(self):
...
+                OldPassword(password=self.password))
```
ayyazdaniaryan: Putting unsalted, unencrypted password in the db here. Old, but still sensitive data :(

assembl/static2/js/app/utils/translations.js
```diff
@@ -285,11 +285,24 @@ const Translations = {
...
       save: "Enregistrer",
       password: "Mot de passe",
-      changePassword: "Modifier",
-      saveSuccess: "Votre profil a été mis à jour avec succès"
+      changePassword: "Modifier mon mot de passe",
+      passwordModifiedSuccess: "Votre mot de passe a été modifié avec succès",
+      saveSuccess: "Votre profil a été mis à jour avec succès",
+      updateUser: {
+        errorMessage: {
+          1: "Nous avons déjà un utilisateur ayant ce nom d'utilisateur.",
```
ayyazdaniaryan: Yeah, so these could be done via the backend already, reducing the json file size and computation on front-end.

assembl/alembic/versions/27179de32822_add_old_password_table.py
vincentfretin: I used the same type as the user.password field. The password is hashed & salted.

assembl/graphql/user.py
vincentfretin: The pattern is "00N: some text".
We currently still have the issue that the backend translation are using the language selected in the v1 interface (cookie) that can be different from the selected lang in the v2 interface (the two issues we have since several sprints). So you can be in French in the v2 interface but having the errors in English. By doing the translations on the frontend, I don't have the issue. And I think it's best to have the translations of the fields and errors messages in the same po file, so we are sure the translators update all new messages.

assembl/models/auth.py
vincentfretin: Which check do you want me to do? self.old_passwords is a relation, so we know it's a list.

assembl/models/auth.py
vincentfretin: `self.password` is the hash & salted password here.

assembl/static2/js/app/utils/translations.js
vincentfretin: Well, for 4 more messages, it doesn't make a lot of difference. If you really want to do something about the bundle size, we probably should load a json file in async when you change the language interface.

assembl/graphql/view.py
vincentfretin: I don't override anything, it's a new variables dict here, only used for the log.

vincentfretin:
This was rebased and merged.

assembl/alembic/versions/27179de32822_add_old_password_table.py
ayyazdaniaryan: Okay, good.

assembl/models/auth.py
ayyazdaniaryan: Ahh of course. Okay, all good.

assembl/static2/js/app/utils/translations.js
ayyazdaniaryan: Yah, it's 4 messages now. But now this is the start of a pattern for handling errors. Yourself and other developers would pick this up and make it the norm of exception exchange between boundaries.

vincentfretin: I'm not against revisiting this. But we need to fix the selected language mismatch between frontend and backend first.

ayyazdaniaryan: That will likely be resolved with https://bluenove.atlassian.net/browse/DEVAS-715.

vincentfretin: Yes, this is this issue.

assembl/models/auth.py
swergas: I understood the same thing by reading the commit code outside of PR, I did not see that there were 2 different variables password and self.password. I think you should rename these 2 variables to make them really look different and more clear. This is a readability problem.

PR #313, opened 2018-01-16
invert vertical order of idea card text (DEVAS-1128)

n0izn0iz:
https://bluenove.atlassian.net/browse/DEVAS-1128


assembl/static2/js/app/components/debate/common/ideaPreview.jsx
```diff
@@ -48,10 +42,16 @@ const IdeaPreview = ({
             >
               <Translate value="debate.thread.seeSubIdeas" count={numChildren} />
             </div>
+            <div>/</div>
```
cedricmessiant: have you tried to remove this ?

n0izn0iz: it's the visual separator between the two lines

PR #317, opened 2018-01-18
Improve V1 vote UI

swergas:


swergas:
This is ready for merge, I will have to update i18n files after merge to add French translation of Guideline (Consigne)

assembl/lib/locale.py
```diff
@@ -133,8 +133,8 @@ def get_preferred_languages(session, user_id):
 def locale_negotiator(request):
     settings = get_config()
     available = settings.get('available_languages').split()
-    locale = (request.cookies.get('_LOCALE_', None) or
-              request.params.get('_LOCALE_', None))
+    locale = (request.params.get('_LOCALE_', None) or
```
ayyazdaniaryan: You want to override a cookie param with a specific query string when bringing users to the V1 interface?

assembl/static/js/app/routeManager.js
```diff
@@ -386,9 +386,13 @@ var routeManager = Marionette.Object.extend({
     });
   },
 
-  voteWidgetFromV2: function(id, arg) {
+  voteWidgetFromV2: function(id, arg, queryString, showInModal) {
     var that = this;
-    var collectionManager = CollectionManager(); 
+    showInModal = (showInModal === true) ? true : false;
```
ayyazdaniaryan: Could just do:
```
showInModal ? showInModal : false
```

assembl/static/js/app/routeManager.js
```diff
@@ -386,9 +386,13 @@ var routeManager = Marionette.Object.extend({
...
+    if (id){
+      id = parseInt(id.match(/\d+/)[0], 10); // extract number from id, in case provided id is "local:Widget/23" instead of "23"
```
ayyazdaniaryan: Not faster to `.split('/')[1].trim()` it instead? Meh, minor optimization.

assembl/static/js/app/routeManager.js
```diff
@@ -408,19 +412,47 @@ var routeManager = Marionette.Object.extend({
           });
       });
     widgetPromise.then(function(widget){
-      var CurrentWidgetView;
+      if (!widget){
+        var error_text = "Requested vote widget could not be found.";
+        alert(error_text);
```
ayyazdaniaryan: Why `alert` in a prod system? Should instead show the error `growl` instead and not show the modal/view instead.

ayyazdaniaryan: `error_text` should also be i18n'ed, and turned into `camelCase`

assembl/static/js/app/routeManager.js
```diff
@@ -408,19 +412,47 @@ var routeManager = Marionette.Object.extend({
...
+
+      if ( showInModal ){
+        Ctx.setCurrentModalView(instanciatedView);
+        Assembl.slider.show(instanciatedView);
       }
       else {
-        CurrentWidgetView = Views.TokenVoteSessionModal
+        Assembl.groupContainer.show(instanciatedView);
+        // some CSS hacks to clear default page
+        $("body").css("background", "white");
```
ayyazdaniaryan: Why not make a css class with these values, eg:
```
.token-vote-session-modal {
  body {
    background: $light-background
  }
  #header {
    // Remove it so and make sure you don't cause positional issues on downstream elements
    display: none,
    overflow: visible
  }
}
```

And instead of hard-coding the css here, add this class to the highest element possible in the tree. And You should remove this css class on closing of the view.

assembl/static/js/app/views/tokenVoteSession.js
```diff
@@ -252,8 +252,15 @@ var RemainingCategoryTokensView = Marionette.ItemView.extend({
...
     var categoryNameLangString = category.get("name");
-    var categoryNameBestTranslation = categoryNameLangString.bestWithErrors(that.userLanguagePreferences, false).entry.value();
-    // var s = i18n.sprintf(i18n.ngettext("<span class='available-tokens-number'>%d</span> remaining %s token", "<span class='available-tokens-number'>%d</span> remaining %s tokens", data["remaining_tokens"]), data["remaining_tokens"], category.get("typename")); // TODO: use "name" field instead (LangString)
+    var categoryNameBestTranslation = "";
+    try {
+      categoryNameBestTranslation = categoryNameLangString.bestWithErrors(that.userLanguagePreferences, false).entry.value();
+    }
+    catch (e) {
+      console.error("error managing langstring categoryNameLangString: ", e, "langstring was:", categoryNameLangString);
+      // TODO: understand and fix the root cause of the problem (happens only when displayed in a modal)
```
ayyazdaniaryan: What is the error here? Why would you have to resort to this?

assembl/static/js/app/views/tokenVoteSession.js
```diff
@@ -1708,19 +1719,107 @@ var TokenVoteSessionModal = Backbone.Modal.extend({
...
+
+  onDestroy: function(){
+    if ( this.model ){
+      /*
+      We re-fetch widget data from the server, so that when the user opens the vote popin again,
+      up-to-date information is displayed (because this widget data comes from a model contained
+      in the collection collectionManager::getAllWidgetsPromise()).
+      TODO: This is a bit hacky, maybe we should find a better way, for example when popin opens, re-load user votes (instead of widget) and use these instead of the ones listed in the widget.
+      */
+      this.model.fetch();
```
ayyazdaniaryan: Yeah, I don't think this is the best place to do it. You should re-fetch the model from the `collectionManager` at a view initialization. If all votes have to be updated again, you can force a re-fetch on every widget modal init view. This is sprinkling more code, but more robust that this, as view removal is locked.

assembl/static/js/app/views/tokenVoteSession.js
```diff
@@ -1708,19 +1719,107 @@ var TokenVoteSessionModal = Backbone.Modal.extend({
...
+var TokenVoteSessionSubmittedView = Marionette.LayoutView.extend({
+  constructor: function TokenVoteSessionSubmittedView(){
+    Marionette.LayoutView.apply(this, arguments);
+  },
+
+  template: false,
+  className: 'token-vote-session-submitted',
+
+  initialize(options){
+    this.editInModal = ("editInModal" in options && options.editInModal == true) ? true : false;
```
ayyazdaniaryan: Easier to do this instead?
```'editInModal' in options ? options.editInModal```

File: assembl/static/js/app/views/tokenVoteSession.js [hunk 9]
```diff
@@ -1708,19 +1719,107 @@ var TokenVoteSessionModal = Backbone.Modal.extend({
...
+    console.log("TokenVoteSessionSubmittedView::initialize() editInModal:", this.editInModal);
+  },
+
+  onShow: function(){
+    var that = this;
+    console.log("TokenVoteSessionSubmittedView::onShow() editInModal:", that.editInModal);
```
ayyazdaniaryan: Should put all these `console.log's in a wrapped `Ctx.debugRender`, if you need them at all.

assembl/lib/locale.py
swergas: yes

assembl/static/js/app/routeManager.js
swergas: this CSS would ont work, as body element is not into a .token-vote-session-modal element. Remember it is use-and-throw-away code that will very soon be completely replaced by v2 UI

assembl/static/js/app/views/tokenVoteSession.js
swergas: I could not find the origin of the issue and it was very difficult to reproduce (rare).

assembl/static/js/app/routeManager.js
swergas: or !!showInModal but I prefer explicit for readability

assembl/static/js/app/views/tokenVoteSession.js
swergas: This code was there before and refactoring it was not part of this task (only visual improvements), and applies only to the modal view version of it, not the full page view that is the one that will really be used.

assembl/static/js/app/views/tokenVoteSession.js
swergas: my version ensures variables contains a boolean in the end

PR #319, opened 2018-01-18
add phase2_csv_export view (wip)

vincentfretin:
I did the view to export the phase2 posts http://localhost:6543/data/Discussion/{debateId}/phase2_csv_export
similar to what we have for phase1 csv export.

We should add a test, and the button in the v2 admin


vincentfretin:
Ok, let's merge this and add the frontend in the next sprint.

PR #321, opened 2018-01-24
Display the languages menu in the admin depending on the section

paolina92:


assembl/static2/js/app/reducers/adminReducer/index.js
```diff
@@ -198,6 +198,17 @@ export const discussionLanguagePreferencesHasChanged: DiscussionLanguagePreferen
   }
 };
 
+type ToggleLanguageMenu = boolean;
+type ToggleLanguageMenuReducer = (ToggleLanguageMenu, ReduxAction<Action>) => ToggleLanguageMenu;
+export const toggleLanguageMenu: ToggleLanguageMenuReducer = (state = false, action) => {
+  switch (action.type) {
+  case 'TOGGLE_LANGUAGE_MENU':
+    return action.state;
```
cedricmessiant: if it is a toggle action, I think you should handle the fact that it toggles the value here (so no value in the action and `return !state`). If you need to be able to choose the value, I would rename the action `UPDATE_LANGUAGE_MENU_VISIBILITY` or something)

assembl/static2/js/app/reducers/adminReducer/index.js
```diff
@@ -198,6 +198,17 @@ export const discussionLanguagePreferencesHasChanged: DiscussionLanguagePreferen
   }
 };
 
+type ToggleLanguageMenu = boolean;
+type ToggleLanguageMenuReducer = (ToggleLanguageMenu, ReduxAction<Action>) => ToggleLanguageMenu;
+export const toggleLanguageMenu: ToggleLanguageMenuReducer = (state = false, action) => {
```
cedricmessiant: I think it is really a strange name for a reducer, maybe use something like `languageMenuIsVisible` or `displayLanguageMenu`

paolina92: I agree. Let's do that...

assembl/static2/js/app/reducers/adminReducer/index.js
paolina92: ok

PR #322, opened 2018-01-24
Use mainline social_auth. Create a generic backend (and derived generic oauth2) that takes all info from settings. 

maparent:


assembl/auth/generic_auth_backend.py
```diff
@@ -0,0 +1,152 @@
...
+
+
+class GenericOAuth2(GenericAuth, BaseOAuth2):
+    """A Generic OAuth2 client"""
+
+    # properties and settings from BaseAuth
+    # EXTRA_DATA: None
+    # REQUIRES_EMAIL_VALIDATION: False
+    # SEND_USER_AGENT: False
+    # SSL_PROTOCOL: None
```
ayyazdaniaryan: Safe to send server-to-server request without SSL ?

assembl/auth/generic_auth_backend.py
```diff
@@ -0,0 +1,152 @@
...
+        def getResponseInfo(prop_name):
+            return self.getResponseInfo(prop_name, response)
+
+        name_info = {}
+        for partName in ('first_name', 'last_name', 'fullname'):
+            info = getResponseInfo(partName)
+            if info is not None:
+                name_info[partName] = info
+        fullname, first_name, last_name = self.get_user_names(**name_info)
+        return {'username': getResponseInfo('username') or first_name or '',
```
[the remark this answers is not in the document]
maparent: It's at login time, it's done once afaik... I would not worry about it too much.

assembl/auth/generic_auth_backend.py
maparent: Most of the time it will be SSL. Those are just the defaults extracted from  the social_core code.

maparent:
This was merged. Closing.

PR #323, opened 2018-01-25
coherent locale setting between v1 and v2

swergas:
(using cookies instead of localStorage)

### DIFF
--- a/assembl/static2/js/app/utils/globalFunctions.js
+++ b/assembl/static2/js/app/utils/globalFunctions.js
@@ -182,8 +182,7 @@ export const getCookieItem = (sKey: string) => {
   return (
     decodeURIComponent(
       document.cookie.replace(
-        // eslint-disable-next-line no-useless-escape
-        new RegExp(`(?:(?:^|.*;)\\s*${encodeURIComponent(sKey).replace(/[\-\.\+\*]/g, '\\$&')}\\s*\\=\\s*([^;]*).*$)|^.*$`),
+        new RegExp(`(?:(?:^|.*;)\\s*${encodeURIComponent(sKey).replace(/[-.+*]/g, '\\$&')}\\s*\\=\\s*([^;]*).*$)|^.*$`),
         '$1'
       )
     ) || null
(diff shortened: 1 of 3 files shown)


assembl/static2/js/app/utils/globalFunctions.js
```diff
@@ -173,4 +173,19 @@ export const isMobile = {
...
+export const youtubeRegexp = /(?:youtube\.com\/(?:[^/]+\/.+\/|(?:v|e(?:mbed)?)\/|.*[?&]v=)|youtu\.be\/)([^"&?/ ]{11})/i;
+
+export const getCookieItem = (sKey: string) => {
+  if (!sKey) {
+    return null;
+  }
+  return (
+    decodeURIComponent(
+      document.cookie.replace(
+        // eslint-disable-next-line no-useless-escape
```
cedricmessiant: are there useful escapes that eslint consider as useless ?!

swergas: I did not take the time to try removing them and test if the function still works in all main cases... should I?

vincentfretin: You should. You normally don't have to espace with '\' all characters in brackets.

assembl/static2/js/app/utils/globalFunctions.js
```diff
@@ -173,4 +173,19 @@ export const isMobile = {
...
+        new RegExp(`(?:(?:^|.*;)\\s*${encodeURIComponent(sKey).replace(/[\-\.\+\*]/g, '\\$&')}\\s*\\=\\s*([^;]*).*$)|^.*$`),
```
ayyazdaniaryan: Currently our cookies are pretty small. So, creating this `RegExp` obj each time this method has limited cost, especially since it's being called on the critical path of the front-end. I would be curious to know what the cost/benefit would be of caching this result, perhaps using:
https://www.npmjs.com/package/regex-cache

vincentfretin: This is micro benchmarking here. This won't change anything in time to render.

ayyazdaniaryan: I don't expect much change for the moment. But, this would be a place for optimization if/when our cookies grow.

PR #325, opened 2018-01-25
Token view step1

MilanBarande:


assembl/static2/js/app/graphql/VoteSession.graphql
```diff
@@ -1,8 +1,16 @@
 #import "./fragments/VoteSession.graphql"
 
-query VoteSession($discussionPhaseId: Int!) {
+query VoteSession($discussionPhaseId: Int!, $lang: String!) {
   voteSession(discussionPhaseId: $discussionPhaseId) {
     ...voteSessionGlobals
     ...voteSessionLangstringsEntries
+    headerImage {
+      externalUrl
+    }
```
vincentfretin: `  ...voteSessionGlobals` already retrieve 
`headerImage {
    title
    mimeType
    externalUrl
  }
}`
what is the behaviour when you override a field like this? Does it still load mimeType and title required in the administration?

MilanBarande: My bad, I didn't notice. I just logged the received data as things stand, you still get mimeType and title. I will remove this useless field on develop

PR #326, opened 2018-01-26
Don't display password form if user is connected via social auth

cedricmessiant:


vincentfretin:
@swergas or @ayyazdaniaryan can we have a second review please?

PR #328, opened 2018-01-31
fix synthesis notification email mixed languages

swergas:


assembl/templates/notifications/html_mail_post_synthesis.jinja2
```diff
@@ -9,12 +9,14 @@ notification
...
+lang_prefs
 #}
 
 {% extends "notifications/html_mail_post.jinja2" %}
 
 {%- block post_body %}
-  {{ synthesis.as_html(jinja_env) }}
+  {{ synthesis.as_html(jinja_env, lang_prefs) }}
   {#
   {{ synthesis.introduction }}
```
vincentfretin: You should fix this one `synthesis.introduction` and `synthesis.conclusion` to avoid a warning/error in celery, even if it's just a comment, I guess replace it with `synthesis.introduction.best_lang(lang_prefs).value|safe` as well?

swergas: I did not see it produced any warning or error, and does it really as it is commented code? And this was out of scope of this fix but OK I remove this legacy commented code.

assembl/tests/model_tests/test_notifications.py
```diff
@@ -98,6 +98,64 @@ def test_notification_follow_synthesis(test_session, discussion,
     assert notification_count == initial_notification_count + 1, "The synthesis post should have matched and created a notification"
 
 
+def test_synthesis_notification_in_user_preferred_language_fr(test_session, discussion,
+                                                           participant1_user, participant1_user_language_preference_fr_cookie,
```
ayyazdaniaryan: PEP8 here.

assembl/tests/model_tests/test_notifications.py
```diff
@@ -98,6 +98,64 @@ def test_notification_follow_synthesis(test_session, discussion,
...
+    dispatcher.processPostCreated(synthesis_post_1.id)
+
+    last_notification = participant1_user.notifications[-1]
+
+    html_content = last_notification.render_to_email_html_part()
+
+    assert "subject FR" in html_content
+    assert "introduction FR" in html_content
+    assert "conclusion FR" in html_content
+    # TODO: also check that syntheses of several ideas do also show in the user's preferred language, and other gettext strings if available
```
ayyazdaniaryan: Can you `lxml` here to quickly parse the rendered tree and get the value of an idea in the preferred language set by test, and count the number of `div`s of ideas to verify that rendering is working.

```
from lxml import html
html_tree = html.fromstring(html_content)
```

From there, you can parse the tree according to the appropriate divs and make counts. More info:
http://lxml.de/lxmlhtml.html#html-element-methods

assembl/tests/model_tests/test_notifications.py
```diff
@@ -98,6 +98,64 @@ def test_notification_follow_synthesis(test_session, discussion,
...
+        creation_origin=NotificationCreationOrigin.USER_REQUESTED
+    )
+    test_session.add(subscription)
+
+    initial_notification_count = test_session.query(Notification).count()
+    dispatcher = ModelEventWatcherNotificationSubscriptionDispatcher()
+
+    notification_count = test_session.query(Notification).count()
+    
+    dispatcher.processPostCreated(synthesis_post_1.id)
```
ayyazdaniaryan: I am not clear how this doesn't throw an error in travis env, as a lot of methods downstream of this are supposed to be multi-processing. Perhaps they operate on the same thread, therefore they are perfectly okay on the same thread of execution as this test? Not clear to me.

assembl/tests/model_tests/test_notifications.py
swergas: Yes I wanted to do it in a second pass

assembl/tests/model_tests/test_notifications.py
swergas: Sorry I don't understand, could you explain why it would throw an error? (This method is already used in other tests of this file)

ayyazdaniaryan: Nah, don't worry about it. I just wasn't certain how these methods that are called downstream, which have blocking calls for resources from our database are able to function properly. In any case, this is fine.

PR #329, opened 2018-01-31
Create fab command to setup nginx ssl stapling

mohamedzayed92:


docker/nginx_default.jinja2
```diff
@@ -28,6 +28,9 @@ server {
     add_header Strict-Transport-Security max-age=15768000;
     ssl_stapling on;
     ssl_stapling_verify on;
+    {% if oscp_path %}
+      ssl_trusted_certificate {{ oscp_path }};
```
ayyazdaniaryan: You will want to flush this line as well.

configs/base_env.rc
```diff
@@ -79,6 +79,9 @@ _projectname = assembl
 # _projectpath
 # _dbdumps_dir
 
+# If you wish to have ssl_stapling for nginx, path to certificate
```
ayyazdaniaryan: Rephrase: `Optional - Path to SSL Stapling certificates, must be single file`

fabfile.py
```diff
@@ -1662,6 +1662,20 @@ def docker_compose():
     # run("docker-compose -f docker/build/docker-compose.yml up")
 
 
+@task
+def get_ssl_certificates():
```
ayyazdaniaryan: It's not a `get`, but a `set`, actually.

fabfile.py
```diff
@@ -1662,6 +1662,20 @@ def docker_compose():
...
+    if env.oscp_path:
+        root_certificate = run('curl https://letsencrypt.org/certs/isrgrootx1.pem.txt')
+        intermediate_certificate_1 = run('curl https://letsencrypt.org/certs/lets-encrypt-x3-cross-signed.pem.txt')
+        intermediate_certificate_2 = run('curl https://letsencrypt.org/certs/letsencryptauthorityx3.pem.txt')
+        with open(env.oscp_path, 'w') as certificates_file:
+            for certificate_file in [root_certificate, intermediate_certificate_1, intermediate_certificate_2]:
```
ayyazdaniaryan: Put these elements in a tuple instead of a list. In this case, completely insignificant. But for your future knowledge:
https://stackoverflow.com/questions/68630/are-tuples-more-efficient-than-lists-in-python

fabfile.py
```diff
@@ -1662,6 +1662,20 @@ def docker_compose():
...
+                certificates_file.write(certificate_file)
+                certificates_file.write('\n')
+    else:
+        print("Can't get ssl certificates, env.oscp_path does not exist")
```
ayyazdaniaryan: Use `print(yellow("....")` And it's not that it does not exist, the `oscp_path` is not set.

PR #330, opened 2018-01-31
fix reactions counters on synthesis page

vincentfretin:


assembl/graphql/view.py
```diff
@@ -13,18 +15,40 @@
...
+
             getLogger().debug(
                 'graphql', op=info.operation.operation,
-                opname=info.operation.name.value, vars=variables)
+                opname=info.operation.name.value, vars=modified_variables)
         return next(source, gargs, context, info, *args, **kwargs)
 
 
+def get_graphql_params(request, data):
+    query, variables, operation_name = original_get_graphql_params(request, data)
```
ayyazdaniaryan: Yikes. Fully monkey-patching it eh?

assembl/graphql/view.py
```diff
@@ -13,18 +15,40 @@
...
+    modified_variables = {}
+    for key, value in variables.items():
+        if 'password' in key or 'Password' in key:
+            modified_variables[key] = 'xxxxxxxxxxxxx'
+        else:
+            modified_variables[key] = value
+
+    operation = query.split()[0]
+    getLogger().debug(
```
ayyazdaniaryan: Do we lose logging under non-API call's then? Eg. no logs for testing, etc? Not a big deal otherwise.

assembl/graphql/view.py
```diff
@@ -44,5 +68,8 @@ def graphql_api(request):
     if not discussion.user_can(user_id, CrudPermissions.READ, permissions):
         raise HTTPUnauthorized()
 
-    solver = graphql_wsgi(Schema, middleware=[LoggingMiddleware()])
+    # Using a middleware transforms the request to a promise that
+    # doesn't play nice with multiprocesses or multithreading and sqlalchemy session.
```
ayyazdaniaryan: That's a shame.

assembl/graphql/view.py
```diff
@@ -13,18 +15,40 @@
 class LoggingMiddleware(object):
     def resolve(self, next, source, gargs, context, info, *args, **kwargs):
         if source is None:
-            variables = {}
+            modified_variables = {}
             for key, value in info.variable_values.items():
```
maparent: what about just copying the dict and deleting password?

assembl/graphql/view.py
vincentfretin: I don't understand what you mean. The logging is the same as before my changes, we log in debug every graphql request.

assembl/graphql/view.py
vincentfretin: Keeping the key tells me that the user tried to change her password. We lose this info if you delete the key.

assembl/graphql/view.py
ayyazdaniaryan: no, this is not the same code path. Using a middleware means calling the `execute` will also trigger the logging. However, in tests, when we call `schema.execute`, we miss this logging.

ayyazdaniaryan: Significant? Probably not. But, something to be aware of.

vincentfretin: This is graphql_wsgi which calls schema.execute with given middleware. In tests we use schema.execute directly without middleware, so no logs in tests.  `get_graphql_params` here is used in graphql_wsgi just before the `schema.execute` call. So there is no changes in logging.

PR #332, opened 2018-02-01
calculate csv column data from the social json

maparent:


assembl/models/social_data_extraction.py
```diff
@@ -0,0 +1,88 @@
...
+        json_list = [data for (id, data) in json_list]
+    return get_social_columns_from_json_list(json_list, columns_info)
+
+
+def get_provider_id_for_discussion(discussion):
+    profile_name = discussion.preferences['authorization_server_backend']
+    if profile_name:
+        profile_id = discussion.db.query(IdentityProvider.id).filter_by(provider_type=profile_name)
+        if profile_id:
+            profile_id = (profile_id,)
```
vincentfretin: The query returns a list of tuple right, something like `[(123,)]`
so here profile_id will be something like `[((123,),)]`
this is what you want? just to be sure.

assembl/views/api2/discussion.py
```diff
@@ -1613,6 +1617,13 @@ def phase1_csv_export(request):
         SENTIMENT_CREATION_DATE.encode('utf-8'),
     ]
 
+    extra_columns_info = load_social_columns_info(discussion, language)
+    if extra_columns_info:
+        # insert after email
```
vincentfretin: This is actually after sentiments count, not creator email.
Are we allowed to change the order of the fields here @ayyazdaniaryan ?
This phase1 export is used by bigdatext.

assembl/models/social_data_extraction.py
maparent: good catch.

assembl/views/api2/discussion.py
maparent: corrected, thank you.

maparent:
OK, I'll definitely want Aryan's advice on BDT before merging.

maparent:
I decided to add a GET parameter that would prevent the extra columns from being added, if needed by BDT. I think that closes the issue.

PR #333, opened 2018-02-06
Vote admin step3 backend

vincentfretin:


vincentfretin:
You have now proposals and modules for each proposal.
Modules in vote session are all the modules non attached to a proposal, so the templates from step 2.
(modules are vote specifications)

assembl/graphql/vote_session.py
```diff
@@ -266,6 +294,16 @@ def mutate(root, args, context, info):
...
             )
+            proposal_id = args.get('proposal_id')
+            if proposal_id:
+                proposal_id = int(Node.from_global_id(proposal_id)[1])
+                proposal = models.Idea.get(proposal_id)
+                vote_spec.criterion_idea = proposal
+
+            vote_spec_template_id = args.get('vote_spec_template_id')
+            if vote_spec_template_id:
+                vote_spec.settings = json.dumps({'vote_spec_template_id': vote_spec_template_id})
```
ayyazdaniaryan: This can be a bit unsafe. Verify first that it is a JSON-type before you putting this blob into the db.

assembl/graphql/vote_session.py
```diff
@@ -509,6 +576,16 @@ def mutate(root, args, context, info):
                 nb_ticks=args['nb_ticks'],
                 unit=args['unit']
             )
+            proposal_id = args.get('proposal_id')
```
ayyazdaniaryan: This copy/pasta can become a method. Repeated across multiple mutations.

assembl/graphql/vote_session.py
```diff
@@ -545,10 +623,117 @@ def mutate(root, args, context, info):
...
+    success = graphene.Boolean()
+
+    @staticmethod
+    @abort_transaction_on_exception
+    def mutate(root, args, context, info):
+        proposal_id = args.get('id')
+        proposal_id = int(Node.from_global_id(proposal_id)[1])
+        proposal = models.Idea.get(proposal_id)
+        require_instance_permission(CrudPermissions.DELETE, proposal, context)
+        proposal.db.delete(proposal)
```
ayyazdaniaryan: You'd want to delete the `idea_link` (or tombstone it) as well, no?

assembl/static2/js/app/graphql/VoteSession.graphql
```diff
@@ -10,58 +67,29 @@ query VoteSession($discussionPhaseId: Int!, $lang: String!) {
...
-            ...langStringEntry
-          }
-        }
+      modules: voteSpecifications {
+        ...tokenVoteSpecification
+        ...numberGaugeVoteSpecification
+        ...gaugeVoteSpecification
       }
     }
+    modules: voteSpecifications {
```
ayyazdaniaryan: Why repeat these queries if one of the leaves (`proposal`) is fetching it?

assembl/graphql/vote_session.py
ayyazdaniaryan: Also, why put Relay style id's into the settings instead of database id's?

assembl/graphql/vote_session.py
```diff
@@ -545,10 +623,117 @@ def mutate(root, args, context, info):
...
+            proposal = cls(
+                discussion_id=discussion_id,
+                title=title_ls,
+                description=description_ls
+            )
+            db.add(proposal)
+            identifier = 'voteSession{}'.format(vote_session_id)
+            root_thematic = get_root_thematic_for_phase(discussion, identifier)
+            if root_thematic is None:
+                root_thematic = create_root_thematic(discussion, identifier)
```
ayyazdaniaryan: So you HAVE to create a proposal in order for the root thematic for vote session to exist? Are proposals now the "questions" that are asked in each line, where tokens/gauges can be associated to?

vincentfretin: proposal are actually Idea objects (not Question) inside a hidden Thematic (identifier field is only on Thematic)

assembl/graphql/vote_session.py
vincentfretin: IdeaLink is removed when you delete the idea. We could tombstone both idea and idealink instead. Do we really want to keep proposals history?

assembl/static2/js/app/graphql/VoteSession.graphql
vincentfretin: voteSession.modules are the templates only, those non associated to a proposal.
modules in a proposal are the modules associated to this proposal only.

assembl/graphql/vote_session.py
vincentfretin: Which copy pasta are you talking about here, args.get?

assembl/graphql/vote_session.py
vincentfretin: I don't see what is unsafe here. The vote_spec_template_id here is already validated by graphene, it's an ID type, so basically a string.

I should probably use a vote_spec_template_id foreign key, yes, I was just lazy yesterday. I need to write another alembic step, sigh.

vincentfretin: Ok, it's now a forieign key

assembl/graphql/vote_session.py
vincentfretin: I tombstone now the proposal instead of deleting it in the DeleteProposal mutation.

vincentfretin:
I'll merge this in develop. If there are other changes needed, we can do that in a branch. @MilanBarande needs this to work on step3.

PR #339, opened 2018-02-12
refactor tabbedColumns and multiColumns to use the same code

vincentfretin:
refactor tabbedColumns and multiColumns to use the same code, this will avoid futur bugs


assembl/static2/js/app/components/debate/multiColumns/postColumn.jsx
```diff
@@ -43,7 +42,7 @@ const PostColumn = ({
   width,
   withColumnHeader
 }: Props) => (
-  <div className="column-view" style={{ width: width }}>
+  <div className="column-view" style={width && { width: width }}>
```
cedricmessiant: do you really need this ? From what I remember, the property is ignored if its value is `undefined`

vincentfretin: You're right, this is not needed.

PR #343, opened 2018-02-15
Create automated backup script for piwik

mohamedzayed92:


fabfile.py
```diff
@@ -183,6 +183,50 @@ def update_vendor_config():
         system("cd %s ; git pull" % config_file_dir)
 
 
+@task
+def docker_fabric_task():
```
swergas: bad task name. Please rename so we know exactly what this task does. Example : docker_piwik_database_backup

fabfile.py
```diff
@@ -183,6 +183,50 @@ def update_vendor_config():
...
+                run("cd piwik-docker-compose && docker-compose up -d")
+                run("docker exec -d piwikdockercompose_db_1 mysqldump --result-file=dump.sql")
+            except:
+                run("cd piwik-docker-compose && docker-compose up -d")
+                run("docker exec -d piwikdockercompose_db_1 mysqldump --result-file=dump.sql")
+        cli.copy_resource(u"piwikdockercompose_db_1", '/dump.sql', cwd + "/dump.sql")
+    else:
+        get('/etc/nginx/sites-available', cwd)
+
+    def copy_resource_from_localhost_to_remote_container(resource, container, temp, dest_file_name):
```
swergas: This function seems to not be used anywhere in the code. Then remove it?

fabfile.py
```diff
@@ -1628,6 +1672,23 @@ def as_rc(ini_filename):
     return r
 
 
+@task
+def backup_piwik_data():
```
swergas: Bad task name, as it does not really do what it says. This task seems to just fill a Jinja template and upload it to a server, it does not backup anything, right? Example of task name that could be used instead: upload_backup_script

fabfile.py
```diff
@@ -183,6 +183,50 @@ def update_vendor_config():
...
+            try:
+                run("git clone git@github.com:siutin/piwik-docker-compose.git")
+                run("cd piwik-docker-compose && docker-compose up -d")
+                run("docker exec -d piwikdockercompose_db_1 mysqldump --result-file=dump.sql")
+            except:
+                run("cd piwik-docker-compose && docker-compose up -d")
+                run("docker exec -d piwikdockercompose_db_1 mysqldump --result-file=dump.sql")
+        cli.copy_resource(u"piwikdockercompose_db_1", '/dump.sql', cwd + "/dump.sql")
+    else:
+        get('/etc/nginx/sites-available', cwd)
```
swergas: I'm not sure I understand this line. What is it meant to do?

fabfile.py
```diff
@@ -183,6 +183,50 @@ def update_vendor_config():
...
+    from dockerfabric.utils.files import temp_dir
+    cwd = os.getcwd()
+    if env.docker_piwik:
+        with temp_dir() as remote_tmp:
+            print remote_tmp
+            try:
+                run("git clone git@github.com:siutin/piwik-docker-compose.git")
+                run("cd piwik-docker-compose && docker-compose up -d")
+                run("docker exec -d piwikdockercompose_db_1 mysqldump --result-file=dump.sql")
+            except:
```
swergas: This try/except block looks a bit weak, maybe rather check for the presence of `piwik-docker-compose` folder, and if it does not exist then do the `git clone` ?

fabfile.py
mohamedzayed92: @swergas this line download the data base backup in the case of not using piwik inside a docker container, and stores it on your localhost, normally, the databse backup is supposed to be in this location

fabfile.py
```diff
@@ -183,6 +183,32 @@ def update_vendor_config():
...
+    cwd = os.getcwd()
+    if env.docker_piwik:
+        with temp_dir() as remote_tmp:
+            print remote_tmp
+            from fabric.contrib import files
+            if files.exists("/home/assembl_user/piwik-docker-compose/docker-compose.yml"):
+                run("cd piwik-docker-compose && docker-compose up -d")
+                run("docker exec -d piwikdockercompose_db_1 mysqldump --result-file=dump.sql")
+            else:
+                run("git clone git@github.com:siutin/piwik-docker-compose.git")
```
ayyazdaniaryan: Again, these are just checking that docker-piwik is installed or not. That is not the responsibility of this task. You should instead check, via rc files, if the targeted server has a piwik container or a piwik folder. Then, run the `mysqldump` and move the file appropriately.

fabfile.py
```diff
@@ -183,6 +183,32 @@ def update_vendor_config():
...
+                run("cd piwik-docker-compose && docker-compose up -d")
+                run("docker exec -d piwikdockercompose_db_1 mysqldump --result-file=dump.sql")
+        cli.copy_resource(u"piwikdockercompose_db_1", '/dump.sql', cwd + "/dump.sql")
```
ayyazdaniaryan: Instead of `cwd`, you should either use the `projectdirectory` in the fabric env (from rc files) or define a path for the dump in the rc file for this. Same goes for container name. 

Also, does this copy the resource from outside the container to inside? If so, you need to fetch it from inside the container and move it to the route that is managed by BORG.

borg/assembl_borg_backup.sh.jinja2
```diff
@@ -0,0 +1,41 @@
...
+cd $ASSEMBL_PATH
+#In case the database backup fails
+$ASSEMBL_PATH/venv/bin/assembl-db-manage local.ini backup
+#Make sure we back up the database dump from the last deployment:
+cp --dereference $ASSEMBL_PATH/assembl-backup.pgdump $ASSEMBL_PATH/assembl-backup-real.pgdump
+NAME="`hostname`-`basename $ASSEMBL_PATH`-`date --iso-8601='minutes'`"
+#set -x
+borg create \
+    $REPOSITORY::$NAME \
+    $ASSEMBL_PATH \
```
ayyazdaniaryan: You should also add a path to an RC file, where piwik backups will go to on the server, and add that path in the template here. That's part of the reason for moving this document to a template :)

mysql_dump_script.sh
```diff
@@ -0,0 +1,21 @@
...
+ 
+#on boucle sur chaque dossier (for découpe automatiquement par l'espace)
+for SQL in $LISTEBDD
+ 
+do
+ 
+if [ $SQL == "piwik" ]; then
+ 
+#echo $SQL
+mysqldump --user=root --password=AHJnd028.5m! $SQL | gzip > /home/$SQL"_mysql-new_4"$DATE.sql.gz
```
swergas: This script file should not have a password written in it, but should instead read a password from configuration file or command argument. Also the output path and the file name do not look correct

mysql_dump_script_docker.sh
```diff
@@ -0,0 +1,21 @@
...
+ 
+#on boucle sur chaque dossier (for découpe automatiquement par l'espace)
+for SQL in $LISTEBDD
+ 
+do
+ 
+if [ $SQL == "piwik" ]; then
+ 
+#echo $SQL
+mysqldump --user=root --password=wwitri2Dzhishme $SQL | gzip > /home/$SQL"_mysql_docker"$DATE.sql.gz
```
swergas: This script file should not have a password written in it, but should instead read a password from configuration file or command argument. Also the output path and the file name do not look correct

mohamedzayed92: I fully agree ... this is why I templated the mysqldump file

configs/bluenove-server-configs/piwik_docker_mysql_dump.rc
```diff
@@ -0,0 +1,7 @@
+_extends = base_env.rc 
+public_hostname = culturemanager.assembl.org
+assembl_user=mozayed
+_projectpath = /Users/mozayed/Assembl
+mysql_password=wwitri2Dzhishme
```
swergas: This is still not in the right repository.

configs/bluenove-server-configs/piwik_mysql_dump.rc
```diff
@@ -0,0 +1,6 @@
+_extends = base_env.rc 
+public_hostname = stats.bluenove.com
+_projectpath = /home/assembl_user/assembl
+assembl_user=mozayed
+mysql_password=AHJnd028.5m!
```
swergas: This is still not in the right repository.

swergas:
The `configs/bluenove-server-configs` is specific to bluenove and is managed by another repository. No files should be commited in this folder on the `assembl` repository. But this PR adds files `piwik_docker_mysql_dump.rc` and `piwik_mysql_dump.rc` in this folder. Please move them instead either to `configs/` if they are considered as generic enough and reusable enough by other users, or to the other repository if they are only for us or contain sensitive information.

maparent:
Seconding what Quentin said, and since the password _was_ leaked, it is mandatory to change it.

PR #344, opened 2018-02-16
various css fixes with Fred and Roxana

felixsaintgenis:


assembl/static2/css/components/text.scss
```diff
@@ -110,6 +110,14 @@
   font-family: $font-family4;
 }
 
+.light-title-7 {
```
ayyazdaniaryan: Title 6 and 7 are not shown in the `styleguide`, especially in these formats. @paolina92 and I can help you edit the styleguide so that everyone can visually see what these changes that are being proposed looks like.

assembl/static2/css/variables.scss
```diff
@@ -123,6 +123,7 @@ $font-size-sm: 16px;
 $font-size-m: 18px;
 $font-size-l: 26px;
 $font-size-xl: 36px;
+$font-size-xxl: 48px;
```
ayyazdaniaryan: We could even add these to styleguide, since they aren't there.

assembl/static2/css/components/text.scss
swergas: Also, if we are getting to 7 different light title styles, isn't there a problem of unity in the UI?

assembl/static2/css/views/administration.scss
```diff
@@ -72,14 +72,21 @@
...
     }
 
     .admin-content {
       padding: 30px 40px 80px;
       text-align: center;
     }
 
+    .admin-paragraph {
+      font-size: 13px;
+      font-family: LatoMedium;
```
swergas: you should probably rather use $font-family2

assembl/static2/js/app/components/administration/landingPage/selectModulesForm.jsx
```diff
@@ -21,9 +21,8 @@ export const DumbSelectModulesForm = ({ hasErrors, modulesByIdentifier, moduleTy
           return (
             <Checkbox
               key={moduleType.identifier}
-              checked={module && module.get('enabled')}
-              onChange={() => toggleModule(moduleType.identifier)}
-              disabled={moduleType.required}
```
swergas: you should probably keep this disabled line

felixsaintgenis: Roxana wants the required checkbox to be in the same style as the other. Should I change the checkbox required style with CSS rather than changing the js file ?

assembl/static2/css/variables.scss
cedricmessiant: the more things are in the styleguide, the happier I am !

PR #347, opened 2018-02-23
Fix root page

swergas:


assembl/lib/frontend_urls.py
```diff
@@ -237,7 +237,8 @@ def get_discussion_url(self, request=None, force_v1=False):
         else:
             get_route = create_get_route(request, self.discussion)
             if force_v1:
-                route = get_route('oldDebate')
```
maparent: OR: route = get_route('home-auto')
(which I think preserves the idea better.)

PR #348, opened 2018-02-23
fix email share button 

swergas:
(bug from react-share lib, see https://github.com/nygardk/react-share/issues/132)


MilanBarande:
Great! I had tried to investigate this bug a couple of months ago and noticed it was an issue with the lib but I hadn't found this solution at the time. Not much to be said, good to merge if it solves the issue

PR #350, opened 2018-02-26
updating opengraph and twitter meta data

mohamedzayed92:


assembl/views/__init__.py
```diff
@@ -362,9 +363,78 @@ def process_export_list(ls):
...
+    return locale
+
+
+def get_description(request):
+    """Returns the description corresponding to the locale of the current discussion"""
+    opengraph_locale = get_opengraph_locale(request)
+    from ..auth.util import get_current_discussion
+    discussion = get_current_discussion()
+    if discussion:
+        if discussion.preferences["extra_json"]["objectives"]["descriptionEntries"][opengraph_locale]:
```
ayyazdaniaryan: This code is not defensive enough. In the `extra_json`, if `objectives` is not set (which is very possible), then this will crash and the page will fail to load. Code more defensively. You can use python `dict`'s `get` method to pass a default value.

assembl/views/__init__.py
```diff
@@ -362,9 +363,78 @@ def process_export_list(ls):
...
+            return discussion.preferences["extra_json"]["objectives"]["descriptionEntries"][locale]
+
+
+def get_topic(request):
+    """Returns the topic corresponding to the locale of the current discussion"""
+    opengraph_locale = get_opengraph_locale(request)
+    from ..auth.util import get_current_discussion
+    discussion = get_current_discussion()
+    if discussion:
+        if discussion.preferences["extra_json"]["topic"]["titleEntries"][opengraph_locale]:
```
ayyazdaniaryan: Similarly, more defensive coding needed.

assembl/views/__init__.py
```diff
@@ -362,9 +363,78 @@ def process_export_list(ls):
...
+            locale = discussion.preferences['preferred_locales'][0]
+            return discussion.preferences["extra_json"]["topic"]["titleEntries"][opengraph_locale]
+
+
+def get_landing_page_image():
+
+    from ..auth.util import get_current_discussion
+    discussion = get_current_discussion()
+    if discussion:
+        return discussion.preferences['extra_json']['headerBackgroundUrl']
```
ayyazdaniaryan: Yep, more defence.

assembl/views/__init__.py
```diff
@@ -362,9 +363,78 @@ def process_export_list(ls):
...
+        "get_topic": get_topic(request)
+    })
+
+    return base
+
+
+def private_social_sharing():
+
+    from ..auth.util import get_current_discussion
+    discussion = get_current_discussion()
```
ayyazdaniaryan: You could have avoided the use of this method and fetched the discussion from the request. However, this is fair enough.

assembl/templates/base.jinja2
```diff
@@ -6,20 +6,48 @@
...
+            <meta name="twitter:title" content="Assembl">
+            <meta name="twitter:description" content="{{get_description}}">
+            <meta name="twitter:image" content="{{get_landing_page_image}}">
+        {% endif %}
+    {% endblock meta_tags_public %}
+    
+    {% block meta_tags_private %}
+        {% if private_social_sharing %}
+            
+            <meta name="description" content="">
```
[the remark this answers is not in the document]
mohamedzayed92: spoke with fred and Frank about it, they don't want description for private discussion, actually, they even said that there should not be twitter sharing for private debates ...

assembl/lib/frontend_urls.py
```diff
@@ -237,7 +238,7 @@ def get_discussion_url(self, request=None, force_v1=False):
             from assembl.views import create_get_route
             get_route = create_get_route(request, self.discussion)
             if force_v1:
-                route = get_route('home')
+                route = self.get_frontend_url('oldDebate', slug=self.discussion.slug)
```
swergas: probably a merge issue here: the correct one is `route = get_route('home')`

assembl/templates/base.jinja2
```diff
@@ -6,20 +6,46 @@
...
-    <meta name="googlebot" content="noindex" />
-    {% endblock meta_tags %}
+    {% block meta_tags_public %}
+        {% if not private_social_sharing %}
+            <meta name="description" content="{{get_description}}">
+            <meta property="og:title" content="{{get_topic}}" />
+            <meta property="og:type" content="website" />
+            <meta property="og:url" content="{{ application_url }}" />
+            <meta property="og:image" content="{{get_landing_page_image}}" />
+            <meta property="og:image:width" content="1600" />
```
swergas: Have you tried to use real image width and height? Use them if possible, otherwise is it allowed by the standard to just not provide this information?

assembl/templates/base.jinja2
```diff
@@ -6,20 +6,46 @@
...
+            <meta property="og:image:height" content="222" />
+            <meta property="og:locale" content="{{opengraph_locale}}" />
+            <meta name="robots" content="noindex" />
+            <meta name="googlebot" content="noindex" />
+
+            <meta name="twitter:card" content="summary_large_image">
+            <meta name="twitter:site" content="@bluenove">
+            <meta name="twitter:creator" content="@bluenove">
+            <meta name="twitter:title" content="Assembl">
```
swergas: You should probably use the same value for twitter:title as for og:title (`{{get_topic}}`)

assembl/templates/base.jinja2
```diff
@@ -6,20 +6,46 @@
...
+            <meta property="og:url" content="{{ application_url }}" />
+            <meta property="og:image" content="{{get_landing_page_image}}" />
+            <meta property="og:image:width" content="1600" />
+            <meta property="og:image:height" content="222" />
+            <meta property="og:locale" content="{{opengraph_locale}}" />
+            <meta name="robots" content="noindex" />
+            <meta name="googlebot" content="noindex" />
+
+            <meta name="twitter:card" content="summary_large_image">
+            <meta name="twitter:site" content="@bluenove">
```
swergas: Are you sure we can't provide a more custom per-debate value for twitter:site and twitter:creator?

assembl/templates/base.jinja2
```diff
@@ -6,20 +6,46 @@
...
+            <meta property="og:image" content="{{ application_url }}/static/css/themes/default/img/crowd2.jpg" />
+            <meta property="og:image:width" content="1600" />
+            <meta property="og:image:height" content="222" />
+            <meta property="og:locale" content="{{opengraph_locale}}" />
+
+            <meta name="twitter:card" content="summary_large_image">
+            <meta name="twitter:site" content="@bluenove">
+            <meta name="twitter:creator" content="@bluenove">
+            <meta name="twitter:title" content="Assembl">
+            <meta name="twitter:description" content="">
```
swergas: probably use {{get_description}}

assembl/views/__init__.py
```diff
@@ -362,9 +363,104 @@ def process_export_list(ls):
...
+    if discussion:
+        dict = discussion.preferences["extra_json"]
+        objectives_dict = dict.get("objectives", "default objectives")
+        if type(objectives_dict) == str:
+            return objectives_dict
+        else:
+            objectives_dict = objectives_dict["descriptionEntries"]
+            locale = discussion.preferences['preferred_locales'][0]
+            return objectives_dict.get(opengraph_locale, objectives_dict[locale])
```
swergas: Did you make sure that whatever the content, a preview shows nicely? Namely to escape characters that could break <meta> tag structure or that could show not right in the social network's preview? Like quotes and html tags for example

assembl/templates/base.jinja2
```diff
@@ -6,20 +6,46 @@
...
-    <meta property="og:image" content="{{ application_url }}/static/css/themes/default/img/crowd2.jpg" />
-    <meta property="og:image:width" content="1600" />
-    <meta property="og:image:height" content="222" />
-    <meta property="og:image" content="{{ application_url }}/static/css/themes/default/img/logo.png" />
-    <meta property="og:image:width" content="111" />
-    <meta property="og:image:height" content="35" />
-    <meta name="robots" content="noindex" />
-    <meta name="googlebot" content="noindex" />
-    {% endblock meta_tags %}
+    {% block meta_tags_public %}
```
swergas: You should probably also do the same in `templates/index_react.jinja2` than in `templates/index.jinja2`: have the same `private_social_sharing` condition and the different cases of general debate (landing page and any non-idea and non-post URL), idea page, and direct link to a message. Or maybe the scope of this task was only about the landing page? (I think while we are at it, it is good to do it for ideas and messages)

assembl/templates/base.jinja2
mohamedzayed92: actually no ... the purpose is to make the description not visible when it is a private debate, so after discussions with fred and frank, they told me to leave this empty

assembl/lib/frontend_urls.py
mohamedzayed92: I did a rebase so should be fine by now

assembl/templates/base.jinja2
mohamedzayed92: It is done, however, I believe that later on I should probably modify the index file ... @ayyazdaniaryan  said that it is not recommended to do function calls in jinja2 templates, so probably later I will modify this file again to only use variables and not function calls

assembl/templates/base.jinja2
mohamedzayed92: Well on the twitter documentation, this was the given dimension and it appears fine when tested, I don't really get what you mean by real image width and height :D

assembl/templates/base.jinja2
mohamedzayed92: well theoritically I can add whatever value I want, or write a function that renders any result that I want ... but I don't have an idea on what should be put instead ... what do you propose ?

ayyazdaniaryan: Well, `@bluenove` isn't valid here either. It should be the url of the discussion. This can be fetched from the discussion object as `{{discussion.get_discussion_urls()[1]}}` which would deliver the `https` based url of the discussion.

assembl/templates/base.jinja2
ayyazdaniaryan: So for a private debate, please remove the meta tag if it is not a required tag. Semantically, it means something to have the meta tag with `content=""`.

assembl/views/__init__.py
```diff
@@ -371,12 +371,25 @@ def process_export_list(ls):
...
     })
 
     return base
 
 
+def get_discussion_url():
+    from ..auth.util import get_current_discussion
+    discussion = get_current_discussion()
+    if discussion:
+        if (len(discussion.get_discussion_urls()) == 2):
```
ayyazdaniaryan: Okay, I know I said that you could use one of these, but I take it back. Remove this check block. Instead, use the `FrontendUrls` object (along with the discussion object) to call the `get_discussion_url`. This will ensure the correct URL, including port number, scheme, V1 vs V2 url, etc.

assembl/views/__init__.py
```diff
@@ -371,12 +371,25 @@ def process_export_list(ls):
...
+            return discussion.get_discussion_urls()[1]
+        else:
+            return discussion.get_discussion_urls()[0]
+    else:
+        return "bluenove"
```
ayyazdaniaryan: No, there should be a fallback. If there is no discussion, I think you can give the global URL of the server as absolute fallback. You can get it by using `get_global_base_url` method.

PR #351, opened 2018-02-26
Debate giant menu scroll by columns

amensouissi:


assembl/static2/js/app/components/debate/navigation/debateLink.jsx
```diff
@@ -32,11 +42,20 @@ class DebateLink extends React.Component<*, DebateLinkProps, DebateLinkState> {
     this.setState({ menuActive: false });
   };
 
+  handleClickOutside = (event: MouseEvent) => {
+    if (this.debate && !this.debate.contains(event.target)) {
```
swergas: Maybe remove the event listener after hide? (And create it on hover) Otherwise this function will be called on every click in the application for the whole user navigation, right?

PR #354, opened 2018-03-02
Don't show deleted top posts without answers

vincentfretin:


vincentfretin:
@ayyazdaniaryan Currently in the backend we tombstone a post if there is no live descendants, and in the graphql query we get the tombstoned posts as well.
Here the behavior we want is to show tombstoned posts when it's not a top post. So I filter out from the frontend here.

PR #356, opened 2018-03-06
Richtext survey

MilanBarande:


assembl/static2/js/app/components/debate/survey/question.jsx
```diff
@@ -55,21 +69,36 @@ class Question extends React.Component {
...
+  updateCharCount = (newValue) => {
+    this.setState({ charCount: newValue });
+  };
 
   render() {
     const { index, title, screenWidth, screenHeight } = this.props;
-    const height = screenHeight - document.getElementById('timeline').clientHeight;
+    let height = 0;
+    const timeline = document && document.getElementById('timeline');
+    // This is necessary to bypass an issue with Flow
```
MilanBarande: https://stackoverflow.com/questions/44979394/why-flow-still-complains-about-null-values-for-document-getelementbyid

PR #357, opened 2018-03-07
sticky-topPostForm-bug

MilanBarande:
coarsePostsCount only counts parent posts but it's only needed to know if there is at least one post in the thread

### DIFF
--- a/assembl/static2/js/app/components/debate/common/topPostFormContainer.jsx
+++ b/assembl/static2/js/app/components/debate/common/topPostFormContainer.jsx
@@ -11,7 +11,8 @@ type TopPostFormContainerProps = {
   messageColumns: Object,
   isColumnViewInline: boolean,
   ideaId: string,
-  refetchIdea: Function
+  refetchIdea: Function,
+  coarsePostsCount: number
 };
 
 type TopPostFormContainerState = {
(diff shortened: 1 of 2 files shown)


assembl/static2/js/app/components/debate/common/topPostFormContainer.jsx
```diff
@@ -82,10 +83,12 @@ class TopPostFormContainer extends React.Component<*, TopPostFormContainerProps,
   }
 
   render() {
-    const { ideaId, refetchIdea, messageColumns = [], isColumnViewInline } = this.props;
+    const { ideaId, refetchIdea, messageColumns = [], isColumnViewInline, coarsePostsCount } = this.props;
     const columnsInfos = this.getColumnsInfos();
+    const { sticky } = this.state;
+    const containerClassNames = sticky && messageColumns.length <= 1 && coarsePostsCount >= 1 ? 'top-post-sticky' : '';
```
ayyazdaniaryan: Can leverage `classnames` here

MilanBarande: Isn't classnames only useful if you have a case with certain classes and another with the same + others ? In this case there is only one class which is present under certain conditions so I don't see why the library is necessary.

vincentfretin: I agree, classnames is overkill just for one class.

PR #359, opened 2018-03-07
Vote results

vincentfretin:


assembl/graphql/vote_session.py
```diff
@@ -247,6 +303,18 @@ class Meta:
...
+    average_result = graphene.Float(required=True)
+
+    def resolve_average_result(self, args, context, info):
+        vote_cls = self.get_vote_class()
+        voting_avg = self.db.query(func.avg(getattr(vote_cls, 'vote_value'))).filter_by(
+            vote_spec_id=self.id,
+            tombstone_date=None,
+            idea_id=self.criterion_idea_id).first()
+        # when there is no votes, query.first() equals (None,)
+        avg = voting_avg[0] or 0
```
ayyazdaniaryan: Okay, this is a repeating pattern, let's turn to a helper function

assembl/tests/fixtures/vote_session.py
```diff
@@ -286,3 +294,121 @@ def fin():
 
     request.addfinalizer(fin)
     return vote_spec
+
+
+@pytest.fixture(scope="function")
+def number_gauge_vote_specification_associated_to_proposal(request, test_session, discussion, graphql_request, vote_session, number_gauge_vote_specification, vote_proposal, graphql_registry):
+    mutation = graphql_registry['createNumberGaugeVoteSpecification']
+    vote_session_id = to_global_id("VoteSession", vote_session.id)
```
ayyazdaniaryan: Can probably get away with `vote_session.graphene_id()` too yeah?

assembl/graphql/idea.py
```diff
@@ -152,6 +152,12 @@ def resolve_index(self, args, context, info):
         return count
 
 
+class VoteResults(graphene.ObjectType):
+
+    num_participants = graphene.Int()
+    participants = graphene.List(AgentProfile)
```
cedricmessiant: I think we don't need this for now but it will be useful in the future

assembl/graphql/vote_session.py
cedricmessiant: I'm not sure about this. How would you call this function ?

assembl/static2/js/app/graphql/VoteSession.graphql
```diff
@@ -24,13 +25,21 @@ fragment tokenVoteSpecification on TokenVoteSpecification {
...
     ... on TokenVote {
       voteValue
       proposalId
       tokenCategoryId
     }
   }
+  numVotes
+  tokenVotes {
+    tokenCategoryId
+    numToken
```
cedricmessiant: isn't it more `tokenCount` ?

cedricmessiant: (I'm not a big fan of `num` in variables name but it also works I guess: `numToken` => "number of tokens"

vincentfretin: That's right, number of tokens. I wanted some consistency on the naming. We already have numContributors, numPosts...

assembl/tests/fixtures/vote_session.py
vincentfretin: probably

assembl/graphql/vote_session.py
vincentfretin: What is the pattern? I don't see it. This is a specific query. Or you're talking about just the `voting_avg[0] or 0` part? You want to call a function just for this?

ayyazdaniaryan:
Latest commits reviewed. LGTM.

PR #361, opened 2018-03-08
Task to get snapshot on etalab server, Task to set the snapshot on etalab server

mohamedzayed92:


configs/bluenove-server-configs/etalab-snapshot.rc
```diff
@@ -0,0 +1,7 @@
+#In order to get credentials for your script, log into https://api.ovh.com/createToken
```
swergas: This file should not be created in this repository but in the other one

fabfile.py
```diff
@@ -165,6 +165,40 @@ def listdir(path):
...
+        consumer_key=env.consumer_key,
+    )
+
+    print "Welcome", client.get('/me')['firstname']
+    result = client.get('/vps/' + env.vps_identifier + '/snapshot')
+    print json.dumps(result, indent=4)
+
+
+@task
+def set_snapshot_etalab_server():
```
swergas: Maybe make the name of these 2 commands more generic? Something like create_ovh_vps_snapshot and get_ovh_vps_snapshot ? We have abstracted all our fabfile commands to not mention a perticular server we own but rather a type of server that has some characteristics (sentry, piwik, certbot, etc).

configs/bluenove-server-configs/etalab-snapshot.rc
swergas: And probably these new keys should be in a category like `[ovh_hosting]` and documented in `doc/configuration.rst`

mohamedzayed92: yeah good idea ... probably should be documented in doc/configuration.rst ... but what do you mean by in a new category [ovh_hosting] ?

swergas: I mean that when you open local.ini, you see that variables are into groups that are shown like [this]

swergas: And in an .rc file, when you define a variable that is in a group, the variable name starts with the group name, then double underscore, then the variable name. Example: `theme_repositories__git-urls`

fabfile.py
```diff
@@ -165,6 +166,197 @@ def listdir(path):
...
+                                      'DKIM',
+                                      'LOC',
+                                      'MX',
+                                      'NAPTR',
+                                      'NS',
+                                      'PTR',
+                                      'SPF',
+                                      'SRV',
+                                      'SSHFP',
+                                      'TXT'])
```
vincentfretin: There is no need to create this dict each time you call the function. You should create this out of the function and name it other than type which override the built-in, considered bad practice.

fabfile.py
```diff
@@ -165,6 +166,197 @@ def listdir(path):
...
+    return field in type['choices']
+
+
+@task
+def check_dns_entries(data):
+    """Checks if the inputs in your dns_entries.json file are valid"""
+    my_list = []
```
vincentfretin: Really bad name. What this variable represents? entries_types?

fabfile.py
```diff
@@ -165,6 +166,197 @@ def listdir(path):
...
+            my_list.append(base["type"])
+    for item in my_list:
+        print item
+
+    if all(check_field_type_exists(item) for item in my_list):
+        return True
+    else:
+        for item in my_list:
+            if not check_field_type_exists(item):
+                print "The field type " + item + "is not valid"
```
vincentfretin: You should use the format syntax.

fabfile.py
```diff
@@ -165,6 +166,197 @@ def listdir(path):
...
+@task
+def check_domains(data):
+    """Checks if the domains you provided in the dns_entries.json are valid"""
+    domain_list = get_domain_list()
+    if all(domain in domain_list for domain in data):
+        return True
+    else:
+        for domain in data:
+            if domain not in domain_list:
+                print "The domain " + domain + " is not in the domain list"
```
vincentfretin: You should use the format syntax.

fabfile.py
```diff
@@ -165,6 +166,197 @@ def listdir(path):
...
+                return False
+
+
+@task
+def get_dns_entries_from_file():
+    import json
+    data = json.load(open('configs/bluenove-server-configs/dns_entries.json'))
+    print json.dumps(data, indent=4)
```
vincentfretin: You should maybe create a function print_json at this point.

fabfile.py
```diff
@@ -165,6 +166,197 @@ def listdir(path):
...
+    elif not check_field_type_exists(field_type_entry):
+        print "The field " + field_type_entry + " is not a valid field type"
+        print "Valid field types are A, AAAA, CNAME, DKIM, LOC, MX, NAPTR, NS, PTR, SPF, SRV, SSHFP, TXT"
+    else:
+        print "Creating new DNS record for the domain " + domain_entry + " with field type " + field_type_entry + " with target " + target_entry
+        client.post('/domain/zone/' + domain_entry + '/record', fieldType=field_type_entry, target=target_entry)
+
+
+@task
+def create_ovh_client():
```
vincentfretin: What the point of being a task?

fabfile.py
```diff
@@ -165,6 +166,197 @@ def listdir(path):
...
+        print "The domain " + domain_entry + " is not a valid domain"
+        print "The valid domains are:"
+        for domain in get_domain_list():
+            print domain
+    elif not check_field_type_exists(field_type_entry):
+        print "The field " + field_type_entry + " is not a valid field type"
+        print "Valid field types are A, AAAA, CNAME, DKIM, LOC, MX, NAPTR, NS, PTR, SPF, SRV, SSHFP, TXT"
+    else:
+        print "Creating new DNS record for the domain " + domain_entry + " with field type " + field_type_entry + " with target " + target_entry
+        client.post('/domain/zone/' + domain_entry + '/record', fieldType=field_type_entry, target=target_entry)
```
vincentfretin: Use the format syntax.

fabfile.py
```diff
@@ -165,6 +166,197 @@ def listdir(path):
...
+
+    print json.dumps(result, indent=4)
+
+
+@task
+def remove_secondary_dns_names_ovh():
+    """Read list of secondary domain names to remove from the file domain_name_list.rc"""
+    client = create_ovh_client()
+    with open("configs/bluenove-server-configs/domain_name_list.rc", "r") as f:
+        list_of_domaines_to_remove = f.readlines()
```
vincentfretin: `domains`, not `domaines`. And `domains_to_remove` is enough.

fabfile.py
```diff
@@ -165,6 +166,197 @@ def listdir(path):
...
+    print json.dumps(result, indent=4)
+
+
+@task
+def add_secondary_dns_names_ovh():
+    """Read list of secondary domain names to add from the file domain_name_list.rc"""
+    client = create_ovh_client()
+
+    with open("configs/bluenove-server-configs/domain_name_list.rc", "r") as f:
+        list_of_domains_to_add = f.readlines()
```
vincentfretin: Just call it domains_to_add, not need to specify the type in the variable. If the variable is plural, then it's a list.

fabfile.py
```diff
@@ -165,6 +180,198 @@ def listdir(path):
...
+    record_ids = client.get('/domain/zone/{domain}/record/'.format(domain=domain))
+    for record_id in record_ids:
+        info = client.get('/domain/zone/{domain}/record/{record_id}'.format(domain=domain, record_id=record_id))
+        records[info['subDomain']] = info
+    print json.dumps(records, indent=4)
+    return records
+
+
+@task
+def delete_dns_record(domain, entry_id):
```
ayyazdaniaryan: How is the `entry_id` identified? Is it the same as the `record_id`?

fabfile.py
```diff
@@ -165,6 +180,198 @@ def listdir(path):
...
+def create_new_dns_record_cmd_line_arguments(domain_entry, field_type_entry, target_entry):
+    client = create_ovh_client()
+    if domain_entry not in get_domain_list():
+        print("The domain {domain_entry} is not a valid domain".format(domain_entry=domain_entry))
+        print "The valid domains are:"
+        for domain in get_domain_list():
+            print domain
+    elif not check_field_type_exists(field_type_entry):
+        print("The field {field_type_entry} is not a valid field type".format(field_type_entry=field_type_entry))
+        print "Valid field types are A, AAAA, CNAME, DKIM, LOC, MX, NAPTR, NS, PTR, SPF, SRV, SSHFP, TXT"
```
ayyazdaniaryan: You can do 

```
print(yellow("Valid field types are %s" % (", ".join(FIELD_TYPE['choices'])) ))
```

DRY - don't repeat yourself

fabfile.py
```diff
@@ -165,6 +180,198 @@ def listdir(path):
...
+@task
+def create_new_dns_record_cmd_line_arguments(domain_entry, field_type_entry, target_entry):
+    client = create_ovh_client()
+    if domain_entry not in get_domain_list():
+        print("The domain {domain_entry} is not a valid domain".format(domain_entry=domain_entry))
+        print "The valid domains are:"
+        for domain in get_domain_list():
+            print domain
+    elif not check_field_type_exists(field_type_entry):
+        print("The field {field_type_entry} is not a valid field type".format(field_type_entry=field_type_entry))
```
ayyazdaniaryan: ```print(red(...))```

fabfile.py
```diff
@@ -165,6 +180,198 @@ def listdir(path):
...
+    data = get_dns_entries_from_file()
+    if check_dns_entries(data) and check_domains(data):
+        client = create_ovh_client()
+        for domain in data:
+            for entry in data[domain]:
+                client.post('/domain/zone/{domain}/record'.format(domain=domain), fieldType=entry["type"], target=entry["value"])
+
+
+@task
+def create_new_dns_record_cmd_line_arguments(domain_entry, field_type_entry, target_entry):
```
ayyazdaniaryan: Rename to `create_new_dns_record`

fabfile.py
```diff
@@ -165,6 +180,198 @@ def listdir(path):
...
+@task
+def get_domain_list():
+    client = create_ovh_client()
+    result = client.get('/domain')
+    print json.dumps(result, indent=4)
+    return result
+
+
+@task
+def create_new_dns_record():
```
ayyazdaniaryan: Rename to `generate_dns_records_from_config`

fabfile.py
```diff
@@ -165,6 +180,198 @@ def listdir(path):
...
+        print "The valid domains are:"
+        for domain in get_domain_list():
+            print domain
+    elif not check_field_type_exists(field_type_entry):
+        print("The field {field_type_entry} is not a valid field type".format(field_type_entry=field_type_entry))
+        print "Valid field types are A, AAAA, CNAME, DKIM, LOC, MX, NAPTR, NS, PTR, SPF, SRV, SSHFP, TXT"
+    else:
+        print("Creating new DNS record for the domain {domain_entry} with field type {field_type_entry} with target {target_entry}".format(
+            domain_entry=domain_entry, field_type_entry=field_type_entry, target_entry=target_entry))
+        client.post('/domain/zone/{domain_entry}/record'.format(domain_entry=domain_entry), fieldType=field_type_entry, target=target_entry)
```
ayyazdaniaryan: For `POST` operations, you should signal success or error and display the error. So, take the response of this `post`, check for success, return in `green` a success message, or in `red`, the failure message and return the exception, which would make `fabric` quit.

fabfile.py
```diff
@@ -165,6 +180,198 @@ def listdir(path):
...
+@task
+def check_domains(data):
+    """Checks if the domains you provided in the dns_entries.json are valid"""
+    domain_list = get_domain_list()
+    if all(domain in domain_list for domain in data):
+        return True
+    else:
+        for domain in data:
+            if domain not in domain_list:
+                print("The domain {domain} is not in the domain list".format(domain=domain))
```
ayyazdaniaryan: color red

fabfile.py
```diff
@@ -165,6 +180,198 @@ def listdir(path):
...
+            entries_types.append(base["type"])
+    for item in entries_types:
+        print item
+
+    if all(check_field_type_exists(item) for item in entries_types):
+        return True
+    else:
+        for item in entries_types:
+            if not check_field_type_exists(item):
+                print("The field type {item} is not valid".format(item=item))
```
ayyazdaniaryan: color red

fabfile.py
```diff
@@ -165,6 +180,198 @@ def listdir(path):
...
+    """Checks if the inputs in your dns_entries.json file are valid"""
+    return field in FIELD_TYPE['choices']
+
+
+@task
+def check_dns_entries(data):
+    """Checks if the inputs in your dns_entries.json file are valid"""
+    entries_types = []
+    for domain in data:
+        print domain
```
ayyazdaniaryan: You should tabulate these print statements in a different way.

fabfile.py
```diff
@@ -165,6 +180,198 @@ def listdir(path):
...
+        consumer_key=env.ovh_hosting__consumer_key,
+    )
+    return client
+
+
+@task
+def get_ovh_vps_snapshot():
+    client = create_ovh_client()
+    result = client.get('/vps/{vps_identifier}/snapshot'.format(vps_identifier=env.ovh_hosting__vps_identifier))
+    print json.dumps(result, indent=4)
```
ayyazdaniaryan: Tabulate this result instead of printing in JSON. The final goal is that humans aren't the only ones using the results of these tasks

fabfile.py
```diff
@@ -165,6 +180,198 @@ def listdir(path):
...
+    client = create_ovh_client()
+    result = client.put('/vps/{vps_identifier}/snapshot'.format(vps_identifier=env.ovh_hosting__vps_identifier),
+                        description='snapshot-{today_date}'.format(today_date=today_date))
+    print json.dumps(result, indent=4)
+
+
+@task
+def list_secondary_dns_names_ovh():
+    client = create_ovh_client()
+    result = client.get('/vps/{vps_identifier}/secondaryDnsDomains'.format(vps_identifier=env.ovh_hosting__vps_identifier))
```
ayyazdaniaryan: For all of the `get` operations, if there is a failure, you should return the exception.

PR #365, opened 2018-03-14
Fixing bug on mobile:devas-1281. The error was: cannot get clientHeight of null

mohamedzayed92:


vincentfretin:
I think you can remove the code completly no? The timeline doesn't take height space anymore, it opens up on hover and overlap the content area.

PR #369, opened 2018-03-20
FormControlWithLabel update

MilanBarande:
A quick upgrade of the FormControlWithLabel component so that it adds a "*" next to the label/placeholder when the prop "required" is true.
Also added flow and cleaned the various files where we had manually added a "*" to a label


assembl/static2/js/app/components/common/formControlWithLabel.jsx
```diff
@@ -7,25 +8,49 @@
 import React from 'react';
 import { ControlLabel, FormGroup, FormControl, HelpBlock } from 'react-bootstrap';
 import { I18n } from 'react-redux-i18n';
+import { type rawContentState } from 'draft-js';
```
cedricmessiant: the type name is `RawContentState` I think. See `pageForm.jsx` for example

assembl/static2/js/app/components/common/formControlWithLabel.jsx
```diff
@@ -7,25 +8,49 @@
...
+  validationErrors?: Array<ErrorDef>
+};
+
+type FormControlWithLabelState = {
+  errorMessage?: string,
+  validationErrors?: Array<ErrorDef>,
+  validationState: ?string
+};
+
+class FormControlWithLabel extends React.Component<*, FormControlWithLabelProps, FormControlWithLabelState> {
```
cedricmessiant: you should add type for default props instead of `*` (it should be trivial)

assembl/static2/js/app/components/common/formControlWithLabel.jsx
```diff
@@ -7,25 +8,49 @@
...
+  componentWillReceiveProps(nextProps: FormControlWithLabelProps) {
     if (nextProps.validationErrors !== this.props.validationErrors) {
       this.setState(this.getStateFromProps(nextProps));
     }
   }
 
-  getStateFromProps = (props) => {
-    const errors = props.validationErrors;
+  getStateFromProps = ({ validationErrors }: FormControlWithLabelProps) => {
+    const errors = validationErrors && validationErrors;
```
cedricmessiant: why ?! It means that errors may be `false`

assembl/static2/js/app/components/common/formControlWithLabel.jsx
```diff
@@ -36,24 +61,31 @@ class FormControlWithLabel extends React.Component {
...
       validationState = 'error';
     }
 
     this.setState({ errorMessage: errorMessage, validationState: validationState });
-  }
+  };
+
+  getLabel = () => {
+    const { label, required } = this.props;
+    const placeHolder = required ? `${label}*` : label;
```
cedricmessiant: you should probably remove the `const` and just return here

assembl/static2/js/app/components/common/formControlWithLabel.jsx
MilanBarande: Not false but falsy, it can be undefined or null I believe. There is a flow error when you don't do this. I think they have a github issue regarding that

assembl/static2/js/app/components/common/formControlWithLabel.jsx
MilanBarande: ```Return assignment should not contain assignment``` or did I get you wrong?

assembl/static2/js/app/components/common/formControlWithLabel.jsx
cedricmessiant: ok so you should at least add a comment (and probably a link to the github issue). Can you tell me what is the flow error if you remove it please?

assembl/static2/js/app/components/common/formControlWithLabel.jsx
cedricmessiant: `return required ? `${label}*` : label;` should be ok, no?

assembl/static2/js/app/components/common/formControlWithLabel.jsx
```diff
@@ -33,11 +33,19 @@ type FormControlWithLabelState = {
   validationState: ?string
 };
 
-class FormControlWithLabel extends React.Component<*, FormControlWithLabelProps, FormControlWithLabelState> {
+class FormControlWithLabel extends React.Component<Object, FormControlWithLabelProps, FormControlWithLabelState> {
```
cedricmessiant: why do you use `Object`? you know exactly the form of defaultProps, right?

assembl/static2/js/app/components/common/formControlWithLabel.jsx
MilanBarande: My bad, it is not the case here. And this const errors = validationErrors is actually useless, I'll remove it

PR #371, opened 2018-03-21
Fix create proposals

cedricmessiant:
Fixes DEVAS-1339

### DIFF
--- a/assembl/graphql/vote_session.py
+++ b/assembl/graphql/vote_session.py
@@ -137,8 +137,14 @@ class UpdateVoteSession(graphene.Mutation):
             ))
 
         db.add(vote_session)
-        db.flush()
 
+        # create the root thematic on which we will attach all proposals for this vote session
+        identifier = 'voteSession{}'.format(vote_session.id)
+        root_thematic = get_root_thematic_for_phase(discussion, identifier)
+        if root_thematic is None:
+            root_thematic = create_root_thematic(discussion, identifier)
+
+        db.flush()
         return UpdateVoteSession(vote_session=vote_session)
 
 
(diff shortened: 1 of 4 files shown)


assembl/tests/fixtures/vote_session.py
```diff
@@ -29,13 +29,24 @@ def vote_session(request, test_session, discussion, timeline_vote_session,
...
+    identifier = 'voteSession{}'.format(vote_session.id)
+    short_title = u'Phase {}'.format(identifier)
+    root_thematic = Thematic(
+        discussion_id=discussion.id,
+        title=LangString.create(short_title, "en"),
+        identifier=identifier,
+        hidden=True)
+    discussion.root_idea.children.append(root_thematic)
+    test_session.flush()
+
```
vincentfretin: Why don't you use the create_root_thematic function here?

assembl/graphql/vote_session.py
```diff
@@ -775,7 +781,8 @@ def mutate(root, args, context, info):
             identifier = 'voteSession{}'.format(vote_session_id)
             root_thematic = get_root_thematic_for_phase(discussion, identifier)
             if root_thematic is None:
-                root_thematic = create_root_thematic(discussion, identifier)
+                raise Exception(
+                    "There is no root thematic for this vote session.")
```
vincentfretin: You will have an issue on prod if a vote session was already created and no proposals created yet. You should maybe create an alembic step to create the root thematic for existing vote session?

assembl/tests/fixtures/vote_session.py
cedricmessiant: there was a circular import issue

assembl/graphql/vote_session.py
cedricmessiant: No vote session has been created in prod yet. I will check with Fred if this is an issue

assembl/tests/fixtures/vote_session.py
vincentfretin: Even if you put the import here locally?

cedricmessiant: it's good, thanks

PR #372, opened 2018-03-21
Upgrading preprod server documentation

mohamedzayed92:


doc/preprod_update.rst
```diff
@@ -0,0 +1,74 @@
...
+
+
+This brings the entire dependency up to latest.
+However, for ngnix, you have to copy/paste the static folders to a volumne that can be shared with other containers.
+
+Nginx container::
+
+	cd /opt/assembl_static
+	./update.sh
+	rm - rf static_backup
```
swergas: remove space

doc/preprod_update.rst
```diff
@@ -0,0 +1,74 @@
+Updating the preprod Server
+===========================
+
+This method is a temporary method in order to quickly update the preprod server with the latest version of the code.
+
+Preprod server has an assembl instance running on different containers. The whole instance is run via a .yml file run with docker-compose
+
+
+Note: 6f41608ee826 is the identity of the container running the image assembl:develop
```
ayyazdaniaryan: I wouldn't do this. The next time the image is rebuilt, the docs will have to change to represent this new container ID. I would only describe that the container id can be found using `docker ps | grep assembl`

doc/preprod_update.rst
ayyazdaniaryan: Yup, no space after dash. `rm -rf static_backup`

doc/preprod_update.rst
```diff
@@ -0,0 +1,74 @@
...
+This brings the entire dependency up to latest.
+However, for ngnix, you have to copy/paste the static folders to a volumne that can be shared with other containers.
+
+Nginx container::
+
+	cd /opt/assembl_static
+	./update.sh
+	rm - rf static_backup
+	mv static static_backup2
+	mv static2 static_backup2
```
ayyazdaniaryan: I think now the shell script manages this as well. So you can remove this line and the one above.

assembl/static2/js/app/components/administration/exportSection.jsx
```diff
@@ -0,0 +1,108 @@
+// @flow
```
cedricmessiant: why this file is in this commit ? Maybe a bad merge ?

PR #374, opened 2018-03-22
Token categories separator fix

MilanBarande:
[Jira Ticket](https://bluenove.atlassian.net/browse/DEVAS-1337)

### DIFF
--- a/assembl/static2/css/views/administration.scss
+++ b/assembl/static2/css/views/administration.scss
@@ -16,6 +16,10 @@
     position: relative;
   }
 
+  .no-box-shadow {
+    box-shadow: none !important;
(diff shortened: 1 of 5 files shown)


assembl/static2/css/views/administration.scss
```diff
@@ -16,6 +16,10 @@
     position: relative;
   }
 
+  .no-box-shadow {
+    box-shadow: none !important;
```
paolina92: If possible, avoid to use `!important`

MilanBarande: Doesn't work without it. It's to override the style of the component which we import from react-color

PR #375, opened 2018-03-22
Refactor styles and grid of the menu sticky

paolina92:


assembl/static2/js/app/components/voteSession/availableTokens.jsx
```diff
@@ -14,33 +15,98 @@ type Props = {
...
+  case 1:
+    return ['left'];
+  case 2:
+    return ['left', 'right'];
+  case 3:
+    return ['left', '', 'right'];
+  case 4:
+    return ['left', 'right', 'left', 'right'];
+  default:
+    return [12];
```
ayyazdaniaryan: Huh? Returning an `Array<Int>` instead of `Array<string>`? I'm not a fan of functions that return different types. But, I understand you're trying to yield the default 12 grids on a large screen.

PR #377, opened 2018-03-22
vote results page when vote session is closed

vincentfretin:


assembl/static2/js/app/components/voteSession/proposalResults.jsx
```diff
@@ -0,0 +1,84 @@
...
+
+type Props = {
+  description: ?string,
+  id: string,
+  modules: Array<VoteSpecification>,
+  numParticipants: number,
+  title: ?string
+};
+
+class ProposalResults extends React.Component<void, Props, void> {
```
MilanBarande: Why is a class based component needed here ?

vincentfretin: Not really needed. I copied the component from proposal.jsx and removed the state. This could be transformed to a functional component indeed.

PR #378, opened 2018-03-22
Use HashFS to store files instead of database.

maparent:
Antivirus and backup modified. Docker TBD.

### DIFF
--- a/assembl/tests/views_tests/test_browser.py
+++ b/assembl/tests/views_tests/test_browser.py
@@ -23,7 +23,7 @@ def test_mocha(browser, test_server, discussion, test_session,
     # print browser.driver.get_log('browser')
     assert not extractor.failures_number
 
-
+@pytest.mark.skip(reason="test_app_no_login_real_policy has security side-effect")
 @pytest.mark.xfail
 def test_private_discussion_log_in_form_exists_and_works(test_server_no_login_real_policy, browser, test_session, discussion, participant1_user, test_webrequest):
     url = test_server_no_login_real_policy.url + test_webrequest.route_path(


vincentfretin:
Looks good.

assembl/lib/hash_fs.py
```diff
@@ -0,0 +1,17 @@
...
+
+_hashfs = None
+
+
+def get_hashfs():
+    global _hashfs
+    if _hashfs is None:
+        root = get('upload_root', 'var/uploads')
+        if root[0] != '/':
+            root = path.abspath(path.join(__file__, '../../..', root))
```
ayyazdaniaryan: is it not one `...` too many?

assembl/models/attachment.py
```diff
@@ -177,12 +180,59 @@ def __init__(self, *args, **kwargs):
...
+    def path_of(cls, id):
+        return cls.hashfs.get(id).abspath.encode('ascii')
+
+    @property
+    def path(self):
+        return self.path_of(self.file_identity)
+
+    @property
+    def handoff_url(self):
+        return b'/private_uploads' + self.path[len(self.hashfs.root):]
```
ayyazdaniaryan: use `path_join` here perhaps.

assembl/scripts/files_gc.py
```diff
@@ -0,0 +1,35 @@
...
+    db = boostrap_configuration(args.configuration)
+    from assembl.models import File
+    hashfs = File.hashfs
+    ids = {x for (x,) in db.query(File.file_identity)}
+    for root, dirs, files in os.walk(hashfs.root):
+        base = ''.join(root[len(hashfs.root):].split('/'))
+        for fname in files:
+            id = base + fname.split('.')[0]
+            if id not in ids:
+                print("Unknown file: ", "/".join([root]+dirs+[fname]))
```
ayyazdaniaryan: log +/- raven instead of print

assembl/views/api2/attachments.py
```diff
@@ -44,17 +76,42 @@ def get_file(request):
...
+            app_iter = environ['wsgi.file_wrapper'](fs, _BLOCK_SIZE)
+        if app_iter is None:
+            app_iter = FileIter(fs, _BLOCK_SIZE)
+        kwargs=dict(app_iter=app_iter)
+
+    r = Response(
+        content_length=f.size,
         content_type=str(f.mime_type),
+        last_modified=f.creation_date,
+        expires=datetime.now()+timedelta(days=365),
```
ayyazdaniaryan: This might become an issue in case of deletes/moderation/etc, no?

assembl/views/api2/attachments.py
```diff
@@ -44,17 +76,42 @@ def get_file(request):
...
+        accept_ranges="bytes" if handoff_to_nginx else "none",
         content_disposition=
             'attachment; filename="%s"; filename*=utf-8\'\'%s' # RFC 6266
-            % (escaped_double_quotes_filename, url_quoted_utf8_filename)
+            % (escaped_double_quotes_filename, url_quoted_utf8_filename),
+        **kwargs
     )
+    if handoff_to_nginx:
+        r.headers[b'X-Accel-Redirect'] = f.handoff_url
```
ayyazdaniaryan: why add the `b` before the header?

assembl/lib/hash_fs.py
maparent: Ah, may work better with a dirname first.

assembl/models/attachment.py
maparent: no / between components

assembl/views/api2/attachments.py
maparent: yes, does not play well with deletes, but it's what gives us caching. The user would have to know the file ID...
That's a judgement call. I noticed that you never edit files, but replace them, so we don't worry about that. I am going this way, you may go another.

assembl/views/api2/attachments.py
maparent: python-3 compatible, fwiw.

assembl/scripts/files_gc.py
maparent: Not sure, it's a script, so the user would be running it and looking at output.

PR #379, opened 2018-03-23
Share a thematic

paolina92:
Finally, we don’t want to have like and disagree buttons. Just the possibility to share a thematic.


MilanBarande:
LGTM!

assembl/static2/js/app/components/debate/common/headerActions.jsx
```diff
@@ -0,0 +1,46 @@
...
+        </div>
+        <div className="action-button-label">
+          <Translate value="debate.share" />
+        </div>
+      </div>
+    </div>
+  );
+};
+
+const mapStateToProps = ({ debate }) => {
```
cedricmessiant: since you already have `debate` in the `Header` component, you should probably just pass `useSocialMedia` to `HeaderActions` and remove the redux connect

assembl/static2/js/app/components/svg/disagree.jsx
```diff
@@ -26,8 +26,8 @@ class Disagree extends React.Component {
             1.39072848 14,
             1.39072848 L14,
             1.39072848 Z"
-            stroke="#F75959"
-            fill="#FFFFFF"
+            stroke={color || '#F75959'}
+            fill={backgroundColor || '#ffffff'}
```
cedricmessiant: you should probably use default props for all these (same for `dontUnderstand.jsx` and `like.jsx` and `moreInfo.jsx`)

assembl/static2/js/app/components/debate/common/headerActions.jsx
```diff
@@ -0,0 +1,46 @@
+// @flow
+import React from 'react';
+import { connect } from 'react-redux';
+import { Translate } from 'react-redux-i18n';
+import { openShareModal } from '../../../utils/utilityManager';
+
+type Props = {
+  routerParams: Map<string>,
```
cedricmessiant: I think I added a flow type for router params (in `flow/object_types.js`). Also note that here you use `Map` as a type but you don't import it from immutable so your type is not right.

PR #380, opened 2018-03-26
Add Cancel module customization for custom gauges in step 3

cedricmessiant:
If the gauge is custom, display a "Cancel settings" link that replaces customization by the template data
Closes DEVAS-1367

### DIFF
--- a/assembl/static2/tests/unit/reducers/adminReducer/voteSession.spec.js
+++ b/assembl/static2/tests/unit/reducers/adminReducer/voteSession.spec.js
@@ -282,6 +282,125 @@ describe('voteSession admin reducers', () => {
     });
   });
 
+  describe('voteProposalsHaveChanged reducer', () => {
+    const { voteProposalsHaveChanged } = reducers;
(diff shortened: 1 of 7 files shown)


assembl/static2/tests/unit/reducers/adminReducer/voteSession.spec.js
```diff
@@ -1155,5 +1274,264 @@ describe('voteSession admin reducers', () => {
...
+              localeCode: 'en',
+              value: 'My custom instructions'
+            }
+          ],
+          choices: ['custom-choice1', 'custom-choice2']
+        },
+        template: {
+          _hasChanged: false,
+          _isNew: false,
+          _toDelete: true,
```
vincentfretin: `_toDelete: true` is this normal?

cedricmessiant: yes, whatever. We just have to ensure that the template data remains unchanged.

cedricmessiant:
about the change of the gauge type, I don't know. For now, it is still hidden (there is a task about it in the backlog)

PR #382, opened 2018-03-27
fix navbar activelink bug

MilanBarande:


assembl/static2/js/app/components/navbar/navbar.jsx
```diff
@@ -63,19 +63,21 @@ const SectionLink = ({ section, options }) => {
       </a>
     );
   }
+  const sectionName = sectionType.toLowerCase().replace('_', '');
+  const isActiveUrl = location.href.includes(sectionName);
```
vincentfretin: Be careful, this will match if sectionName is found in the domain or in the discussion slug or in somemorecharacter-sectionName in the path. You need to use a more robust solution.

   const isActiveUrl = location.pathname.split('/').slice(2).join('/').indexOf(`/${sectionName}`) > -1

This will first transform "/debate-syntheses/debate/thread" to "debate/thread" then search "/thread" in this string.

MilanBarande: Okay perfect, this is indeed avoiding the potential false positives. But indexOf always returns -1 because it's not an array but a string. Using your transformation with includes(sectionName) works fine though

PR #384, opened 2018-03-28
Metrics on thematic devas 1280

vincentfretin:


assembl/tests/views_tests/test_graphql.py
```diff
@@ -537,12 +538,19 @@ def test_get_question_via_node_query(graphql_request, thematic_and_question):
...
         }
     }""" % first_question_id, context_value=graphql_request)
     assert json.loads(json.dumps(res.data)) == {
-        u'node': {u"__typename": u"Question",
-                  u"title": u"Comment qualifiez-vous l'emergence de l'Intelligence Artificielle dans notre société ?"}}
+        u'node': {
+            u"__typename": u"Question",
+            u"numPosts": 0,
+            u"numContributors": 0,
+            u"totalSentiments": 0,
```
cedricmessiant: shouldn't we test cases with sentiments, contributors, posts, etc ?

assembl/static2/js/app/components/common/headerStatistics.jsx
```diff
@@ -0,0 +1,66 @@
...
+  metricNameTranslateKey: 'home.contribution'
+});
+
+export const statMessages = (numPosts: number) => ({
+  iconName: 'message',
+  metricValue: numPosts,
+  metricNameTranslateKey: 'home.messages'
+});
+
+export const statParticipants = (numParticipants: number) => ({
```
cedricmessiant: type return values for all these functions ?

assembl/tests/views_tests/test_graphql.py
vincentfretin: I added assertions to existing test_mutation_add_sentiment tests

PR #390, opened 2018-04-04
Vote report and graphql IDs in APIv2

maparent:


assembl/views/api2/votes.py
```diff
@@ -196,3 +200,56 @@ def vote_results_csv(request):
...
+@view_config(context=InstanceContext, request_method='GET',
+             ctx_instance_class=VotingWidget,
+             name="vote_results_csv", permission=P_DISC_STATS)
+def global_vote_results_csv(request):
+    ctx = request.context
+    user_id = request.authenticated_userid
+    if not user_id:
+        raise HTTPUnauthorized
+    widget = ctx._instance
+    if widget.activity_state != "ended":
```
ayyazdaniaryan: The activity state isn't an `enum` already?

assembl/views/api2/votes.py
```diff
@@ -196,3 +200,56 @@ def vote_results_csv(request):
...
+        if P_ADMIN_DISC not in permissions:
+            raise HTTPUnauthorized()
+    user_prefs = LanguagePreferenceCollection.getCurrent()
+    ideas = widget.db.query(Idea
+        ).join(AbstractIdeaVote, AbstractIdeaVote.idea_id==Idea.id
+        ).join(AbstractVoteSpecification
+        ).filter(AbstractVoteSpecification.widget_id==widget.id
+        ).distinct().all()
+    idea_ids = [i.id for i in ideas]
+    titles = [(idea.title.best_lang(user_prefs).value, idea.id) for idea in ideas]
```
ayyazdaniaryan: Not defensive enough. `title` can be `None`.

assembl/views/traversal.py
```diff
@@ -204,14 +204,7 @@ def __init__(self, parent, cls):
...
         from assembl.models import NamedClassMixin, Preferences
-        num_key = None
-        instance = None
-        try:
-            num_key = int(key)
-        except ValueError:
-            pass
-        if num_key is not None:
-            instance = self._class.get_instance(num_key)
+        instance = instance = self._class.get_instance(key)
```
ayyazdaniaryan: Is there a bad copy/pasta here with the `instance`?

assembl/views/api2/votes.py
```diff
@@ -196,3 +200,56 @@ def vote_results_csv(request):
...
+    q = q.group_by(Idea.id)
+    r = q.all()
+    r = {x[0]: x for x in r}
+    output = StringIO()
+    csvw = csv.writer(output)
+    csvw.writerow([""]+[t.encode('utf-8') for (t, spec) in specs])
+    for title, idea_id in titles:
+        row = [title.encode('utf-8')]
+        sourcerow = r[idea_id][1:]
+        for i, (t, spec) in enumerate(specs):
```
ayyazdaniaryan: Can we get a small comment on the counting? Reading what `r` will produce and following the algo along without running it is non-trivial.

assembl/lib/sqla.py
```diff
@@ -528,9 +529,13 @@ def get_instance(cls, identifier, session=None):
...
-            if not uri.startswith('local:') or '/' not in uri:
-                return
-            uriclsname, num = uri[6:].split('/', 1)
+            if uri.startswith('local:') and '/' in uri:
+                uriclsname, num = uri[6:].split('/', 1)
+            else:
+                try:
+                    uriclsname, num = Node.from_global_id(uri)
+                except Exception as e:
+                    return
```
ayyazdaniaryan: You should log the exception to raven, then return `None`. Will there be defensive logic downstream for a potential `None` db id?

assembl/views/api2/votes.py
maparent: nope. good idea, but in scope?

assembl/lib/sqla.py
maparent: This is used all over the place, and code downstream is supposed to be defensive.
As for logging... if we knew it was either a local: type ID or a GraphQL idea or... then it would make sense to log here, but the point of this function is that it checks if your ID makes sense at all in any known scheme. The failure of one scheme (even the last scheme in line) is not information. (also, maybe there's another scheme to be tried in an overridden function, I'm thinking about it...) SO I say logging should be downstream as well. What I can do is add a comment about it.

assembl/views/api2/votes.py
maparent: You got to be kidding me... That should almost be made non-null in the DB.

maparent: But I did do a defensive version.

PR #392, opened 2018-04-05
Extract vote data page

MilanBarande:
There is still an issue with the snapshot test but the rest is reviewable

### DIFF
--- a/assembl/static2/js/app/components/administration/exportSection.jsx
+++ b/assembl/static2/js/app/components/administration/exportSection.jsx
@@ -15,15 +15,8 @@ type Props = {
   translate?: boolean
 };
 
-type State = {
-  exportLocale: ?string,
-  translate: boolean
-};
-
-class ExportSection extends React.Component<Object, Props, State> {
+class ExportSection extends React.Component<Object, Props, void> {
   props: Props;
-
-  state: State;
 
   static defaultProps = {
     languages: null,


assembl/static2/js/app/components/administration/exportSection.jsx
```diff
@@ -0,0 +1,161 @@
+// @flow
```
paolina92: It should be a new section in admininistration > voteSession > export instead of administration > export. No?

assembl/static2/js/app/components/administration/exportSection.jsx
```diff
@@ -0,0 +1,161 @@
...
+    if (isSurveyExport) {
+      exportLink = `/data/Discussion/${debateId}/phase1_csv_export${translate && exportLocale ? `?lang=${exportLocale}` : ''}`;
+    } else if (!isSurveyExport && voteSessionId) {
+      exportLink = `/data/Discussion/${debateId}/widgets/${voteSessionId}/vote_results_csv`;
+    }
+    return (
+      <div className="admin-box survey-admin-export-section">
+        <SectionTitle title={I18n.t('administration.survey.2')} annotation={I18n.t('administration.surveyExport.annotation')} />
+        <div className="admin-content">
+          {isSurveyExport && (
```
paolina92: ha ok. So we have just one section at the high level of the menu with cases depending on the data that we want to export

assembl/static2/js/app/components/administration/exportSection.jsx
```diff
@@ -0,0 +1,161 @@
...
+  render() {
+    const { languages, exportType, voteSessionId } = this.props;
+    const isSurveyExport = exportType === 'survey';
+    const { translate } = this.state;
+    const debateId = getDiscussionId();
+    if (!debateId) return null;
+    const exportLocale = this.state.exportLocale || (languages && languages[0].locale);
+    let exportLink;
+    if (isSurveyExport) {
+      exportLink = `/data/Discussion/${debateId}/phase1_csv_export${translate && exportLocale ? `?lang=${exportLocale}` : ''}`;
```
paolina92: Add it to `routes.json`

assembl/static2/js/app/components/administration/exportSection.jsx
```diff
@@ -0,0 +1,161 @@
...
+    } else if (!isSurveyExport && voteSessionId) {
+      exportLink = `/data/Discussion/${debateId}/widgets/${voteSessionId}/vote_results_csv`;
```
paolina92: Same here

assembl/static2/js/app/components/administration/exportSection.jsx
cedricmessiant: I agree

assembl/static2/js/app/components/administration/exportSection.jsx
```diff
@@ -0,0 +1,161 @@
...
+  props: Props;
+
+  state: State;
+
+  static defaultProps = {
+    languages: null,
+    voteSessionId: null
+  };
+
+  static ExportLanguageDropDown = ({ languages, onSelect, activeKey }) => {
```
cedricmessiant: I thought there is no language select on this page but maybe I misunderstood (maybe check with Fred). If it stays, I would put this in a component in `common/` and use it in both export sections. I guess you should put all the translation mechanism in a component also (the radio buttons and state management).

assembl/static2/js/app/components/administration/exportSection.jsx
MilanBarande: Why having two components doing the same thing ? This component as in adminstration/survey and I moved it one folder up and use it for both survey and voteSession

assembl/static2/js/app/components/administration/exportSection.jsx
```diff
@@ -0,0 +1,161 @@
...
+  exportType: string,
+  voteSessionId?: string
+};
+
+type State = {
+  exportLocale: ?string,
+  translate: boolean
+};
+
+class DumbExportSection extends React.Component<Object, Props, State> {
```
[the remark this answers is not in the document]
MilanBarande:
```type Props = {
  languages?: Array<Object>,
  exportType: string,
  voteSessionId?: string
};```
already implies that languages and voteSessionId might be null, doesn't it ?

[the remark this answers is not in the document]
Comment (cedricmessiant) [reply assembl/static2/js/app/components/administration/exportSection.jsx]: yes, it might but the default prop is `null`

File: assembl/static2/js/app/components/administration/exportSection.jsx (hunk 1 above)
Comment (cedricmessiant) [reply assembl/static2/js/app/components/administration/exportSection.jsx]: yes, I wrote this comment before I saw that it was in fact a refactoring

File: assembl/static2/js/app/components/administration/exportSection.jsx [hunk 7]
```diff
@@ -1,35 +1,35 @@
...
-  voteSessionId?: string
+  handleTranslationChange: Function,
+  handleExportLocaleChange: Function,
+  withLanguageOptions?: boolean,
+  exportLink: string,
+  exportLocale?: string,
+  translate?: boolean
 };
 
 type State = {
```
cedricmessiant: you don't have a state anymore, right ?

assembl/static2/js/app/pages/surveyAdmin.jsx
```diff
@@ -64,16 +74,38 @@ class SurveyAdmin extends React.Component {
...
+  handleTranslationChange = (shouldTranslate) => {
+    this.setState({ translate: shouldTranslate });
+  };
+
   render() {
-    const { section, thematicsHaveChanged } = this.props;
+    const { section, thematicsHaveChanged, debateId, languages } = this.props;
+    const { translate } = this.state;
+    const exportLocale = this.state.exportLocale || (languages && languages[0].locale);
+    const translation = translate && exportLocale ? `?lang=${exportLocale}` : '';
```
cedricmessiant: do you really need translate here ? I would have put it in the local state of `exportSection` component and use it only to know if I have to display the options or not. But it requires to set export locale to `''` when the radio is changed to "don't translate". I guess it's a matter of personal preference here

assembl/static2/js/app/pages/surveyAdmin.jsx
```diff
@@ -82,9 +114,11 @@ class SurveyAdmin extends React.Component {
   }
 }
 
-const mapStateToProps = ({ admin: { thematicsById, thematicsHaveChanged, thematicsInOrder } }) => ({
+const mapStateToProps = ({ admin: { thematicsById, thematicsHaveChanged, thematicsInOrder }, context, i18n }) => ({
   thematicsHaveChanged: thematicsHaveChanged,
-  thematics: thematicsInOrder.toArray().map(id => thematicsById.get(id).toJS())
+  thematics: thematicsInOrder.toArray().map(id => thematicsById.get(id).toJS()),
+  debateId: context.debateId,
```
cedricmessiant: cool !

assembl/static2/js/app/pages/surveyAdmin.jsx
MilanBarande: Ok so if I get you right we still want the link to be built in the parent of exportSection but exportSection is in charge of telling if it wants translation and send it up to its parent... The problem in this case is that exportSection will have to do a graphql Query and be connected to the store despite not always needing it.

PR #394, opened 2018-04-05
change style in survey view

felixsaintgenis:


assembl/static2/css/components/text.scss
```diff
@@ -59,8 +59,9 @@
 .dark-title-5 {
   color: $title5-dark-color;
   font-family: $font-family4;
-  font-size: $title5-size;
+  font-size: $title5-size + 2px;
```
MilanBarande: There is no $font-size variable to get the same result ?

felixsaintgenis: nope. Should I create a new one or it's better keeping the css this way ?

paolina92: You should create a new class

PR #395, opened 2018-04-09
check if inputs are empty to display a confirmModal or not

MilanBarande:
I just transformed voteProposalForm in a class based component to be able to use the local state in order to keep track of the inputs' content. If the fields are empty we don't ask to confirm deletion with a modal

### DIFF
--- a/assembl/static2/js/app/components/administration/voteSession/voteProposalForm.jsx
+++ b/assembl/static2/js/app/components/administration/voteSession/voteProposalForm.jsx
@@ -46,42 +46,49 @@ type VoteProposalFormProps = {
   validationErrors: ValidationErrors
 };
 
+type VoteProposalFormState = {
+  titleContent: string,
+  descriptionContent: string
+};
+
 export const getValidationState = (validationErrors: ?Array<ErrorDef>): ?string =>
   (validationErrors && validationErrors.length > 0 ? 'error' : null);
 
-const DumbVoteProposalForm = ({
-  index,
-  title,
-  description,
-  _toDelete,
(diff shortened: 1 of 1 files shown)


vincentfretin:
I'll leave it to @cedricmessiant to review this. :)

cedricmessiant:
Why do you need a local state to keep track of the inputs ? you have the props for this, right ?

assembl/static2/js/app/components/administration/voteSession/voteProposalForm.jsx
```diff
@@ -116,6 +116,10 @@ const DumbVoteProposalForm = ({
     displayCustomModal(content, true, 'gauge-settings-modal');
   };
 
+  const isTitleEmpty = title === '' || title === null;
+  const isDescriptionEmpty = description === null || description.blocks.every(block => block.text === '');
```
cedricmessiant: title & description may not be undefined ?

cedricmessiant: you should probably use `rawContentStateIsEmpty` (in `utils/draftjs.js`) that will work if there is an atomic block in your raw content state

MilanBarande: for title & description they are set to null before anything is typed in, do you see a case where they would turn to undefined ? 
I'll check rawContentStateIsEmpty ;)

cedricmessiant: no, it should be ok

PR #399, opened 2018-04-11
Add cookie support v2

ayyazdaniaryan:
More tests to be added and front-end updates. However, a review is needed at this point.

### DIFF
--- a/assembl/models/auth.py
+++ b/assembl/models/auth.py
@@ -1811,8 +1811,7 @@ class UserLanguagePreferenceCollection(LanguagePreferenceCollection):
     def recalculate_locale(self, db):
         self.calculate_locale_prefs(db)
 
-    # The real add_locale
-    def process_locale(self, locale_code, source_of_evidence):
+    def process_locale(self, locale_code, source_of_evidence, **kwargs):
         session = self.user.db
         locale_code = to_posix_string(locale_code)
         # Updated: Now Locale is a model. Converting posix_string into its
@@ -1830,7 +1829,7 @@ class UserLanguagePreferenceCollection(LanguagePreferenceCollection):
             if len(lang_pref_signatures[source_of_evidence]) == 1:
                 lang_pref_signatures[source_of_evidence][0].locale = locale
                 session.flush()
-                return
+                return lang_pref_signatures[source_of_evidence][0]
             # else creation below
         else:
             lang_pref_signatures = {
@@ -1838,12 +1837,29 @@ class UserLanguagePreferenceCollection(LanguagePreferenceCollection):
                 for lp in self.user.language_preference
             }
             if (locale.id, source_of_evidence) in lang_pref_signatures:
-                return
+                return filter(
+                    lambda x: x.locale_id == locale.id and x.source_of_evidence == source_of_evidence, self.user.language_preference)[0]
         lang = UserLanguagePreference(
             user=self.user, source_of_evidence=source_of_evidence.value, locale=locale)
         session.add(lang)
         session.flush()
         self.calculate_locale_prefs(session)
+        return lang
+
+    def process_post_locale(self, post_id, **kwargs):
+        if post_id in self.post_prefs:
+            pref = self.post_prefs[post_id]
+            for k, v in kwargs.iteritems():
+                setattr(pref, k, v)
+        else:
+            pref = PostUserLanguagePreference(
+                post_id=post_id,
+                user=self.user,
+                **kwargs
+            )
+            self.user.db.add(pref)
+            self.user.db.flush()
+        return pref
 
     def calculate_locale_prefs(self, db):
         # create another dictionary for posts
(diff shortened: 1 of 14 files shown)


assembl/models/auth.py
```diff
@@ -1812,6 +1812,10 @@ def recalculate_locale(self, db):
         self.calculate_locale_prefs(db)
 
     def process_locale(self, locale_code, source_of_evidence, **kwargs):
+        """
+        How the locales are procssed for a user language preference. This method SHOULD
```
vincentfretin: procssed => processed

assembl/models/auth.py
```diff
@@ -1752,7 +1752,8 @@ def getCurrent(cls, req=None, session=None):
             user_id = req.authenticated_userid
             if user_id and user_id != Everyone:
                 try:
-                    discussion_id = req.matchdict['discussion_id'] if (req.matchdict and 'discussion_id' in req.matchdict) else None
+                    discussion_id = req.matchdict['discussion_id'] if (
+                        'matchdict' in req.__dict__ and req.matchdict and 'discussion_id' in req.matchdict) else None
```
vincentfretin: This feels hacky and complicated. matchdict really can be missing in req?
I tend to prefer `.get` and `gettattr` syntax instead of `in` like you did. This line should do the same:

    discussion_id = req.matchdict.get('discussion_id', None) if getattr(req, 'matchdict', None) is not None else None

If we always have a req.matchdict attribute, an empty dict or None, even simpler:

    discussion_id = req.matchdict.get('discussion_id', None)

assembl/models/auth.py
```diff
@@ -1768,18 +1775,19 @@ def known_languages(self):
...
-    def __init__(self, locale_code):
-        self.default_locale = Locale.get_or_create(locale_code)
+    def __init__(self, locale_code, session=None):
+        self.default_locale = Locale.get_or_create(locale_code, session)
 
     def default_locale_code(self):
         return self.default_locale
 
-    def find_locale(self, locale, db=None):
+    def find_locale(self, locale, **kwargs):
```
vincentfretin: Was it necessary to change db to kwargs here?

assembl/models/auth.py
```diff
@@ -1989,3 +2128,34 @@ def __repr__(self):
...
+                        backref=backref('user_language_preferences'))
+
+    def unique_query(self):
+        query, _ = super(PostUserLanguagePreference, self).unique_query()
+        query = query.filter_by(post_id=self.post_id)
+        return query, True
+
+    # Query a property for fetching a user's preferences on a post
+    def __repr__(self):
+        string_rep = super(PostUserLanguagePreference, self).__repr__()
```
vincentfretin: You shouldn't call `__repr__()` yourself, use the build-in `repr(super(PostUserLanguagePreference, self))`

assembl/tests/fixtures/user_language_preference.py
```diff
@@ -84,6 +85,34 @@ def fin():
...
+
+    locale_from = en_locale
+    ulp = UserLanguagePreference(
+        user=admin_user,
+        locale=locale_from,
+        preferred_order=0,
+        source_of_evidence=LanguagePreferenceOrder.Server.value)
+
+    def fin():
+        print "finalizer user_language_preference_en_cookie"
```
vincentfretin: This doesn't match the fixture name.

assembl/tests/model_tests/test_langstring.py
```diff
@@ -635,7 +635,7 @@ def test_identify_ambiguous_body_in_post_by_creator(
     assert post_body_locale_determined_by_creator.subject.\
         first_original().locale_code == 'fr'
 
-
+@pytest.mark.xfail(reason="IntegrityError on teardown on language_preference. TODO resolve!")
```
vincentfretin: Did yo utry to resolve it?

assembl/models/auth.py
ayyazdaniaryan: Ooops, bad copy/pasta. I had previously changed the interface of having `find_locale`, but it simply caused way too many issues with translation services and post creation, so I reverted my changes. Missed this!

assembl/models/auth.py
ayyazdaniaryan: Fair, are there triggers that get called this way?

assembl/tests/fixtures/user_language_preference.py
ayyazdaniaryan: 🍝

assembl/tests/model_tests/test_langstring.py
ayyazdaniaryan: No, this one I will have to look at with @maparent . It likely has to do with the global Locale's, but for the time being, marked it as `xfail`

assembl/models/auth.py
vincentfretin: yes, same for len which call __len__, str which call __str__ etc.

paolina92:
@ayyazdaniaryan What do we do with this PR?

assembl/models/auth.py
```diff
@@ -1989,3 +2128,34 @@ def __repr__(self):
                 LanguagePreferenceOrder(self.source_of_evidence).name,
                 self.preferred_order or 0
             )
+
+
+class PostUserLanguagePreference(UserLanguagePreference):
```
amensouissi: I think we need to set up a centralized system for user preferences. Currently, for the V1 (and the V2), we have models.preferences.Preferences (Json) and UserPreferenceCollection a wrapper for reading and writing data. This system is centralized, extensible and maintainable but is not sufficiently expressive (for complex needs). We also have an ad hoc system. In this case the preferences are decentralized and not easily maintainable but sufficiently expressive. I think we need to have a centralized system for user preference. For example, a  table of preferences 'Preference' (not Json but can contain a Json) to describe complex needs. In this table we add the column 'language_preference' (for your need) which references an entry in a table 'Translation' and another column 'harvesting_translation' which references another entry in the same table 'Translation'. And we can add the Cookies preferences in this 'Preference' table.

PR #400, opened 2018-04-11
DEVAS-1282 Collapse message form on input blur if both title and content input are empty

swergas:
As I said earlier: While thinking about it and implementing it I saw that if the user clicks on the input to expand then on the content field to fill this one first, then it collapses.
So I guess there is a UX problem in the expected result?
Or maybe the "when i click somewhere else" means outside the whole blue block. Then instead of the browser onblur event we should implement something more complicated, like handleClickOutside in `assembl/static2/js/app/components/debate/navigation/debateLink.jsx` or handleTouchOutside in `assembl/static2/js/app/components/debate/navigation/timelineSegment.jsx`.
I'm waiting for instructions as to keep it as is or implement the second solution.

### DIFF
--- a/assembl/static2/js/app/components/common/richTextEditor/index.jsx
+++ b/assembl/static2/js/app/components/common/richTextEditor/index.jsx
@@ -15,6 +15,7 @@ import attachmentsPlugin from './attachmentsPlugin';
 type RichTextEditorProps = {
   rawContentState: RawContentState,
   handleInputFocus: Function,
+  handleInputBlur: Function,
   maxLength: number,
   placeholder: string,
   textareaRef: Function,
@@ -51,6 +52,7 @@ export default class RichTextEditor extends React.Component<Object, RichTextEdit
 
   static defaultProps = {
     handleInputFocus: null,
+    handleInputBlur: null,
     maxLength: 0,
     toolbarPosition: 'top',
     withAttachmentButton: false,
@@ -83,6 +85,7 @@ export default class RichTextEditor extends React.Component<Object, RichTextEdit
       },
       () => {
         this.props.updateContentState(this.getCurrentRawContentState());
+        this.props.handleInputBlur();
       }
     );
   };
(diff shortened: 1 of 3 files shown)


vincentfretin:
The title is mandatory anyway, she should fill in first, it's the first focused field. If the user click on title input, then on content, it will collapse and she has to click again on the input and fill the title input. For me it's okay, but I'm not the one to decide.

assembl/static2/js/app/components/common/richTextEditor/index.jsx
```diff
@@ -51,6 +52,7 @@ export default class RichTextEditor extends React.Component<Object, RichTextEdit
 
   static defaultProps = {
     handleInputFocus: null,
+    handleInputBlur: null,
```
vincentfretin: you should use `() => {}` here instead of null

assembl/static2/js/app/components/common/richTextEditor/index.jsx
```diff
@@ -83,6 +85,7 @@ export default class RichTextEditor extends React.Component<Object, RichTextEdit
       },
       () => {
         this.props.updateContentState(this.getCurrentRawContentState());
+        this.props.handleInputBlur();
```
vincentfretin: This will crash if handleInputBlur is null, see my comment above.

assembl/static2/js/app/components/common/textInputWithRemainingChars.jsx
```diff
@@ -21,6 +22,7 @@ export const TextInputWithRemainingChars = ({
         maxLength={maxLength}
         value={value}
         onFocus={handleInputFocus || null}
+        onBlur={handleInputBlur || null}
```
vincentfretin: We should'nt have those `||null` for onFocus and onBlur but defined a defaultPropos `()=>{}` in the component.

assembl/static2/js/app/components/common/richTextEditor/index.jsx
swergas: Thanks for spotting this :) As of the solution I would prefer adding an if this function exists

assembl/static2/js/app/components/common/textInputWithRemainingChars.jsx
swergas: I would prefer adding the onBlur (respectively onFocus) attribute only if handleInputBlur is a defined function

vincentfretin: As you wish, fine by me.

PR #402, opened 2018-04-12
Updating documentation for production instance installation

mohamedzayed92:
Currently, if you run the `bootstrap_from_checkout` command using the `assembl_user`, it will request for a sudo password on when executing `check_and_create_database_user`.  But `assembl_user` is not supposed to be sudoer so `check_and_create_database` should be executed with your sudoer user.


maparent:
LGTM

PR #403, opened 2018-04-12
upgrade to react 16.3

vincentfretin:
I upgraded react-test-renderer@latest react@latest react-dom@latest searchkit@latest react-datepicker@latest and
some minor upgrade of draft-convert react-router react-i18nify react-redux-i18n lodash linkifyjs react-bootstrap, not necessary the latest minor version for duplicate versions reason. I updated manually several time the yarn.lock and rerun yarn to be sure to not have duplicates.
The bundle.js size lost 400 kB mainly I think because we don't bundle two versions of jquery :) linkifyjs included another version of jquery that annotatorjs apparently.
There is zero warning in the console or in jest.

### DIFF
--- a/assembl/static2/package.json
+++ b/assembl/static2/package.json
@@ -9,7 +9,7 @@
   "scripts": {
     "test": "jest",
     "test:watch": "npm test -- --watch",
-    "build": "rm -rf node_modules/react-i18nify/node_modules;webpack --progress --colors",
+    "build": "webpack --progress --colors",
     "coverage": "jest --coverage",
     "dev": "webpack -d --watch",
     "eslint": "eslint js/app tests --ext .js --ext .jsx --cache",
@@ -62,7 +62,8 @@
     "prettier": "^1.8.2",
     "prettier-eslint": "^8.2.2",
     "prettier-eslint-cli": "^4.4.0",
-    "react-test-renderer": "^15.4.1",
+    "raf": "^3.4.0",
+    "react-test-renderer": "^16.3.1",
     "sass-loader": "^6.0.6",
     "style-loader": "^0.18.2",
     "stylelint": "^8.2.0",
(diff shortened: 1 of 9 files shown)


vincentfretin:
I didn't take care of the warnings in strict mode about deprecated api, this will be for another time, see https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html and https://reactjs.org/docs/strict-mode.html if you didn't read them already. But for new components you can already avoid the deprecated api and use the new ones.

assembl/static2/package.json
```diff
@@ -9,7 +9,7 @@
   "scripts": {
     "test": "jest",
     "test:watch": "npm test -- --watch",
-    "build": "rm -rf node_modules/react-i18nify/node_modules;webpack --progress --colors",
+    "build": "webpack --progress --colors",
```
ayyazdaniaryan: This might need to be back in another environment. Only been tested on your debian, yeah?

assembl/static2/package.json
```diff
@@ -96,27 +97,27 @@
     "piwik-react-router": "assembl/piwik-react-router#skip_piwik_initialization",
     "prop-types": "^15.5.8",
     "rc-slider": "^8.6.0",
-    "react": "^15.4.1",
+    "react": "^16.3.1",
```
ayyazdaniaryan: What kind of impact will this have on `react-virtualized` ? I can imagine that their package will either benefit greatly from this, which in turn, might mean breaking change upgrade for use (hopefully not).

assembl/static2/package.json
vincentfretin: We now use a version of react-i18nify that doesn't use a hard coded react version. This `rm -rf` was needed to not have duplicate packages in the bundle.

assembl/static2/package.json
vincentfretin: I don't see the relation with react-virtualized. Are you talking about async rendering? This is not currently enabled in react 16.3, it will be in react 17. There is lots of fixes to do on deprecated api before we can upgrade to a react 17.
https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html

vincentfretin:
We should do a QA on a machine locally with the console open to see the eventual errors or warnings.

PR #407, opened 2018-04-16
Change a taxonomy name + add a new taxonomy

paolina92:


assembl/models/idea_content_link.py
```diff
@@ -257,7 +262,7 @@ class Enum(Enum):
         display_thread = 8
         display_tokens = 9
         display_open_questions = 10
-        display_bright_mirror = 11
+        display_design_fiction = 11
```
vincentfretin: You shouldn't rename that.

assembl/models/idea_content_link.py
```diff
@@ -300,9 +305,9 @@ class Enum(Enum):
             "fr": "Activer Q° ouvertes",
             "en": "Display Open questions",
         },
-        Enum.display_bright_mirror: {
-            "fr": "Activer Bright Mirror",
-            "en": "Display Bright Mirror"
+        Enum.display_design_fiction: {
```
vincentfretin: You shouldn't rename that. Modify only the translations.

assembl/models/idea_content_link.py
paolina92: ok

PR #409, opened 2018-04-17
Confirm unsaved changes

cedricmessiant:
In any admin section, warn the user if he has unsaved changes when he tries to leave the page
(added flow for the admin pages when they were missing)

Closes DEVAS-1476


assembl/static2/js/app/pages/surveyAdmin.jsx
```diff
@@ -65,6 +82,7 @@ class SurveyAdmin extends React.Component {
 
       runSerial(mutationsPromises)
         .then(() => {
+          this.setState({ refetching: true });
```
vincentfretin: Don't you need to set again refatching to false?
`refetchThematics().then(() => this.setState({ refetching: false }));` ?

assembl/static2/js/app/router.js
```diff
@@ -0,0 +1,26 @@
+import { useRouterHistory } from 'react-router';
+import { createHistory, useBeforeUnload } from 'history';
+
+import PiwikReactRouter from 'piwik-react-router';
+import get from 'lodash/get';
+import 'bootstrap/dist/css/bootstrap.css';
```
vincentfretin: Why do have this bootstrap.css import here? Bad copy paste?

cedricmessiant: oh yes, forgot to remove it, thanks

assembl/static2/js/app/pages/surveyAdmin.jsx
cedricmessiant: you're right, it's probably better (even if there is no visible problem as it is)

PR #411, opened 2018-04-19
Sidemenu ie bugs

MilanBarande:
This is fixing the bug [DEVAS-1480](https://bluenove.atlassian.net/browse/DEVAS-1480)
Now, the table of content will disappear at the moment its bottom touches the top of the conclusion block.

### DIFF
--- a/assembl/static2/js/app/components/synthesis/sideMenu.jsx
+++ b/assembl/static2/js/app/components/synthesis/sideMenu.jsx
@@ -11,14 +11,16 @@ type SideMenuProps = {
   rootIdeas: Array<SynthesisIdea>,
   descendants: Array<SynthesisIdea>,
   synthesisPostId: string,
-  ideaOnScroll?: string
+  ideaOnScroll?: string,
+  show: boolean,
+  innerRef?: Function
 };
 
 const SideMenu = (props: SideMenuProps) => {
-  const { rootIdeas, descendants, synthesisPostId, ideaOnScroll, innerRef } = props;
+  const { rootIdeas, descendants, synthesisPostId, ideaOnScroll, innerRef, show } = props;
   const slug = getDiscussionSlug();
   return (
-    <div className="synthesis-side-menu" ref={innerRef}>
+    <div className="synthesis-side-menu" ref={innerRef} style={{ display: show ? 'block' : 'none' }}>
       <Translate value="synthesis.tableOfContents" className="dark-title-4" />
       <div className="title-hyphen block">&nbsp;</div>
       {rootIdeas.map((rootIdea, index) => (
@@ -38,7 +40,8 @@ const SideMenu = (props: SideMenuProps) => {
 };
 
 SideMenu.defaultProps = {
-  ideaOnScroll: null
+  ideaOnScroll: null,
+  innerRef: null
 };
 
 export default SideMenu;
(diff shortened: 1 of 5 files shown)


assembl/static2/js/app/pages/synthesis.jsx
```diff
@@ -129,7 +170,13 @@ export class DumbSynthesis extends React.Component<void, SynthesisProps, Synthes
               </Col>
             </Row>
             {conclusion && (
-              <Section title="conclusion" translate className="synthesis-block" id="conclusion">
+              <Section
+                title="conclusion"
+                translate
+                className="synthesis-block"
+                id="conclusion"
+                innerRef={this.updateConclusionBlock}
```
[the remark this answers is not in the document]
MilanBarande: works just fine, but you make think of a potential refactoring using this syntax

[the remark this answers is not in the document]
ayyazdaniaryan: I haven't coded JSX in a while, hence why I asked. If it works, then all good!

assembl/static2/js/app/components/synthesis/sideMenu.jsx
```diff
@@ -11,14 +11,16 @@ type SideMenuProps = {
...
+  innerRef?: Function
 };
 
 const SideMenu = (props: SideMenuProps) => {
-  const { rootIdeas, descendants, synthesisPostId, ideaOnScroll } = props;
+  const { rootIdeas, descendants, synthesisPostId, ideaOnScroll, innerRef, show } = props;
   const slug = getDiscussionSlug();
   return (
-    <div className="synthesis-side-menu">
+    <div className="synthesis-side-menu" ref={innerRef} style={{ display: show ? 'block' : 'none' }}>
```
cedricmessiant: Is there a reason to have it hidden by style instead of not adding it to the DOM ?

MilanBarande: Isn't it better in terms of performance ? At first I did this because I thought I could still measure the element's height if it's hidden but it's not possible. I just kept it like this

cedricmessiant: about performance, I imagine it's the contrary because you don't render this part of the DOM. But maybe I am missing something @vincentfretin ?

vincentfretin: Destroying and remounting a component tree is far more expensive than just flipping a css property.  So what @MilanBarande did is ok.

assembl/static2/js/app/pages/synthesis.jsx
```diff
@@ -103,15 +136,23 @@ export class DumbSynthesis extends React.Component<void, SynthesisProps, Synthes
...
-                    descendants={descendants}
-                    synthesisPostId={synthesisPostId}
-                    ideaOnScroll={ideaOnScroll}
-                  />
-                )}
+              <Col
+                md={3}
+                className="affix"
+                style={{
+                  top: `${sideMenuTopPercentage} %`,
```
vincentfretin: Does it work with a space before `%` ? I never put space before unit in css.

MilanBarande: Yes it does

swergas:
OK for me

PR #412, opened 2018-04-19
Add storybook and first stories

cedricmessiant:
Storybook config, doc and first stories

### DIFF
--- a/assembl/static2/.storybook/webpack.config.js
+++ b/assembl/static2/.storybook/webpack.config.js
@@ -1,0 +1,21 @@
+const path = require('path');
+
+module.exports = {
+  module: {
+    rules: [
+      {
+        test: /\.scss$/,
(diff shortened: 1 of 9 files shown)


assembl/static2/.storybook/webpack.config.js
```diff
@@ -0,0 +1,21 @@
+const path = require('path');
```
swergas: Why is this file needed in folder .storybook?

swergas: Oh is it because storybook runs its own web server?

vincentfretin: yes, it starts its own webpack server on port 9001

cedricmessiant:
@swergas we need to extend default storybook webpack config to have scss files support See https://storybook.js.org/configurations/custom-webpack-config/#extend-mode

PR #413, opened 2018-04-20
fix survey phase export link

swergas:
Empty string does not work, looks like there is a problem in parse() or get() function of literalStringParser.js introduced in https://github.com/assembl/assembl/commit/1a84b7702ec6396ded8e3aa9ba4b6d0eba1133e7#diff-161b47c83cd40ca16c78a9b658c5ca73


swergas:
Question is do we want to fix the underlying problem, that can be much more tricky to solve and have repercussions in other string interpolations elsewhere?

cedricmessiant:
I'd rather fix the real problem. I guess that with unit tests for these functions it should be safe and easy to fix and it will avoid to have the same issues for other urls in some time

PR #414, opened 2018-04-20
Delete my account devas1468

mohamedzayed92:


assembl/graphql/user.py
```diff
@@ -175,3 +178,40 @@ def mutate(root, args, context, info):
...
+        discussion_id = context.matchdict['discussion_id']
+        user_id = context.authenticated_userid or Everyone
+
+        global_id = args.get('id')
+        id_ = int(Node.from_global_id(global_id)[1])
+        user = cls.get(id_)
+
+        permissions = get_permissions(user_id, discussion_id)
+        allowed = user.user_can(
+            user_id, CrudPermissions.READ, permissions)
```
vincentfretin: Please use the new api to check the permission for new mutations.

assembl/static2/js/app/utils/translations.js
```diff
@@ -386,6 +386,12 @@ const Translations = {
       changePassword: "Modifier mon mot de passe",
       passwordModifiedSuccess: "Votre mot de passe a été modifié avec succès",
       saveSuccess: "Votre profil a été mis à jour avec succès",
+      deleteMyAccount: "Supprimer mon compte",
+      deleteMyAccountConfirmation: "Supprimer mon compte et l'ensemble de mes données",
+      deleteMyAccountText:
+        "En supprimant votre compte, vous supprimez définitivement l'ensemble de vos données personnelles ainsi que votre compte. Vous ne pourrez plus contribuer à la consultation sans compte.",
+      deleteMyAccountModale:
```
vincentfretin: modal sans e

assembl/static2/js/app/components/profile/avatar.jsx
```diff
@@ -0,0 +1,25 @@
+// @flow
```
swergas: Is it normal that the component is called avatar but does not contain any avatar (picture) of a user?

swergas:
This PR should probably contain the following feature of the "To be verified" list:
"add unique token per form display on remove account form to protect against hackers that trick users into using links that delete their accounts"

PR #415, opened 2018-04-24
add the option to delete an uploaded image and change fileUploader design in the survey pageForm

MilanBarande:
[Link to the JIRA ticket](https://bluenove.atlassian.net/browse/DEVAS-1373)


cedricmessiant:
just added the back-end part for the image deletion. @swergas @vincentfretin a little review of this part would be nice

assembl/static2/js/app/components/common/fileUploader.jsx
```diff
@@ -116,9 +124,28 @@ class FileUploader extends React.Component<Object, FileUploaderProps, FileUpload
...
+            </OverlayTrigger>
           </div>
-        ) : null}
-        {!isImage && <div className="preview-title">{this.state.fileName}</div>}
+        ) : (
+          <div className={fileSrc && isImage ? 'preview' : 'hidden'}>
+            <img
+              src={fileSrc}
+              ref={(p) => {
+                this.preview = p;
```
swergas: this.preview seems to be not used in the rest of the component

assembl/static2/tests/unit/components/common/__snapshots__/fileUploader.spec.jsx.snap
```diff
@@ -47,6 +47,14 @@ exports[`FileUploader component should render a file uploader with the filename
       Choose a file to upload
     </span>
   </button>
+  <div
+    className="hidden"
+  >
+    <img
+      alt="preview"
+      src={undefined}
```
swergas: is it undefined on purpose?

assembl/static2/js/app/components/common/fileUploader.jsx
MilanBarande: @cedricmessiant is there a purpose for this ref ?

assembl/static2/tests/unit/components/common/__snapshots__/fileUploader.spec.jsx.snap
MilanBarande: Yes, this snapshot is for this test: ```FileUploader component should render a file uploader with the filename if the file is not an image (fileOrUrl is a File object)```

vincentfretin:
backend is ok

assembl/static2/js/app/components/common/fileUploader.jsx
cedricmessiant: I guess that there was a long time ago but it does not seem to be the case anymore. Can't find when that changed with blame and history so I guess we can remove it if everything seems fine

MilanBarande: Ok I'll remove it

PR #417, opened 2018-04-24
Adding functions for restore from backup

mohamedzayed92:
Some functions that I will use for automation of backup restore. Documentation is commited on the a branch from master since the doc is in bluenove_server_configs

### DIFF
--- a/fabfile.py
+++ b/fabfile.py
@@ -777,8 +777,6 @@ def updatemaincode(backup=False):
             run('git fetch')
             run('git checkout %s' % env.gitbranch)
             run('git pull %s %s' % (env.gitrepo, env.gitbranch))
-    else:
-        run('git rev-parse HEAD > .git_commit')
 
 
 @task


mohamedzayed92:
@vincentfretin  no we can't just `bootstrap_from_checkout` because it updates the code with the latest version which we do not want. The backup is designed to keep the database with the corresponding code that runs with it at the time it was backed up. As for the `databse_restore` it is designed for production instances or for staging and it has some commands that are not relevant to the restore operation.

vincentfretin:
You can add a parameter to bootstrap_from_checkout to not update the code.

fabfile.py
```diff
@@ -724,35 +724,43 @@ def bootstrap(projectpath):
...
-    execute(app_compile_nodbupdate)
-    execute(set_file_permissions)
-    execute(backup_database_restore)
-    execute(app_reload)
-    execute(webservers_reload)
+    execute(bootstrap_from_checkout, backup=backup)
+
+
+@task
+def install_assembl_requirements():
```
swergas: This task is mentioned nowhere. We don't know when we have to run it. If you add it, mention it in documentation (INSTALL.md ?)

PR #423, opened 2018-05-02
Taxonomie part 4 devas 1247

mohamedzayed92:


assembl/views/api2/discussion.py
```diff
@@ -540,6 +540,13 @@ def extract_taxonomy_csv(request):
             message = "no message"
         if not message:
             message = "no message"
+
+        if thematic == "no thematic associated":
+            idea_ids = m.Idea.get_idea_ids_showing_post(query.id)
+            for thematic_id in reversed(idea_ids):
+                if db.query(m.Idea).filter(m.Idea.id == thematic_id).first().title:
```
amensouissi: use a var here: first_thematic_title = db.query(m.Idea).filter(m.Idea.id == thematic_id).first().title

swergas: you could put `db.query(m.Idea).filter(m.Idea.id == thematic_id).first().title` in a variable here for readability because you reuse it

assembl/views/api2/discussion.py
```diff
@@ -553,13 +560,13 @@ def extract_taxonomy_csv(request):
         else:
             qualify_by_action = "no qualify by action"
         owner_of_the_message = db.query(m.User).filter(m.User.id == query.creator_id).first().name
-        published_on = str(query.creation_date)
+        published_on = str(query.creation_date)[:19]
```
swergas: put 19 in a const (capitalized) variable and add a comment explaniing how the value of 19 has been obtained

assembl/views/api2/discussion.py
```diff
@@ -540,6 +540,13 @@ def extract_taxonomy_csv(request):
             message = "no message"
         if not message:
             message = "no message"
+
+        if thematic == "no thematic associated":
```
amensouissi: add a global var here: no_thematic_associated = "no thematic associated"

assembl/views/api2/discussion.py
amensouissi: I think we need to format the date before. It's not clean.

assembl/views/api2/discussion.py
```diff
@@ -540,6 +540,13 @@ def extract_taxonomy_csv(request):
...
+            idea_ids = m.Idea.get_idea_ids_showing_post(query.id)
+            for thematic_id in reversed(idea_ids):
+                if db.query(m.Idea).filter(m.Idea.id == thematic_id).first().title:
+                    thematic = db.query(m.Idea).filter(m.Idea.id == thematic_id).first().title.best_lang(user_prefs).value
```
cedricmessiant: rename variable to `thematic_title` and I agree that you should add a variable for the thematic

assembl/views/api2/discussion.py
```diff
@@ -553,13 +560,13 @@ def extract_taxonomy_csv(request):
...
-        published_on = str(query.creation_date)
+        published_on = str(query.creation_date)[:19]
         harvester = db.query(m.User).filter(m.User.id == extract.owner_id).first().name
-        harvested_on = str(extract.creation_date)
+        harvested_on = str(extract.creation_date)[:19]
         nugget = "Yes" if extract.important else "No"
         extract_info = {
             "Thematic": thematic.encode('utf-8'),
-            "Message": sanitize_html(message).encode('utf-8'),
+            "Message": sanitize_text(message).encode('utf-8'),
```
cedricmessiant: why do you have to change this ?

assembl/views/api2/discussion.py
cedricmessiant: yes, if `query.creation_date` is a `datetime.datetime`, use python tools to format the date : https://docs.python.org/2/library/datetime.html#strftime-strptime-behavior

assembl/views/api2/discussion.py
mohamedzayed92: @cedricmessiant I have to change this because when I export the csv, it exports the text with HTML tags inside.

PR #429, opened 2018-05-14
Upgrade flow version from 0.52 to 0.71 with all required code changes

swergas:


assembl/static2/js/app/components/administration/exportSection.jsx
```diff
@@ -16,7 +16,7 @@ type Props = {
   annotation: string
 };
 
-class ExportSection extends React.Component<Object, Props, void> {
+class ExportSection extends React.Component<Props, void> {
   props: Props;
```
vincentfretin: With latest flow version, you can remove `props: Props;` and `state: State` on the class.

assembl/static2/js/app/components/administration/voteSession/customizeGaugeForm.jsx
```diff
@@ -46,7 +46,7 @@ type State = {
   saving: boolean
 };
 
-export class DumbCustomizeGaugeForm extends React.Component<DefaultProps, Props, State> {
+export class DumbCustomizeGaugeForm extends React.Component<Props, State> {
```
vincentfretin: You should remove DefaultProps type above now.

assembl/static2/js/app/components/common/attachFileForm.jsx
```diff
@@ -13,7 +13,8 @@ type AttachFileFormState = {
   file: File | null
 };
 
-class AttachFileForm extends React.Component<*, AttachFileFormProps, AttachFileFormState> {
+class AttachFileForm extends React.Component<AttachFileFormProps, AttachFileFormState> {
+  static defaultProps: *;
```
vincentfretin: Was this really needed? I guess its flow-upgrade that put this there? I think you can remove all `static defaultProps: *` that was added. I think we really meant void instead of any there.

assembl/static2/js/app/components/common/modifyPasswordForm.jsx
```diff
@@ -35,23 +35,23 @@ class ModifyPasswordForm extends React.Component<void, Props, State> {
     newPassword2: ''
   };
 
-  handleOldPasswordChange = (e: SyntheticEvent) => {
+  handleOldPasswordChange = (e: SyntheticEvent<>) => {
```
vincentfretin: You normally can use `SyntheticEvent<HTMLInputElement>` and remove the `if (e.target instanceof HTMLInputElement) `

assembl/static2/js/app/components/navbar/FlatNavbar.jsx
```diff
@@ -1,18 +1,12 @@
...
 import UserMenu from './UserMenu';
 
-export default class FlatNavbar extends React.PureComponent {
-  state: {
-    leftWidth: number,
-    rightWidth: number,
-    languageMenuWidth: number
-  };
-
+export default class FlatNavbar extends React.PureComponent<$FlowFixMeProps, $FlowFixMeState> {
```
vincentfretin: the State type was definied before

vincentfretin:
react-apollo and apollo-client removed flow types from their repo in latest version.
There are apollo-link-http_v1.2.x and react-apollo_v2.x.x in flow-typed now, but I'm not sure this is compatible with the version of react-apollo we currently use.

vincentfretin:
This is unfortunate we have @noflow on lots of files. We won't have warnings when refactoring code.
This is why I wanted to upgrade flow together with latest react-apollo and draft.js. Did you try upgrading draft.js? Latest version normally includes a version of flow types compatible with latest flow version.

vincentfretin:
FYI, with flow 0.71, we can now use https://flow.org/en/docs/strict/ to avoid using Object, any, and Function.

cedricmessiant:
yes, I'm concerned that we will have lots `@noflow` for a long period of time and so we will miss errors that could be useful. Maybe we should at least try to upgrade draft-js (it is easily testable). Apollo packages may be trickier...

assembl/static2/js/app/components/administration/voteSession/customizeGaugeForm.jsx
swergas: I did

assembl/static2/js/app/components/common/attachFileForm.jsx
swergas: Yes, this code change has been made by flow-upgrade. I did not try to optimize what it did, my knowledge of flow subtilities is still limited. What "any" are you talking about?

swergas:
I'm in favor of a succession of smallest possible pull requests that are each easy to merge (and to fix conflicts) instead of a big one that would try to address lots of problems but that would be a pain to merge. To me, lots of things you mention are interesting but out of scope of this smallest possible PR.

swergas:
Also I don't see any relationship with draftjs? From what I recall, flow errors I encountered did not mention this lib.

vincentfretin:
When I tried to upgrade flow a few months ago, there were lots of errors from draft-js types that wasn't compatible with latest flow.
Maybe the flow types generated by apollo-codegen are not compatible too?

It seems you are the only one having issues with flow 0.52, something related to Ubuntu 16.04 maybe.
Let's take our time to try to fix more flow issues, I'll look into your branch end of next week.
In the meantime, you can execute flow in a docker container like this:
```
cd static2
docker run --rm -v $PWD:/app node:6 bash -c "cd /app; npm run flow"
```
and commit with 
```
SKIP=flow git commit ...
```

vincentfretin:
I rebased, squashed, and did some more cleanup. There are still files with noflow but a lot less. All graphql and react-apollo are not covered.

vincentfretin:
I checked on develop, the graphql coverage wasn't properly working either, I think because we use separate graphql files. So we don't lose anything here.

PR #432, opened 2018-05-18
create page for thread and multicolumn extract

MilanBarande:


assembl/static2/css/components/multiColumnAdmin.scss
```diff
@@ -0,0 +1,21 @@
...
+    .admin-content {
+      text-align: center;
+
+      .form-group {
+        text-align: left;
+        display: inline-block;
+
+        .export-language-dropdown {
+          margin-top: 20px;
+          position: absolute;
```
ayyazdaniaryan: I think because there is no position on any of the parents, this position will be absolute to the viewport, which might not be what you're looking for in case of mobile. Please verify.

assembl/static2/js/app/pages/multiColumnsAdmin.jsx
```diff
@@ -1,9 +1,97 @@
...
+
+  componentDidMount() {
+    this.props.router.setRouteLeaveHook(this.props.route, this.routerWillLeave);
+  }
+
+  componentWillUnmount() {
+    this.props.router.setRouteLeaveHook(this.props.route, null);
+  }
+
+  handleExportLocaleChange = (locale) => {
```
ayyazdaniaryan: I would bind these functions in the constructor of this class. 
eg. 
```
this.handleExportLocaleChange = this.handleExportLocaleChange.bind(this);
```

https://medium.com/shoutem/react-to-bind-or-not-to-bind-7bf58327e22a

assembl/static2/js/app/pages/multiColumnsAdmin.jsx
```diff
@@ -1,9 +1,97 @@
...
+    this.setState({ exportLocale: locale });
+  };
+
+  handleTranslationChange = (shouldTranslate) => {
```
ayyazdaniaryan: Same here.

assembl/static2/js/app/pages/multiColumnsAdmin.jsx
```diff
@@ -1,9 +1,97 @@
...
+            handleTranslationChange={this.handleTranslationChange}
+            exportLink={exportLink}
+            translate={translate}
+            exportLocale={exportLocale}
+            languages={languages}
+            annotation="multicolumnAnnotation"
+          />
+        )}
+        {!isNaN(currentStep) && (
+          <Navbar currentStep={currentStep} totalSteps={1} phaseIdentifier="multiColumn" />
```
ayyazdaniaryan: Why show the Navbar when the admin currently only has 1 step? If it's for the future, I would put another check, 

```
const totalSteps = 1;
!isNan(currentStep) && totalSteps > 1 && ...
```

In future, if `totalSteps` increases, the code will be rendered, otherwise the navbar is left out.

ayyazdaniaryan: Unless you absolutely want a navbar to be present even in the case of 1 step

assembl/static2/js/app/pages/threadAdmin.jsx
```diff
@@ -1,9 +1,97 @@
...
+
+  componentDidMount() {
+    this.props.router.setRouteLeaveHook(this.props.route, this.routerWillLeave);
+  }
+
+  componentWillUnmount() {
+    this.props.router.setRouteLeaveHook(this.props.route, null);
+  }
+
+  handleExportLocaleChange = (locale) => {
```
ayyazdaniaryan: Bind these methods as well

assembl/static2/js/app/pages/threadAdmin.jsx
```diff
@@ -1,9 +1,97 @@
...
+            handleTranslationChange={this.handleTranslationChange}
+            exportLink={exportLink}
+            translate={translate}
+            exportLocale={exportLocale}
+            languages={languages}
+            annotation="threadAnnotation"
+          />
+        )}
+        {!isNaN(currentStep) && (
+          <Navbar currentStep={currentStep} totalSteps={1} phaseIdentifier="thread" />
```
ayyazdaniaryan: Same comment as above...

assembl/static2/js/app/utils/translations.js
```diff
@@ -1286,6 +1288,8 @@ const Translations = {
       export: {
         defaultAnnotation: "You can export all of the data by clicking on the export button",
         surveyAnnotation: "You can export all the survey module data by clicking on the Export button",
+        threadAnnotation: "You can export all the thread module data by clicking on the Export butotn",
+        multicolumnAnnotation: "You can export all the multicolumn data by clicking on the Export button",
```
ayyazdaniaryan: `multi-column` or `Multi Column` as there is no word like this in English.

assembl/static2/js/app/pages/multiColumnsAdmin.jsx
MilanBarande: But the point of using an arrow function here is to avoid having to bind the function

assembl/static2/js/app/pages/multiColumnsAdmin.jsx
MilanBarande: At first I hadn't put it but I thought it made it clear that there is only one step of admin to deal with at the moment (1/1) and not that others are available but not displaying for any sort of reason (bug).

assembl/static2/js/app/utils/translations.js
MilanBarande: Oups, true :)

assembl/static2/css/components/multiColumnAdmin.scss
ayyazdaniaryan: Oh yea! There is no mobile admin section! Okay, then keep this. Good call.

assembl/static2/js/app/pages/multiColumnsAdmin.jsx
MilanBarande: Okay I didn't know. I think we have been using arrow functions all over the project

MilanBarande: So I made the modifications. But I just found this https://github.com/facebook/react/issues/9851
apparently it's pretty much the same thing

vincentfretin: @ayyazdaniaryan class properties are transformed with babel (click on the Try button in the example https://babeljs.io/docs/plugins/transform-class-properties/) to an efficient syntax. You can have perf issue with binding in the render method yes because it will create new functions, but class properties are completely different here. We prefer the class properties syntax in the project because of flow.

assembl/static2/js/app/pages/threadAdmin.jsx
vincentfretin: Arrow functions and class properties like this are automatically binded to this @ayyazdaniaryan

assembl/static2/js/app/pages/multiColumnsAdmin.jsx
MilanBarande: Good idea

assembl/static2/js/app/pages/threadAdmin.jsx
paolina92: Since we're using Flow, we should use arrow function?

assembl/static2/js/app/pages/threadAdmin.jsx
```diff
@@ -1,9 +1,97 @@
 import React from 'react';
+import { compose, graphql } from 'react-apollo';
+import { connect } from 'react-redux';
+import { get } from '../utils/routeMap';
+import ExportSection from '../components/administration/exportSection';
+import Navbar from '../components/administration/navbar';
+import DiscussionPreferenceLanguage from '../graphql/DiscussionPreferenceLanguage.graphql';
```
paolina92: Rename in `DiscussionPreferenceLanguageQuery` for more clarity?

assembl/static2/js/app/pages/multiColumnsAdmin.jsx
```diff
@@ -1,9 +1,97 @@
 import React from 'react';
+import { compose, graphql } from 'react-apollo';
+import { connect } from 'react-redux';
+import { get } from '../utils/routeMap';
+import ExportSection from '../components/administration/exportSection';
+import Navbar from '../components/administration/navbar';
+import DiscussionPreferenceLanguage from '../graphql/DiscussionPreferenceLanguage.graphql';
```
paolina92: Rename in `DiscussionPreferenceLanguageQuery` for more clarity?

assembl/static2/js/app/pages/threadAdmin.jsx
vincentfretin: yes

PR #434, opened 2018-05-22
fix gauge admin minimum and maximum fields were not showing

swergas:


assembl/static2/js/app/components/common/formControlWithLabel.jsx
```diff
@@ -140,7 +142,7 @@ class FormControlWithLabel extends React.Component<Object, FormControlWithLabelP
 
   render() {
     const { id, labelAlwaysVisible, type, value, helperText } = this.props;
-    const displayLabel = labelAlwaysVisible || type !== 'rich-text' ? value : false;
+    const displayLabel = (labelAlwaysVisible || type !== 'rich-text') ? (value || (value === 0)) : false;
```
vincentfretin: I just had an issue with the previous code where labelAlwaysVisible wasn't taken into account and I had to use 
```
const displayLabel = labelAlwaysVisible || (type !== 'rich-text' ? value : false);
```
to fix it. So I guess you should change the code here to:
```
const displayLabel = labelAlwaysVisible || (type !== 'rich-text' ? (value || (value === 0) : false);
```

PR #442, opened 2018-05-29
format the stats depending on the language

MilanBarande:
[Jira Ticket](https://bluenove.atlassian.net/browse/DEVAS-1504)

I just created a global function to format the number following the user story and use it in headerStatistics

### DIFF
--- a/assembl/static2/js/app/components/common/headerStatistics.jsx
+++ b/assembl/static2/js/app/components/common/headerStatistics.jsx
@@ -16,7 +16,7 @@ export const statContributions = (numContributions: number): StatisticElementPro
   metricNameTranslateKey: 'home.contribution'
 });
 
-export const statMessages = (numPosts: number) => ({
+export const statMessages = (numPosts: number): StatisticElementProps => ({
   iconName: 'message',
   metricValue: numPosts,
   metricNameTranslateKey: 'home.messages'
--- a/assembl/static2/js/app/components/common/headerStatistics.jsx
+++ b/assembl/static2/js/app/components/common/headerStatistics.jsx
@@ -46,7 +46,7 @@ const StatisticElement = ({ metricNameTranslateKey, iconName, metricValue }: Sta
       <div className={`stat-icon assembl-icon-${iconName} white`} />
       <div className="stat">
         <div className="stat-nb">
-          <Localize value={10000} />
+          <Localize value={metricValue} />
         </div>
         <div className="stat-nb stat-label">
           <Translate value={metricNameTranslateKey} count={metricValue} />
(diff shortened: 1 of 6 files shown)


assembl/static2/js/app/utils/globalFunctions.js
```diff
@@ -17,6 +17,11 @@ export const getDiscussionId = () => getInputValue('discussion-id');
 
 export const getDiscussionSlug = () => getInputValue('discussion-slug');
 
+export const formatNumber = (number: number | string, locale: string) => {
+  const separator = locale !== 'en' ? ' ' : ',';
+  return number.toString().replace(/\B(?=(\d{3})+(?!\d))/g, separator);
+};
+
```
vincentfretin: You are only supporting two languages here. Please use existing lib for this. Look at react-redux-i18n Localize component for this which uses the standard Intl.NumberFormat behind the scene.

vincentfretin:
https://www.npmjs.com/package/react-redux-i18n
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/NumberFormat

assembl/static2/js/app/utils/globalFunctions.js
MilanBarande: Well the story only talks about dealing with english and every other language should have whitespaces. But I agree if we have a simple way to deal with every language why not. Need to take a look at this, I didn't know about it

assembl/static2/js/app/components/common/headerStatistics.jsx
```diff
@@ -10,52 +10,54 @@ type StatisticElementProps = {
 
 type StatElements = Array<StatisticElementProps>;
 
-export const statContributions = (numContributions: number): StatisticElementProps => ({
+export const statContributions = (numContributions: number) => ({
```
vincentfretin: Why did you remove `: StatisticElementProps`?

assembl/static2/js/app/components/common/headerStatistics.jsx
```diff
@@ -40,22 +40,24 @@ export const statParticipations = (numParticipations: number): StatisticElementP
...
+          <Translate value={metricNameTranslateKey} count={metricValue} />
         </div>
       </div>
     </div>
   </div>
 );
 
 const mapElementsPropsToComponents = elemsProps =>
-  elemsProps.map((elementProps, index) => <StatisticElement key={index} {...elementProps} />);
+  elemsProps.map((elementProps, index) => <StatisticElement key={`stat-${index}`} {...elementProps} />);
```
vincentfretin: you can use elemsProps.metricNameTranslateKey as key instead to avoid the eslint warning

PR #445, opened 2018-06-01
Adding extra fields to csv export

mohamedzayed92:


assembl/views/api2/discussion.py
```diff
@@ -740,12 +745,21 @@ def get_visitors(request):
...
     for st in discussion.agent_status_in_discussion:
         if not getattr(st, attribute, None):
             continue
+        profile_fields = db.query(ProfileField).filter(ProfileField.discussion_id == discussion.id).filter(
+            ProfileField.agent_profile_id == st.agent_profile.id).all()
         data = {"time": getattr(st, attribute),
                 "name": (st.agent_profile.name or '').encode("utf-8"),
                 "email": (st.agent_profile.get_preferred_email() or '').encode("utf-8")}
+
+        for configurable_field in configurable_fields:
```
cedricmessiant: I think that it is not very performant and you should be able to do a query with a join to get all the profile fields for your user that are "linked" to a custom configurable field.

assembl/views/api2/discussion.py
```diff
@@ -740,12 +745,21 @@ def get_visitors(request):
...
+            for profile_field in profile_fields:
+                if profile_field.configurable_field_id == configurable_field.id:
+                    data.update({configurable_field.title.entries[0].value.encode("utf-8"): profile_field.value_data['value'].encode("utf-8")})
```
cedricmessiant: shouldn't you take the title for the chosen language?

vincentfretin: yes, you need to use `.best_lang(user_prefs)` instead of `.entries[0]` here.

assembl/views/api2/discussion.py
```diff
@@ -730,6 +730,11 @@ def get_visitors(request):
     fieldnames = ["time", "name", "email"]
     extra_columns_info = (None if 'no_extra_columns' in request.GET else
                           load_social_columns_info(discussion, "en"))
+    db = discussion.db
+    from assembl.models import TextField
+    configurable_fields = db.query(TextField).filter(TextField.discussion_id == discussion.id).filter(TextField.identifier == "CUSTOM").all()
```
vincentfretin: This is not correct, you're retrieving only TextField, and not SelectField. You need to use `db.query(AbstractConfigurableField)`  here. For the SelectField value, it's an array of graphene_id, you will need to get the id from graphene id, and get the label of the field associated with it. concatenate the items with comma (for now it's only one value in the array, but maybe we will have a multivalued selectfield in the future)

assembl/views/api2/discussion.py
```diff
@@ -730,6 +730,11 @@ def get_visitors(request):
...
+    for configurable_field in configurable_fields:
+        fieldnames.append((configurable_field.title.entries[0].value).encode("utf-8"))
```
vincentfretin: use best_lang

assembl/views/api2/discussion.py
vincentfretin: indeed
```
profile_fields = db.query(ProfileField).filter(ProfileField.discussion_id == discussion.id).join(ProfileField.configurable_field).filter(
            ProfileField.agent_profile_id == st.agent_profile.id).filter(AbstractConfigurableField.identifier == "CUSTOM").all()
```

assembl/views/api2/discussion.py
```diff
@@ -740,12 +745,21 @@ def get_visitors(request):
...
         if not getattr(st, attribute, None):
             continue
+        profile_fields = db.query(ProfileField).filter(ProfileField.discussion_id == discussion.id).filter(
+            ProfileField.agent_profile_id == st.agent_profile.id).all()
         data = {"time": getattr(st, attribute),
                 "name": (st.agent_profile.name or '').encode("utf-8"),
                 "email": (st.agent_profile.get_preferred_email() or '').encode("utf-8")}
+
+        for configurable_field in configurable_fields:
+            for profile_field in profile_fields:
```
vincentfretin: instead of using a inner for loop, you should create a dict profile_field_by_configurable_field_id
and use `profile_field_by_configurable_field_id[configurable_field.id]` here.

assembl/views/api2/discussion.py
```diff
@@ -727,10 +727,19 @@ def get_visit_count(request):
              permission=P_DISC_STATS)
 def get_visitors(request):
     discussion = request.context._instance
+    user_prefs = LanguagePreferenceCollection.getCurrent()
     fieldnames = ["time", "name", "email"]
     extra_columns_info = (None if 'no_extra_columns' in request.GET else
                           load_social_columns_info(discussion, "en"))
-
+    db = discussion.db
+    from assembl import models as m
```
ayyazdaniaryan: I would do, it's a flavour/choice.
```
from assembl.models import AbstractConfigurableField, SelectFieldOption, ProfileField
```

assembl/views/api2/discussion.py
```diff
@@ -743,9 +752,25 @@ def get_visitors(request):
     for st in discussion.agent_status_in_discussion:
         if not getattr(st, attribute, None):
             continue
+        profile_fields = db.query(m.AbstractConfigurableField, m.ProfileField).filter(m.ProfileField.discussion_id == discussion.id).join(m.ProfileField.configurable_field).filter(
```
ayyazdaniaryan: it shouldn't matter in this case, but do an `outerjoin` so that we ensure each row is always present

assembl/views/api2/discussion.py
```diff
@@ -743,9 +752,25 @@ def get_visitors(request):
...
+            m.ProfileField.agent_profile_id == st.agent_profile.id).filter(m.AbstractConfigurableField.identifier == "CUSTOM").filter(m.ProfileField.configurable_field_id == m.AbstractConfigurableField.id).all()
```
ayyazdaniaryan: You can use `&&` operator in the first `filter` method and chain your filters that way instead of calling `filter` multiple times. Both works.

ayyazdaniaryan: Also, don't manually run `CUSTOM` here. The check should be against `ConfigurableFieldIdentifiersEnum.CUSTOM.value`

assembl/views/api2/discussion.py
vincentfretin: No, why? Keep the join. We have a filter `filter(m.AbstractConfigurableField.identifier == "CUSTOM")`, so even if you do outerjoin (left join), the filter will filter the NULL entries anyway. A ProfileField is required to reference a AbstractConfigurableField.

vincentfretin:
LGTM

PR #450, opened 2018-06-07
Survey posts search devas 1496

vincentfretin:
`assembl-reindex-all-contents local.ini`
will be needed on each server.

### DIFF
--- a/assembl/graphql/idea.py
+++ b/assembl/graphql/idea.py
@@ -1,8 +1,10 @@
 import os.path
 from collections import defaultdict
+from itertools import takewhile
 from random import sample as random_sample
 from random import shuffle as random_shuffle
 
+from graphql_relay.connection.arrayconnection import offset_to_cursor
 import graphene
 from graphene.relay import Node
 from graphene_sqlalchemy import SQLAlchemyConnectionField, SQLAlchemyObjectType
@@ -338,7 +340,8 @@ class Question(SecureObjectType, SQLAlchemyObjectType):
     title_entries = graphene.List(LangStringEntry)
     posts = SQLAlchemyConnectionField(
         'assembl.graphql.post.PostConnection',  # use dotted name to avoid circular import  # noqa: E501
-        random=graphene.Boolean())
+        random=graphene.Boolean(),
+        from_node=graphene.ID())
     thematic = graphene.Field(lambda: Thematic)
     total_sentiments = graphene.Int(required=True)
 
@@ -424,6 +427,12 @@ class Question(SecureObjectType, SQLAlchemyObjectType):
             else:
                 query = query.options(
                     models.LangString.joinedload_option(Post.body))
+
+        from_node = args.get('from_node')
+        if from_node and not args.get('after') and not args.get('before'):
+            post_id = int(Node.from_global_id(from_node)[-1])
+            node_idx = len(list(takewhile(lambda post: post[0] != post_id, query.with_entities(Post.id))))
+            args['after'] = offset_to_cursor(node_idx - 1)
 
         # pagination is done after that, no need to do it ourself
         return query
--- a/assembl/graphql/idea.py
+++ b/assembl/graphql/idea.py
@@ -429,7 +429,12 @@ class Question(SecureObjectType, SQLAlchemyObjectType):
                     models.LangString.joinedload_option(Post.body))
 
         from_node = args.get('from_node')
-        if from_node and not args.get('after') and not args.get('before'):
+        after = args.get('after')
+        before = args.get('before')
+        # If `from_node` is specified and after/before is None or empty string,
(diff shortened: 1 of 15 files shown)


assembl/static2/js/app/components/common/flatList.jsx
```diff
@@ -100,7 +106,13 @@ class FlatList extends React.Component<FlatListProps> {
   };
 
   render() {
-    const { networkStatus, items, ListItem, itemData, className } = this.props;
+    const { hash } = window.location;
```
cedricmessiant: can we avoid to use `window.location` by using `react-router` (location prop) ?

assembl/static2/js/app/components/search.jsx
```diff
@@ -108,15 +108,22 @@ if (v1Interface) {
...
       ideaId = hit._source.idea_id.length > 0 ? hit._source.idea_id[0] : null;
       if (!ideaId) {
         return undefined;
       }
       ideaBase64id = btoa(`Idea:${ideaId}`);
       postBase64id = btoa(`Post:${id}`);
-      return `/${slug}/debate/thread/theme/${ideaBase64id}/#${postBase64id}`;
+      const phaseId = hit._source.phase_id || 'thread';
+      if (phaseId === 'thread') {
+        return `/${slug}/debate/thread/theme/${ideaBase64id}/#${postBase64id}`;
```
cedricmessiant: shouldn't you use routes defined in `routes.json`?

assembl/static2/js/app/components/common/flatList.jsx
```diff
@@ -111,6 +123,17 @@ class FlatList extends React.Component<FlatListProps> {
     const entities = items.edges;
     return (
       <div className={className}>
+        {id && loadPreviousMessage ? (
+          <Button
+            className="button-dark button-submit"
+            style={{ marginBottom: '20px' }}
+            onClick={() => {
+              window.location.href = window.location.href.slice(0, window.location.href.indexOf('#'));
```
ayyazdaniaryan: Why not use the `browserhistory.push` here instead of forcing essentially a refresh?

assembl/static2/js/app/components/search.jsx
vincentfretin: yes, done

assembl/static2/js/app/components/debate/survey/posts.jsx
```diff
@@ -94,9 +95,16 @@ const mapDispatchToProps = dispatch => ({
...
   connect(mapStateToProps, mapDispatchToProps),
   graphql(QuestionPosts, {
-    options: props => ({
-      variables: { first: 8, after: '', id: props.questionId }
-    }),
+    options: (props) => {
+      const { hash } = window.location;
+      let id = null;
+      if (hash !== '') {
+        id = hash.replace('#', '').split('?')[0];
```
ayyazdaniaryan: Does this mean that you assume the query string will come after the hash? If so, that is against the recommended way to form a URI: https://tools.ietf.org/html/rfc3986#section-4.1

assembl/static2/js/app/components/search.jsx
ayyazdaniaryan: Doubly agree with @cedricmessiant . You can use existing routes to form these URLs...

assembl/graphql/idea.py
```diff
@@ -425,6 +428,12 @@ def resolve_posts(self, args, context, info):
                 query = query.options(
                     models.LangString.joinedload_option(Post.body))
 
+        from_node = args.get('from_node')
+        if from_node and not args.get('after') and not args.get('before'):
+            post_id = int(Node.from_global_id(from_node)[-1])
+            node_idx = len(list(takewhile(lambda post: post[0] != post_id, query.with_entities(Post.id))))
+            args['after'] = offset_to_cursor(node_idx - 1)
```
ayyazdaniaryan: How were you able to find this method to use?

vincentfretin: I dig into the code ;)

assembl/static2/js/app/components/common/flatList.jsx
vincentfretin: ok, I now use withRouter

assembl/static2/js/app/components/debate/survey/posts.jsx
vincentfretin: It's potentially the share modal that add a query string at the end of the url.

assembl/static2/js/app/components/common/flatList.jsx
vincentfretin: I fixed it, I know use react router, much cleaner.

vincentfretin:
Ok, should be good to merge when it's green.

PR #451, opened 2018-06-08
Delete survey message DEVAS-1512

cedricmessiant:


assembl/static2/js/app/components/debate/common/deletePostButton.jsx
```diff
@@ -0,0 +1,58 @@
...
+import React from 'react';
+import { Button } from 'react-bootstrap';
+import { Link } from 'react-router';
+import { graphql, type DocumentNode, type TVariables } from 'react-apollo';
+import { Translate } from 'react-redux-i18n';
+
+import { displayModal, closeModal } from '../../../utils/utilityManager';
+import deletePostMutation from '../../../graphql/mutations/deletePost.graphql';
+
+function confirmModal(deletePost, postId, refetchQueries) {
```
MilanBarande: ES5 😱 just kidding, I guess it has no importance

assembl/static2/js/app/components/debate/common/post/postBody.jsx
```diff
@@ -77,22 +77,29 @@ const Html = (props) => {
...
+                startOffset: tfi.offsetStart,
+                end: tfi.xpathEnd,
+                endOffset: tfi.offsetEnd
+              });
+              try {
+                const normedRange = range.normalize(html);
+                const nodes = jQuery(normedRange.textNodes()).filter((idx, node) => !white.test(node));
+                nodes.wrap(wrapper);
+              } catch (error) {
+                console.error(error); // eslint-disable-line no-console
```
MilanBarande: Do you only want the error to show in the console ? no displayAlert ?

assembl/static2/js/app/components/debate/common/sentiments.jsx
```diff
@@ -7,7 +9,31 @@ import { inviteUserToLogin, displayModal } from '../../../utils/utilityManager';
...
 
-const Sentiment = ({ sentimentCounts, mySentiment, sentiment, client, isSelected, postId, placement, isPhaseCompleted }) => {
+type SentimentProps = {
+  client: ApolloClient,
+  isPhaseCompleted: boolean,
+  isSelected: boolean,
+  mySentiment: ?string,
+  placement: string,
+  postId: string,
+  sentiment: Object, // TODO:
```
MilanBarande: TODO was about the Object shape I guess

assembl/static2/js/app/components/debate/common/post/postBody.jsx
ayyazdaniaryan: TODO: Adapt Raven to V2 so that instead of console errors, we can start sending front-end errors to Sentry...

assembl/static2/js/app/components/debate/common/sentiments.jsx
```diff
@@ -54,22 +80,19 @@ const Sentiment = ({ sentimentCounts, mySentiment, sentiment, client, isSelected
                     post: {
                       id: postId,
                       sentimentCounts: {
-                        like:
-                            sentiment.camelType === 'like'
-                              ? sentimentCounts.like + 1
-                              : sentimentCounts.like - (mySentiment === 'LIKE' ? 1 : 0),
+                        like: sentiment.camelType === 'like' ? likeCount + 1 : likeCount - (mySentiment === 'LIKE' ? 1 : 0),
```
ayyazdaniaryan: These `LIKE`, `DISAGREE`, etc should all become enums in the types file. Understandably, this isn't the scope of this PR.

assembl/static2/js/app/components/debate/survey/post.jsx
```diff
@@ -110,63 +130,107 @@ class Post extends React.Component {
 
   render() {
     const { post } = this.props.data;
-    const { contentLocale, lang, screenWidth, originalLocale } = this.props;
+    if (!post.publicationState || post.publicationState.startsWith('DELETED')) {
+      return null;
```
ayyazdaniaryan: The ticket's requirements were to delete a post for phase1, but this component is used by all phases, is that not so?

assembl/static2/js/app/components/debate/survey/post.jsx
```diff
@@ -110,63 +130,107 @@ class Post extends React.Component {
...
     );
 
-    const { displayName, isDeleted } = post.creator;
+    let creatorName = '';
+    let userCanDeleteThisMessage = false;
+    if (post.creator) {
+      const { displayName, isDeleted } = post.creator;
+      const connectedUserId = getConnectedUserId();
+      userCanDeleteThisMessage =
+        (post.creator && (connectedUserId === String(post.creator.userId) && connectedUserCan(Permissions.DELETE_MY_POST))) ||
```
ayyazdaniaryan: brackets around please ```connectedUserCan(Permissions.DELETE_MY_POST))) || connectedUserCan(Permissions.DELETE_POST);```

ayyazdaniaryan: Also, not that it makes much of a difference, but why not do an ```int(connectedUserId === post.creator.userId``` 

check instead?

assembl/static2/js/app/components/debate/survey/post.jsx
cedricmessiant: not this one (it's for survey only) but we already have post deletion for thread and multicolumn anyway

assembl/static2/js/app/components/debate/common/sentiments.jsx
cedricmessiant: I was about to do it but since those come from back-end and are used in `SentimentTypes` type in `graphql_types.flow.js`, I'm not sure we should do it

assembl/static2/js/app/components/debate/survey/post.jsx
cedricmessiant: about the brackets, we don't want parens around the 2 permission checks, the second one checks that user can delete the message even if he has not created it

cedricmessiant: about the cast, in fact I took the code from `postActions.jsx`. I guess that the only difference would be that if `connectedUserId` is `undefined`, `Number(connectedUserId)` would return `NaN` (I don't know if it can happen)

assembl/static2/js/app/components/debate/common/sentiments.jsx
cedricmessiant: yeah, I added lots of flow types that were out of the scope of the US so I skiped some. I will try to do this one ;-)

cedricmessiant: done !

PR #452, opened 2018-06-11
Use uncontrolled component with key pattern for RichTextEditor

cedricmessiant:
Fixes DEVAS-1581
I'm opened to discussion on this one. I recommend that you read https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html (an article from a few days ago that made me chose this solution over continuing to use `componentWillReceiveProps` or `getDerivedStateFromProps`)


vincentfretin:
LGTM

PR #454, opened 2018-06-12
Hotfix/whitelist

ayyazdaniaryan:
Allow for optional whitelisting on SSO's, and blocking registration on whitelisting otherwise.

### DIFF
--- a/assembl/models/preferences.py
+++ b/assembl/models/preferences.py
@@ -757,8 +757,8 @@ class Preferences(MutableMapping, Base, NamedClassMixin):
 
         # Allow whitelist to be applied to the regular login process
         {
-            "id": "whitelist_on_login",
-            "name": _("Whitelist on default login"),
+            "id": "whitelist_on_register",
+            "name": _("Whitelist on standard registration"),
             "value_type": "bool",
             # "scalar_values": {value: "label"},
             "description": _(
--- a/assembl/models/preferences.py
+++ b/assembl/models/preferences.py
@@ -740,6 +740,52 @@ class Preferences(MutableMapping, Base, NamedClassMixin):
             "item_default": ""
         },
 
+        # Allow whitelist to be applied to SSO login process
+        {
+            "id": "whitelist_on_authentication_backend",
+            "name": _("Whitelist on authentication service"),
+            "value_type": "bool",
+            # "scalar_values": {value: "label"},
+            "description": _(
+                "Allow white list to be applied to the authentication service chosen"),
+            "allow_user_override": None,
+            "modification_permission": P_ADMIN_DISC,
+            # "frontend_validator_function": func_name...?,
+            # "backend_validator_function": func_name...?,
+            "default": False
+        },
+
+        # Allow whitelist to be applied to the regular login process
+        {
+            "id": "whitelist_on_login",
+            "name": _("Whitelist on default login"),
+            "value_type": "bool",
+            # "scalar_values": {value: "label"},
+            "description": _(
+                "Allow white list to be applied to the default login process"),
+            "allow_user_override": None,
+            "modification_permission": P_ADMIN_DISC,
+            # "frontend_validator_function": func_name...?,
+            # "backend_validator_function": func_name...?,
+            "default": False
+        },
+
+        # A discussion administrator, if different from the server administrator
+        {
+            "id": "discussion_administrators",
+            "name": _("Discussion administrators"),
+            "value_type": "list_of_email",
+            # "scalar_values": {value: "label"},
+            "description": _(
+                "A list of discussion administrators, if different from the server administrator"),
+            "allow_user_override": None,
+            "modification_permission": P_ADMIN_DISC,
+            # "frontend_validator_function": func_name...?,
+            # "backend_validator_function": func_name...?,
+            "default": False
+        },
+
+
         # Show the CI Dashboard in the panel group window
         {
             "id": "show_ci_dashboard",
--- a/assembl/models/preferences.py
+++ b/assembl/models/preferences.py
@@ -767,7 +767,7 @@ class Preferences(MutableMapping, Base, NamedClassMixin):
             "modification_permission": P_ADMIN_DISC,
             # "frontend_validator_function": func_name...?,
             # "backend_validator_function": func_name...?,
-            "default": False
+            "default": True
         },
 
         # A discussion administrator, if different from the server administrator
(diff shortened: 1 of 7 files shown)


assembl/models/preferences.py
```diff
@@ -740,6 +740,52 @@ def get_preference_data_list(self):
...
+            "allow_user_override": None,
+            "modification_permission": P_ADMIN_DISC,
+            # "frontend_validator_function": func_name...?,
+            # "backend_validator_function": func_name...?,
+            "default": False
+        },
+
+        # Allow whitelist to be applied to the regular login process
+        {
+            "id": "whitelist_on_login",
```
maparent: Should be whitelist_on_register

assembl/models/preferences.py
```diff
@@ -740,6 +740,52 @@ def get_preference_data_list(self):
...
+            "name": _("Whitelist on default login"),
+            "value_type": "bool",
+            # "scalar_values": {value: "label"},
+            "description": _(
+                "Allow white list to be applied to the default login process"),
+            "allow_user_override": None,
+            "modification_permission": P_ADMIN_DISC,
+            # "frontend_validator_function": func_name...?,
+            # "backend_validator_function": func_name...?,
+            "default": False
```
maparent: are you sure default should be False?

assembl/views/api2/auth.py
```diff
@@ -546,7 +546,7 @@ def assembl_register_user(request):
                 print "email token:", request.route_url(
                     'user_confirm_email', token=email_token(account))
             if discussion:
-                check_subscription = discussion.preferences['whitelist_on_login']
+                check_subscription = discussion.preferences['whitelist_on_register'] or False
```
maparent: default does not go here, but in prefs where I had original line note

assembl/models/preferences.py
ayyazdaniaryan: This should be false. We don't want to check whitelist by default.

assembl/views/api2/auth.py
ayyazdaniaryan: Ahh when you set a default, does the dict always get populated ?

maparent: yes

ayyazdaniaryan: Fixed on fcef653

maparent: more accurately: if not defined in the json, the getters cascade to the default value.

maparent:
LGTM

assembl/models/preferences.py
maparent: Btw, I did not insist, it's your choice, but an empty whitelist is not checked.
I'm not sure in what case you'd want to populate whitelist and leave both those prefs at their default of false, because it adds an operation, and hence an opportunity for making a mistake.
The most common case (with non-empty whitelist) will be that "whitelist_on_register" will be True, and "whitelist_on_authentication_backend" will be False, and this common case does not interfere with normal operations if the whitelist is left empty.

assembl/static/js/app/views/preferencesView.js
```diff
@@ -540,7 +540,7 @@ var EmailPreferenceView = StringPreferenceView.extend({
   constructor: function EmailPreferenceView() {
     StringPreferenceView.apply(this, arguments);
   },
-  regexp: new RegExp("^[A-Z0-9._%+-]+@[A-Z0-9.-]+\.[A-Z]{2,}$"),
+  regexp: /^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/,
```
maparent: maybe put the source of that regexp in a comment? I know I did not do so myself, sorry.

assembl/views/api2/auth.py
```diff
@@ -490,6 +484,14 @@ def assembl_register_user(request):
     if errors:
         raise errors
 
+    # This logic needs to be above the JSONError checks to ensure that whitelisting is applied
+    # even if the discusison does not have a P_SELF_REGISTER on system.Everyone
```
maparent: typo: discusision->discussion.
I'm trying to understand what I was doing: if the person was not allowed to self-register, I would let them register generically. I'm not 100% that was the right behaviour to start with, maybe I should just have errored.
OTH, what you're doing means that, if forbidden by a non-empty whitelist, they just fail instead of registering generically. I'm quite sure that's the right call, either way.

assembl/static/js/app/views/preferencesView.js
```diff
@@ -523,7 +523,7 @@ var UrlPreferenceView = StringPreferenceView.extend({
   },
   regexp: new RegExp('^(?:(?:http|https|ftp)://)(?:\\S+(?::\\S*)?@)?(?:(?:(?:[1-9]\\d?|1\\d\\d|2[01]\\d|22[0-3])(?:\\.(?:1?\\d{1,2}|2[0-4]\\d|25[0-5])){2}(?:\\.(?:[0-9]\\d?|1\\d\\d|2[0-4]\\d|25[0-4]))|(?:(?:[a-z\\u00a1-\\uffff0-9]+-?)*[a-z\\u00a1-\\uffff0-9]+)(?:\\.(?:[a-z\\u00a1-\\uffff0-9]+-?)*[a-z\\u00a1-\\uffff0-9]+)*(?:\\.(?:[a-z\\u00a1-\\uffff]{2,})))|localhost)(?::\\d{2,5})?(?:(/|\\?|#)[^\\s]*)?$', 'i'),
   processValue: function(value) {
-    if (!this.regexp.test(value)) {
+    if (!this.regexp.test(value.trim())) {
         throw i18n.gettext("This does not appear to be a URL");
```
maparent: should we trim the value outside the test? (ie before assignment.) It might make sense in many cases (url, email, int...) and maybe even all cases.

maparent:
made a few more comments, I'm feeling more strongly about the default, but exercise your judgement and merge. No need for a third review.

assembl/views/api2/auth.py
ayyazdaniaryan: Yeah, I'm being aggressive with the whitelist on register. I think it's the right call.

assembl/models/preferences.py
ayyazdaniaryan: I see your point. This is a preference, really. But, I'll side with you, as I don't trust administrators to fully know what they're doing. 
This commit should resolve this check now (it didn't before, as you had said). 011c525e

assembl/static/js/app/views/preferencesView.js
ayyazdaniaryan: Done

PR #455, opened 2018-06-13
Migration delete extra fields for deleted users

mohamedzayed92:
When We export the extra fields, even though some users are deleted, their extra fields are still in teh database, which is not compliant with GDPR. This is because  we deployed the delete user feature before adding deletion of the extra fields. this migration is to delete all extra fields for all deleted users.

### DIFF
--- a/assembl/alembic/versions/7bf35c9261d9_delete_extra_fields_for_deleted_users.py
+++ b/assembl/alembic/versions/7bf35c9261d9_delete_extra_fields_for_deleted_users.py
@@ -1,0 +1,36 @@
+"""delete_extra_fields_for_deleted_users
+
+Revision ID: 7bf35c9261d9
+Revises: e3f681bd70e0
+Create Date: 2018-06-13 14:14:41.037265
+
+"""
+
+# revision identifiers, used by Alembic.
+revision = '7bf35c9261d9'
+down_revision = 'e3f681bd70e0'
+
+from alembic import context, op
+import sqlalchemy as sa
+import transaction
+
+
+from assembl.lib import config
+
+
+def upgrade(pyramid_env):
+    with context.begin_transaction():
+        pass
+
+    # Do stuff with the app's models here.
+    from assembl import models as m
+    db = m.get_session_maker()()
+    with transaction.manager:
+        extra_fields = db.query(m.ProfileField).join(m.User).filter(m.User.is_deleted).all()
+        for ef in extra_fields:
+            db.delete_extra_field
+
+
+def downgrade(pyramid_env):
+    with context.begin_transaction():
+        pass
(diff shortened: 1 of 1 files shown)


ayyazdaniaryan:
This is a comment

ayyazdaniaryan:
This is another test.

ayyazdaniaryan:
Okay, this one should be rebased and be merged into develop as well. Before merging, please make sure that the `delete` operation only deletes the `ProfileFields`, and not the joined table with users (should be the case, but please test). If all good, this is ready for merge as well.

assembl/alembic/versions/7bf35c9261d9_delete_extra_fields_for_deleted_users.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+from assembl.lib import config
+
+
+def upgrade(pyramid_env):
+    from assembl import models as m
+    db = m.get_session_maker()()
+    with transaction.manager:
+        db.query(m.ProfileField).join(m.User).filter(m.User.is_deleted).delete()
```
ayyazdaniaryan: I would be explicit here `m.User.is_deleted == True`

ayyazdaniaryan: I'm not sure, maybe @vincentfretin can chime in, but confirm that this doesn't delete the joined table rows. If so, you can get the list of ProfileField id's, and delete each id.

vincentfretin: Putting `.delete()` at the end will generate a DELETE query like it should. If `.all()` returns ProfileField objects, then `.delete()` will remove exactly those objects.

vincentfretin: `m.User.is_deleted == True` is better yes.

assembl/alembic/versions/7bf35c9261d9_delete_extra_fields_for_deleted_users.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+def upgrade(pyramid_env):
+    from assembl import models as m
+    db = m.get_session_maker()()
+    with transaction.manager:
+	# SQL Alchemy can't delete after a join operation, so we are getting the ids first
+	# and deleting the corresponding profile fields
+        ids=db.query(m.ProfileField.id).join(m.User).filter(m.User.is_deleted==True).all()
+        ids = [id for (id,) in ids]
```
ayyazdaniaryan:
```
ids_query=db.query(m.ProfileField.id).join(m.User).filter(m.User.is_deleted==True)
for (id, _) in ids_query:
   profile = m.ProfileField.get(id)
   db.delete(profile)
```

ayyazdaniaryan: You don't need to check for existence of the profile, as you're getting the ids right from the source (the db). Just issue a delete statement on the object itself. And you're good to go. Sad that the delete on join doesn't work :(

PR #456, opened 2018-06-14
Fix Save button on admin, move it in the DOM via a React portal

cedricmessiant:
I have to find a way to fix the tests...
I tried to get inspiration from this: https://gist.github.com/505aaron/d5efc2ecc622306cbcc6d3e9c1d7ee9f but the ref is not easy to manage with

### DIFF
--- a/assembl/static2/css/views/administration.scss
+++ b/assembl/static2/css/views/administration.scss
@@ -350,13 +350,6 @@
     z-index: 100;
   }
 
-  .save-button {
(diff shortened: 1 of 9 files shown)


cedricmessiant:
Portals are not this powerful (yet?). Children components are mounted before parent, that's why when we reach `SaveButton`, the ref to `#save-button` is `null`. It is the same if we use `document.getElementById('save-button')`, we need to use another approach.

PR #459, opened 2018-06-21
Admin landing page timeline

paolina92:


assembl/static2/js/app/utils/translations.js
```diff
@@ -1371,7 +1386,9 @@ const Translations = {
         headerSubtitle: "Header subtitle",
         instructionsTitle: "Instructions title",
         instructionsContent: "Instructions content",
-        propositionSectionTitle: "Section title"
+        propositionSectionTitle: "Section title",
+        propositionSectionSubtitle: "Section subtitle",
+        descriptionPhase: "Description phase"
```
ayyazdaniaryan: Phase description

assembl/static2/js/app/app.jsx
```diff
@@ -23,28 +28,44 @@ class App extends React.Component {
...
+      const currentPhaseIdentifier = getCurrentPhaseIdentifier(timeline);
       browserHistory.push(get('debate', { slug: params.slug, phase: currentPhaseIdentifier }));
     }
   }
 
   render() {
+    const { timelineLoading, timeline, putTimelineInStore } = this.props;
     const { debateData, debateLoading, debateError } = this.props.debate;
     const divClassNames = classNames('app', { 'harvesting-mode-on': this.props.isHarvesting });
+    if (!timelineLoading) {
```
cedricmessiant: as you use `withLoadingIndicator` you probably never get here with `timelineLoading === true`. If it happens, there is a problem

assembl/static2/js/app/app.jsx
```diff
@@ -55,7 +76,37 @@ const mapDispatchToProps = dispatch => ({
   },
   addContext: (path, debateId, connectedUserId, connectedUserName) => {
     dispatch(addContext(path, debateId, connectedUserId, connectedUserName));
+  },
+  putTimelineInStore: (timeline) => {
```
cedricmessiant: why do you need timeline in store ? I think you should just use the query each time you need it (or even better, put it in the new context API and use some sort of `<TimelineProvider>` component and we should do the same for everything we need in multiple different places.

Context API doc: https://reactjs.org/blog/2018/03/29/react-v-16-3.html

assembl/static2/js/app/main.jsx
```diff
@@ -4,17 +4,13 @@ import * as React from 'react';
...
 import { getCurrentPhaseIdentifier, type Timeline } from './utils/timeline';
-import { addRedirectionToV1 } from './actions/phaseActions';
 import Navbar from './components/navbar/navbar';
 import Footer from './components/common/footer';
 
-type Debate = { debateData: { timeline: Timeline } };
-
 type Props = {
-  debate: Debate,
+  timeline: Timeline,
```
cedricmessiant: I think you could (should?) use the type from flow (generate it with `apollo-codegen` (see `flow/README` for details)

assembl/static2/js/app/main.jsx
```diff
@@ -23,33 +19,19 @@ type State = { identifier: string, location: string };
...
-      // timeline is not configured
-      isRedirectionToV1 = true;
-    } else {
-      const currentPhase = debateData.timeline.filter(phase => phase.identifier === currentPhaseIdentifier);
-      isRedirectionToV1 = currentPhase[0] && currentPhase[0].interface_v1;
-    }
-    this.props.addRedirectionToV1(isRedirectionToV1);
-  }
-
   componentWillReceiveProps(nextProps) {
```
cedricmessiant: `componentWillReceiveProps` will be deprecated, it could be a good occasion to change it here

assembl/static2/js/app/app.jsx
cedricmessiant: Sorry, this is the real doc: https://reactjs.org/docs/context.html

assembl/static2/js/app/app.jsx
paolina92: yes, there is a problem... I don't know why. I'll look at it again

PR #465, opened 2018-07-03
Fix: use a portal to place save button out of the form

cedricmessiant:
Don't ask me why, but this version works ;-)
(Inspired by https://codesandbox.io/s/j3qry8r38y)


cedricmessiant:
Thanks!

PR #470, opened 2018-07-10
Document graphql

ayyazdaniaryan:
This shows the new documentation of our GraphQL APIs. They're not perfect, but they were a lot of APIs to report and document on. From now on, any graphQL endpoint that is created should have documentation set something like this. We've tried to collate everything in one place. Open to changes, and improvements as the come.


assembl/graphql/docstrings.py
```diff
@@ -0,0 +1,961 @@
...
+    discussion = """The discussion object metadata."""
+    landing_page_module_types = """The metadata object for LandingPageModule object."""
+    landing_page_modules = """A list of LandingPageModules."""
+    text_fields = """A list of ConfigurableField union, where each text field represents a field on a bound entity."""
+    profile_fields = """A list of ConfigurableField union, where each text field represents a field on a profile only."""
+    timeline = """A list of DiscussionPhase objects, descriping the timeline objects on the debate."""
+
+
+class Discussion:
+    __docs__ = """The Discussion object. This object describes certain parts of the core state of the debate."""
```
vincentfretin: you wrote `__docs__` instead of `__doc__` on purpose?

assembl/graphql/docstrings.py
```diff
@@ -0,0 +1,961 @@
...
+    homepage_url = """A URL for the homepage (optional). Often placed on the logo."""
+    title = """The title of the discussion, in the language specified by the input"""
+    subtitle = """The subtitle of the discussion, in the language specified by the input"""
+    button_label = """The value inside of the participation button in the landing page."""
+    header_image = Default.document % ("The file representing the header of the landing page. ", )
+    logo_image = Default.document % ("The file representing the logo of the debate. ", )
+
+
+class UpdateDiscussion:
+    __doc__ = """The mutation that allows to update an existing Discussion object"""
```
vincentfretin: You can just write the docstring on the class, without `__doc__ = `. The string will be available on the `__doc__` class attribute.

```
>>> class A(object):
...   """lalala"""
... 
>>> A.__doc__
'lalala'
```

assembl/graphql/landing_page.py
```diff
@@ -32,18 +34,24 @@ def resolve_title_entries(self, args, context, info):
...
     module_type = graphene.Field(LandingPageModuleType)
     exists_in_database = graphene.Boolean()
     title = graphene.String(lang=graphene.String())
     title_entries = graphene.List(LangStringEntry)
     subtitle = graphene.String(lang=graphene.String())
     subtitle_entries = graphene.List(LangStringEntry)
+=======
+    module_type = graphene.Field(LandingPageModuleType, description=docs.LandingPageModule.module_type)
+    exists_in_database = graphene.Boolean(description=docs.LandingPageModule.exists_in_database)
+>>>>>>> Update landing page docstrings
```
vincentfretin: You seems to have left a conflict here.

assembl/graphql/post.py
```diff
@@ -73,39 +78,41 @@ def resolve_creator(self, args, context, info):
...
     def resolve_extracts(self, args, context, info):
         return self.db.query(models.Extract
-            ).join(models.Content, models.Extract.content == self
-            ).options(joinedload(models.Extract.text_fragment_identifiers)
-            ).order_by(models.Extract.creation_date
-            ).all()
+                             ).join(models.Content, models.Extract.content == self
+                                    ).options(joinedload(models.Extract.text_fragment_identifiers)
+                                              ).order_by(models.Extract.creation_date
+                                                         ).all()
```
vincentfretin: I don't like this change.

assembl/graphql/landing_page.py
ayyazdaniaryan: Oops, that must have been left from the rebase. Fixing now.

assembl/graphql/post.py
ayyazdaniaryan: It's an auto-pep8 service that does this on another dev's machine.

assembl/graphql/docstrings.py
ayyazdaniaryan: Nope. error, good catch.

assembl/graphql/post.py
vincentfretin: You should configure a pep8 config to not allow this indentation. There is no issue with flake8.

assembl/graphql/docstrings.py
ayyazdaniaryan: Yes, absolutely. The reason for this was to keep the editable string of text for document in one place as we one-shot this. Also, for future, any non-dev or front-end devs can edit documentation without python knowledge.

PR #472, opened 2018-07-12
avoid passing down the Tree isHarvesting, use the context api

vincentfretin:


assembl/static2/css/components/harvesting.scss
```diff
@@ -52,50 +52,11 @@
   }
 }
 
-.extracts.extracts--is-harvesting {
```
vincentfretin: those rules weren't used at all

assembl/static2/css/components/posts.scss
```diff
@@ -357,13 +357,10 @@ $deleted-post-border: 1px solid $grey1 !default;
 
   .extracts {
     position: absolute;
-
-    &.extracts--is-not-harvesting {
```
vincentfretin: The extracts--is-not-harvesting class was always present.

assembl/static2/js/app/components/debate/thread/nuggets.jsx
```diff
@@ -57,7 +56,7 @@ class Nuggets extends React.Component {
               ref={(node) => {
                 this.node = node;
               }}
-              className={classnames('extracts', 'extracts--is-not-harvesting', { hidden: isHarvesting })}
```
vincentfretin: I'm simply using `.harvesting-mode-on .extracts { display: none }` now.

PR #473, opened 2018-07-17
run graphql tests with the frontend tests to avoid the 50 min limit

vincentfretin:


ayyazdaniaryan:
It's just a workaround, but it'll stop travis from failing everytime for now. There is a serious issue with how our fixtures are loaded and unloaded per test, and that's the real fix. If you'd like to merge this until the fix is done, fine by me.

PR #481, opened 2018-07-29
add locale and locale labels in bulk

maparent:
speeds up startup.


vincentfretin:
LGTM

assembl/models/langstrings.py
```diff
@@ -267,14 +284,31 @@ def get_or_create(cls, locale_code, db=None):
...
         db.flush()
         cls._locale_uncommitted.append(locale)
         cls.reset_cache()
         locale_object_cache[locale_code] = locale
         return locale
 
+    @classmethod
+    def bulk_get_or_create(cls, locale_codes, db=None):
+        db = db or cls.default_db
+        db.execute("lock table %s in exclusive mode" % cls.__table__.name)
```
ayyazdaniaryan: Not sure if this is accurate, but should the share mode be added to this statement for greater control over the concurrent change of locales? https://www.postgresql.org/docs/9.4/static/sql-lock.html

assembl/models/langstrings.py
```diff
@@ -66,6 +66,23 @@ def ancestry(self):
                 break
             self = self.get_or_create(ancestor)
 
+    @staticmethod
+    def is_right_to_left_code(code):
+        # Good enough approximation
```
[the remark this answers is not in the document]
maparent: I am looking forward to the HTML tag that indicates languages traversing the 4th dimension, as Octopod in Arrival.

assembl/models/langstrings.py
maparent: you mean share instead of exclusive? (It would replace, not add. There is no share exclusive.)
Doc is weird; it says share precludes concurrent data access, yet it does not conflict with the row share option, i.e. another update. I think exclusive is safer, and not that much harder to obtain (unless alembic is running, which would be weird.)

PR #485, opened 2018-08-06
DEVAS1711: filter out brightmirror posts from elasticsearch

cedrtang:
Filter when insert/update/delete and when reindexing with fab command. Doesn't break indexing of other posts.

**Note: I couldn't filter out the posts in "get_indexable_contents" as Aliases classes are used and not all fields/methods are available for filtering but it's done later by "get_data" in utils.py

### DIFF
--- a/assembl/indexing/utils.py
+++ b/assembl/indexing/utils.py
@@ -134,6 +134,10 @@ def get_data(content):
         if not data['idea_id']:
             return None, None
 
+        # If the post is a fiction for Bright Mirror, don't index it.
+        if Idea.get(data['idea_id'][0]).message_view_override == 'brightMirror':
+            return None, None
+
         # If the idea is now in multi columns mode and the post was created before that, don't index it.
         if Idea.get(data['idea_id'][0]).message_columns and content.message_classifier is None:
             return None, None
--- a/assembl/indexing/utils.py
+++ b/assembl/indexing/utils.py
@@ -118,7 +118,7 @@ def get_data(content):
         return get_uid(content), data
 
     elif isinstance(content, Post):
-        from assembl.models.timeline import Phases
+        from assembl.models.idea import MessageView
         data = {}
         data['_parent'] = 'user:{}'.format(content.creator_id)
         if content.parent_id is not None:
@@ -136,7 +136,7 @@ def get_data(content):
             return None, None
 
         # If the post is a fiction for Bright Mirror, don't index it.
-        if Idea.get(data['idea_id'][0]).message_view_override == Phases.brightMirror.value:
+        if Idea.get(data['idea_id'][0]).message_view_override == MessageView.brightMirror.value:
             return None, None
 
         # If the idea is now in multi columns mode and the post was created before that, don't index it.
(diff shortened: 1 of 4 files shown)


assembl/models/post.py
```diff
@@ -560,6 +560,10 @@ def restrict_to_owners(cls, query, user_id):
         "filter query according to object owners"
         return query.filter(cls.creator_id == user_id)
 
+    def is_bright_mirror_fiction(self):
+        parent_idea = self.indirect_idea_content_links_without_cache()
```
vincentfretin: You need to filter on `IdeaRelatedPostLink` only, so `parent_idea = [link.idea_id for link in content.indirect_idea_content_links_without_cache() if link.__class__.__name__ == 'IdeaRelatedPostLink']`

cedrtang:
@vincentfretin Changement fait!

vincentfretin:
LGTM

assembl/indexing/utils.py
```diff
@@ -134,6 +134,10 @@ def get_data(content):
         if not data['idea_id']:
             return None, None
 
+        # If the post is a fiction for Bright Mirror, don't index it.
+        if Idea.get(data['idea_id'][0]).message_view_override == 'brightMirror':
```
ayyazdaniaryan:
```from assembl.models.timeline import Phases
if Idea.get(data['idea_id'][0]).message_view_override == Phases.brightMirror.value:
...
```

assembl/models/post.py
```diff
@@ -560,6 +560,10 @@ def restrict_to_owners(cls, query, user_id):
         "filter query according to object owners"
         return query.filter(cls.creator_id == user_id)
 
+    def is_bright_mirror_fiction(self):
+        parent_idea = [link.idea for link in self.indirect_idea_content_links_without_cache() if link.__class__.__name__ == 'IdeaRelatedPostLink']
+        return len(parent_idea) != 0 and parent_idea[0].message_view_override == 'brightMirror'
```
ayyazdaniaryan: Check against the `Phases` enum type. Keeping things DRY.

vincentfretin: No, don't check against phase, only on message_view_override. We can have in the future a thread phase with a single idea in brightMirror mode. If you mean using the Phases enum instead of 'brightMirror', this is wrong, you should create another enum with only multiColumns and brightMirror as possible values.

assembl/indexing/utils.py
vincentfretin: I didn't read this one first, no this is wrong, you shouldn't use the Phases enum for message_view_override, even if the string is the same here. Create a new enum for message_view_override in this case.

assembl/models/post.py
ayyazdaniaryan: Yeah sure, why not.

```
brightMirrorables = {
   Phases.brightMirror,
   Phases.multiColumns
}
```

You can put it in `post.py` and check that the `message_view_override` is in the `brightMirrorables`.

vincentfretin: So create in `models/idea.py` a new enum
```
class MessageView(Enum):
    multiColumns = 'multiColumns'
    brightMirror = 'brightMirror'
```

vincentfretin:
:+1:

PR #486, opened 2018-08-06
Don't show harvesting on small screens

cedricmessiant:
I had to do some refactoring before to do the task (distinct commits)
Refs DEVAS-1502

### DIFF
--- a/assembl/static2/js/app/utils/globalFunctions.js
+++ b/assembl/static2/js/app/utils/globalFunctions.js
@@ -1,6 +1,9 @@
 // @flow
 import { type Map } from 'immutable';
 
+import { getDisplayedPhaseIdentifier } from './timeline';
+import { HARVESTABLE_PHASES } from '../constants';
+
 const getInputValue = (id: string) => {
   const elem = document.getElementById(id);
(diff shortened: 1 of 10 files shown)


assembl/static2/js/app/utils/globalFunctions.js
```diff
@@ -246,4 +249,8 @@ export const moveItemDown = (itemsById: ItemsById, id: string): ItemsById => {
     order += 1;
   });
   return newItemsById;
-};
\ No newline at end of file
+};
+
+export function isHarvestable(params: RouterParams) {
```
amensouissi: I prefer isHarvestablePhase instead of isHarvestable

cedricmessiant: We discussed it with Vincent. We chose `isHarvestable` because in the future, we will have to change this function to use the info on the idea to know if it is an idea from thread/multicolumns (we will have phases with ideas of different types (survey/thread/votes))

MilanBarande: Indeed, we might not talk about phases anymore in the long term

PR #488, opened 2018-08-06
add confirmation modal for text & multimedia additions

MilanBarande:
Add a confirmation modal in the landing page administration when a user wants to add a Text & Multimedia section

### DIFF
--- a/assembl/static2/js/app/components/administration/landingPage/manageModules.jsx
+++ b/assembl/static2/js/app/components/administration/landingPage/manageModules.jsx
@@ -4,7 +4,7 @@ import { connect } from 'react-redux';
 import { I18n, Translate } from 'react-redux-i18n';
 import type { List, Map } from 'immutable';
 import { Button } from 'react-bootstrap';
-
+import { displayModal, closeModal } from '../../../utils/utilityManager';
 import ModulesPreview from './modulesPreview';
 import SectionTitle from '../../administration/sectionTitle';
 import SelectModulesForm from './selectModulesForm';
@@ -26,7 +26,6 @@ type Props = {
   toggleModule: Function,
   createModule: Function
 };
-
 export const DumbManageModules = ({
   enabledModules,
   moduleTypes,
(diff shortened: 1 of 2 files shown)


assembl/static2/js/app/components/administration/landingPage/manageModules.jsx
```diff
@@ -36,33 +35,55 @@ export const DumbManageModules = ({
...
-        <Translate value="administration.landingPage.manageModules.helper" />
-      </p>
-      <div className="two-columns-admin">
-        <div className="column-left">
-          <SelectModulesForm lang={locale} moduleTypes={moduleTypes} modulesById={modulesById} toggleModule={toggleModule} />
-          <div className="margin-xl">
-            <Button className="button-submit button-dark" onClick={() => createModule(enabledModules.size - 2)}>
-              <Translate value="administration.landingPage.manageModules.textAndMultimediaBtn" />
-            </Button>
+}: Props) => {
```
ayyazdaniaryan: Recently, after speaking with @pascalhuynh regarding making components easier to create/mock. As is, this component's `submitButton` is baked in and cannot be changed. So perhaps you could pull the `displayConfirmationModal` out of the definition of the component, and pass it in as a prop, like `handleSubmitButton`. And the `displayConfirmationModal` method can take the component props as input and pick up what it needs internally. It's a better way to mock/test these components in future.

assembl/static2/js/app/utils/translations.js
```diff
@@ -707,7 +707,8 @@ const Translations = {
           title: "Administrer les modules",
           helper: "Choisissez les modules que vous souhaitez voir apparaître dans la page d'accueil et leurs emplacements.",
           textAndMultimedia: "Texte & Multi-média",
-          textAndMultimediaBtn: "Ajouter un module Texte & Multi-media"
+          textAndMultimediaBtn: "Ajouter un module Texte & Multi-media",
+          confirmationModal: "Voulez vous ajouter un module Texte & Multimédia supplémentaire à la landing page ?"
```
vincentfretin: `Voulez-vous` avec un tiret. :) https://leconjugueur.lefigaro.fr/frtraitunionquestion.php

assembl/static2/js/app/components/administration/landingPage/manageModules.jsx
amensouissi: I think that it's a specific business component and not abstract (ui or business component). We don't need to reuse it. in my opinion the solution is sufficient.

vincentfretin:
you can merge

PR #490, opened 2018-08-10
Extract hash

amensouissi:


assembl/alembic/versions/180022eb0f0d_add_extract_hash.py
```diff
@@ -0,0 +1,37 @@
...
+import sqlalchemy as sa
+import transaction
+
+
+def upgrade(pyramid_env):
+    with context.begin_transaction():
+        op.add_column(
+            'extract',
+            sa.Column(
+                'extract_hash', sa.String,
```
vincentfretin: Why do you prefix the field by `extract_`, there is no need.

assembl/alembic/versions/180022eb0f0d_add_extract_hash.py
```diff
@@ -0,0 +1,37 @@
...
+                nullable=False, server_default=''))
+
+    from assembl import models as m
+    db = m.get_session_maker()()
+    with transaction.manager:
+        model = m.Extract
+        query = db.query(model)
+        for extract in query:
+        	extract.update_hash()
```
vincentfretin: You used a tab instead of spaces here?

assembl/graphql/post.py
```diff
@@ -753,25 +755,37 @@ def mutate(root, args, context, info):
...
-            post.db.flush()
+            db = post.db
+            extract_hash = models.Extract.get_extract_hash(
+                extract.get('lang'),
+                extract.get('xpath_start'),
+                extract.get('xpath_end'),
+                extract.get('offset_start'),
+                extract.get('offset_end'),
+                post_id,
+                extract_nature)
```
vincentfretin: Are you sure about extract_nature, shouldn't it be extract_nature.value? It will maybe give another hash on app restart if it's the instance of enum entry.

assembl/lib/utils.py
```diff
@@ -134,3 +135,10 @@ def snake_to_camel(string):
     # We capitalize the first letter of each component except the first one
     # with the 'title' method and join them together.
     return components[0] + "".join(x.title() for x in components[1:])
+
+
+def get_hash(*args):
+    hasher = sha256()
+    hasher.update("".join([str(a) for a in args]))
```
vincentfretin: please use unicode, `u''` and `unicode(a)`

assembl/alembic/versions/180022eb0f0d_add_extract_hash.py
amensouissi: hash is a python keyword

PR #494, opened 2018-08-16
Single celery

maparent:


assembl/lib/model_watcher.py
```diff
@@ -67,45 +72,112 @@ def processAccountModified(self, id):
         pass
 
 
-class ModelEventWatcherPrinter(object):
+class BaseModelEventWatcher(object):
+    """A dummy :py:class:`IModelEventWatcher` for testing purposes"""
+    interface.implements(IModelEventWatcher)
```
vincentfretin: This is not compatible python 3. You should use `@implementer(IModelEventWatcher)` on the class.

ayyazdaniaryan:
@vincentfretin Yes, you would. There is a process (and I don't believe it's documented anywhere) were you delete ALL `NotificationSubscriptions` on the server. And you remove existing Notification objects in case the status is some is still pending. This process was done only once for dev-staging server, and from which, all local databases are based on.

vincentfretin:
We sometime take a real prod dump, I did it once to fix an issue we had only on prod to work on it locally.
We definitely should have a script we can execute to remove NotificationSubscriptions and pending notifications like you said.

maparent:
Good point on prod dump. It has not been an issue on our (Mac) machines because we don't have an active sendmail, but it would be an issue on a linux dev machine. Will look into it.

maparent:
Ok. Added a parameter to develop.rc so the notify task will obsolete the notification instead of sending it. I think that's the best solution.

maparent:
Is that a merge approval?

PR #496, opened 2018-08-20
Devas1684 post fiction

cedrtang:
Reuse of richtext form component for BrightMirror function. Passing custom labels to the component.

### DIFF
--- a/assembl/static2/tests/unit/components/debate/brightMirror/instructionView.spec.jsx
+++ b/assembl/static2/tests/unit/components/debate/brightMirror/instructionView.spec.jsx
@@ -36,7 +36,7 @@ describe('Bright Mirror page', () => {
     expect(result).toMatchSnapshot();
   });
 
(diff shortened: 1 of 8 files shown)


assembl/static2/tests/unit/components/debate/brightMirror/instructionView.spec.jsx
```diff
@@ -1,17 +1,121 @@
...
+      timeline: timeline,
+      identifier: 'brightMirror'
+    };
+    const shallowRenderer = new ShallowRenderer();
+    shallowRenderer.render(<InstructionView {...props} />);
+    const result = shallowRenderer.getRenderOutput();
+    expect(result).toMatchSnapshot();
+  });
+
+  it('timeline is over', () => {
```
MilanBarande: what happens when timeline is over ? I would write something like "should tell the user if the timeline is over"

PR #497, opened 2018-08-21
Use discussion phase id devas 1750

amensouissi:


assembl/static2/js/app/utils/routeMap.js
```diff
@@ -15,7 +15,14 @@ const convertToContextualName = (name) => {
   return base + workingName;
 };
 const maybePrependSlash = (pre, s) => (pre ? `/${s}` : s);
-export const get = (name, args) => {
+
+export const getQuery = (query) => {
+  if (!query) return '';
+  const entries = Object.keys(query).map(key => `${key}=${query[key]}`);
+  return `?${entries.join('&')}`;
+};
```
vincentfretin: We should maybe start using `query-string` dependency (we already have query-string in the project via react-router/history dependency, but you should add it explicitly in package.json) instead of writing our own function here.

remove this getQuery function and below in `get` function use:
```
import { stringify } from "query-string";

return `${a}?${stringify(query)}`
```

assembl/static2/js/app/utils/timeline.js
```diff
@@ -24,27 +25,32 @@ const dateComparator = (phase1, phase2) => {
...
   }
   // if all phases are closed, take the last closed phase
   const sortedTimeline = [...timeline].sort(dateComparator);
-  return sortedTimeline.length === 0 ? 'thread' : sortedTimeline[sortedTimeline.length - 1].identifier;
+  return sortedTimeline[sortedTimeline.length - 1];
+};
+
+export const getCurrentPhaseIdentifier = (_timeline: Timeline) => {
+  const currentPhase = getCurrentPhase(_timeline);
+  return currentPhase ? currentPhase.identifier : PHASES.thread;
```
vincentfretin: Remove the `getCurrentPhaseIdentifier` function completely if it's not used in the code base anymore.

Like I said in a comment above, you may consider having a function
```
export const getCurrentPhaseData = (_timeline: Timeline) => {
￼   const currentPhase = getCurrentPhase(_timeline);
￼   return currentPhase ? {phase: currentPhase.identifier, phaseId: currentPhase.id} : {phase: PHASES.thread, phaseId: null};
```

assembl/static2/tests/unit/components/debate/common/post/index.spec.jsx
```diff
@@ -47,6 +47,7 @@ export const postProps = {
   id: 'XYZ333',
   ideaId: 'ABC123',
   identifier: 'thread',
+  phaseId: 'phaseFoo',
```
vincentfretin: Shouldn't it be a base64 string?

assembl/tests/views_tests/test_routes.py
```diff
@@ -298,24 +299,26 @@ def test_route_discussion_idea_legacy(discussion, root_post_1, subidea_1,
...
     from assembl.lib.frontend_urls import FrontendUrls
+    from assembl import models
     slug = discussion_with_2_phase_interface_v2.slug
     route = "/%s/posts/%s" % (
         slug, quote_plus(post_related_to_sub_idea_1.uri()))
     print route
     resp = test_app.get(route)
     assert resp.status_int == 303
 
+    thread_phase = test_session.query(models.DiscussionPhase).all()[1]
```
vincentfretin: The order of such query is undefined, thread may not be in second position. You should add order field in your DiscussionPhase in your phases fixture and do an explicit `order_by(models.DiscussionPhase)` here or use your `phases` fixture with `phases['thread']`

vincentfretin:
I don't think you handle the following point in your changes:
- Ensure that shared post urls are not broken (fallback to the first phase with this identifier)

assembl/static2/js/app/pages/administration.jsx
```diff
@@ -342,9 +342,9 @@ export default compose(
   connect(mapStateToProps, mapDispatchToProps),
   graphql(VoteSessionQuery, {
     skip: ({ timeline }) => typeof getPhaseId(timeline, 'voteSession') !== 'string',
-    options: ({ timeline, i18n }) => {
-      const id = timeline ? getPhaseId(timeline, 'voteSession') : null;
-      const discussionPhaseId = id ? atob(id).split(':')[1] : null;
+    options: ({ i18n, location }) => {
+      const phaseId = location.query.phaseId;
+      const discussionPhaseId = phaseId ? atob(phaseId).split(':')[1] : null;
```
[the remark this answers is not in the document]
amensouissi: This code is reviewed. I was inspired by this code. But I agree with you.

assembl/tests/views_tests/test_routes.py
amensouissi: Just filter the query (DiscussionPhase.identifier == 'thread')

vincentfretin:
I use isThematicsTable=True when the root idea for a phase is different from discussion.root_idea. We should show a checkbox isThematicsTable for thread phase only to be able to create a different root idea for a second thread phase. First thread phase still have isThematicsTable=False because we use discussion.root_idea and we need to stay that way to create synthesis in v1 interface for now.

Quid of the multiColumns phase? This is a special phase that filter a thread phase, should we have a relation between a multiColumns and a thread phase to know which thread phase should be filtered from? And we will need a select field listing thread phases for a multiColumns phase in the administration.

vincentfretin:
This isThematicsTable checkbox is not really understandable, it's a pure technical issue. The user won't understand why they want to check it or not. I propose the following for now:
- for the first thread phase, it uses discussion.root_idea
- for a second thread phase, it will create a root idea under discussion.root_idea like we do for survey, brightMirror.
- for mutliColumns phase, it will works only against the first thread phase. So it doesn't make sense to create a second multiColumns phase in this case.

vincentfretin:
What I did finally is just to always create a phase root_idea for any phase except if identifier is thread or multiColumns to keep existing debate working.

I reverted the urls changes, we only have the phase identifier in the url, not the additional phaseId
Phase identifier should be unique, there should be only one phase with the given identifier for a debate, this is not currently enforced.

vincentfretin:
When #507 is merged, I need to rebase this branch to fix usage of getIfPhaseCompletedByIdentifier => getIfPhaseCompletedById in js/app/components/debate/brightMirror/instructionView.js

assembl/static2/js/app/main.jsx
```diff
@@ -1,58 +1,46 @@
...
+      // keep old shared urls working
+      phaseId = getPhaseId(timeline, identifier);
+    }
+    const discussionPhaseId = phaseId ? fromGlobalId(phaseId) : null;
     const children = React.Children.map(this.props.children, child =>
       React.cloneElement(child, {
-        identifier: that.state.identifier
+        identifier: identifier,
+        phaseId: phaseId,
+        discussionPhaseId: discussionPhaseId
```
ayyazdaniaryan: Could make the switch to Context API here, wouldn't that be easier? We are using React 16.3.1, so it's available.

assembl/static2/js/app/pages/administration.jsx
```diff
@@ -311,10 +312,19 @@ const isNotInLandingPageAdmin = isNotInAdminSection('landingPage');
...
-      const discussionPhaseId = id ? atob(id).split(':')[1] : null;
+    skip: ({ location, timeline }) => {
+      if (!location.query.phaseId) {
+        return true;
+      }
+      const phase = getPhaseById(timeline, location.query.phaseId);
+      if (!phase) {
+        return true;
+      }
+      return phase.identifier !== 'voteSession';
```
ayyazdaniaryan: These phase names have constants now. Use it here.

assembl/tests/fixtures/graphql.py
```diff
@@ -28,12 +30,97 @@ def graphql_participant1_request(request, test_participant1_webrequest, discussi
...
+
+
+@pytest.fixture(scope="function")
+def idea_in_thread_phase(phases, graphql_request):
     from assembl.graphql.schema import Schema as schema
     res = schema.execute(u"""
 mutation myFirstMutation {
     createThematic(
-        identifier: "thread",
+        discussionPhaseId: """+unicode(phases['thread'].id)+u""",
```
ayyazdaniaryan: Is this a pythonic style, or literally `+unicode(phases['thread'].id)+u` ?

assembl/graphql/timeline.py
```diff
@@ -78,6 +79,13 @@ def mutate(root, args, context, info):
         cls = models.DiscussionPhase
         require_cls_permission(CrudPermissions.CREATE, cls, context)
         discussion_id = context.matchdict['discussion_id']
+        discussion = models.Discussion.get(discussion_id)
+        identifier = args.get('identifier')
```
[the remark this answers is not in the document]
vincentfretin: In the near future, it will be a free text.

assembl/graphql/timeline.py
```diff
@@ -136,8 +144,9 @@ def mutate(root, args, context, info):
             if description_entries is not None:
                 update_langstring_from_input_entries(phase, 'description', description_entries)
 
-            phase.identifier = args.get('identifier')
-            phase.is_thematics_table = args.get('is_thematics_table')
+            identifier = args.get('identifier')
+            phase.identifier = identifier
+#            phase.is_thematics_table = is_thematics_table
```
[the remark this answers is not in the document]
vincentfretin: No, I use is_thematics_table to indicate that we need a phase root idea. This boolean shouldn't change.

assembl/static2/js/app/components/debate/navigation/menuTable.jsx
```diff
@@ -6,32 +6,40 @@ import { SurveyTable, IdeasTable } from './tables';
...
+export function prefetchMenuQuery(
+  client: ApolloClient,
+  variables: { identifier: string, discussionPhaseId: string, lang: string }
+) {
   const query = queries[variables.identifier];
   client.query({
     query: query || queries.default,
-    variables: variables
+    variables: { lang: variables.lang, discussionPhaseId: variables.discussionPhaseId }
+    // need the same variables order than the query for deduplication to work
```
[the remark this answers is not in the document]
vincentfretin: In the old version of apollo-client we use, yes. In recent version, it sorts by keyname, so the ket order in the object doesn't matter.

assembl/static2/js/app/index.jsx
```diff
@@ -22,7 +22,14 @@ const renderAssembl = (routes) => {
     <AppContainer>
       <ApolloProvider store={store} client={client}>
         <ScreenDimensionsProvider>
-          <Router history={browserHistory} routes={routes} onUpdate={hashLinkScroll} />
+          <Router
+            history={browserHistory}
+            routes={routes}
+            onUpdate={() => {
+              window.scrollTo({ top: 0, left: 0, behavior: 'smooth' });
```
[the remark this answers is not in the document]
vincentfretin: This is to not stay in the middle of a page when you change route, for example going from thread page to home page. This was previously in the Main.jsx. Putting it here is far better, and fixes some edge cases where we had a scroll to top when whole tree was being rendered in the thread page.

assembl/static2/js/app/main.jsx
vincentfretin: This could be an alternative yes. Maybe next time, if we merge the Main and App components for example.

assembl/tests/fixtures/graphql.py
vincentfretin: Not the best way to write it, but it was far more easier to update the tests like this. It was time consuming enough to adapt all the tests.

assembl/static2/js/app/pages/administration.jsx
vincentfretin: Sure. @amensouissi can you use the constant here please?

assembl/static2/js/app/components/administration/discussion/phaseForm.jsx
```diff
@@ -120,7 +120,11 @@ export const DumbPhaseForm = ({
       <div className="text-xs configure-module-text">
         <Translate value="administration.timelineAdmin.configureModule" />
         <Link
-          to={`${get('administration', slug)}${get('adminPhase', { ...slug, phase: identifier })}?section=1`}
+          to={`${get('administration', slug)}${get(
```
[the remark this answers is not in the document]
vincentfretin: Yes, @amensouissi did it in the other branch. He will take care of it when merging both branches.

assembl/tests/views_tests/test_graphql.py
```diff
@@ -64,7 +64,7 @@ def test_get_thematics_with_video(discussion, phases, graphql_request, test_sess
     video_desc_side = models.LangString.create(
         u"Putain",
         locale_code="fr")
-    root_thematic = create_root_thematic(discussion, "survey")
+    root_thematic = create_root_thematic(phases['survey'])
```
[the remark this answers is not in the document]
vincentfretin: no, it's the sqlalchemy object here

assembl/utils.py
```diff
@@ -54,8 +51,69 @@ def get_ideas(discussion_id, phase_id):
...
+    """
+    # Retrieve the phases with posts
+    identifiers_with_posts = [i for i in identifiers if i in PHASES_WITH_POSTS]
+    if not discussion or not identifiers_with_posts:
+        return None
+
+    ideas = []
+    # If survey phase, we need the root thematic
+    if Phases.survey.value in identifiers_with_posts:
+        survey_phase = get_phase_by_identifier(discussion, Phases.survey.value)
```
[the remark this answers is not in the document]
vincentfretin: I created a ticket in jira for this

assembl/static2/js/app/main.jsx
```diff
@@ -1,62 +1,46 @@
...
-    const discussionPhaseId = fromGlobalId(phaseId);
+    const { params, timeline } = this.props;
+    const { themeId } = params;
+    const { currentPhaseIdentifier, currentPhaseId } = getCurrentPhaseData(timeline);
+    let identifier = params.phase || null;
+    let phaseId = params.phaseId || null;
+    if (!identifier) {
+      identifier = currentPhaseIdentifier;
+      phaseId = currentPhaseId;
+    }
```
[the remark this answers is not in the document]
vincentfretin: no necessary

assembl/tests/indexing/test_utils.py
```diff
@@ -13,7 +13,8 @@ def test_get_data_for_extract(extract_submitted_in_post_related_to_sub_idea_1_1_
     assert data['creator_id'] == participant2_user.id
     assert data['discussion_id'] == extract.discussion.id
     assert data['idea_id'] == [subidea_1_1_1.id]
-    assert data['phase_id'] == u'thread'
+    assert data['phase_id'] == phases['thread'].id
```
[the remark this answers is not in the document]
vincentfretin: No, the test here was actually wrong, phase_id needs here to be the primary key of the DiscussionPhase object.

PR #498, opened 2018-08-22
Harvesting in multi column devas 1503

amensouissi:


assembl/static2/js/app/components/debate/common/post/postView.jsx
```diff
@@ -194,6 +194,7 @@ class PostView extends React.PureComponent<Props, State> {
             setHarvestingBoxDisplay={this.setHarvestingBoxDisplay}
             handleClickAnchor={this.handleClickAnchor}
             cancelHarvesting={this.cancelHarvesting}
+            isMultiColumns={multiColumns}
```
vincentfretin: You call this prop `showNuggetAction`

PR #503, opened 2018-08-29
Devas1730 overview all fictions

cedrtang:
- Add fictions list

- When new fiction is created, list is updated and a popin effect is applied

- When hover, popin effect is applied

PR #509, opened 2018-09-05
refactor cookieToggle to deal with the various cases of cookies

MilanBarande:
This is a quick rearrangement of the cookieToggle component (now cookieSetter) because another requirement was added to this component: to display a message indicating that a cookie is required instead of having a toggle set on Accepted and disabled. I already had a ternary operator in cookieToggle's return so I created a sub component dealing with the 3 cases:

1. The cookie is required: show a message saying so.
2. The cookie is not required but is concerning the matomo settings: show a link sending the user to the matomo optout page.
3. The cookie is a regular cookie to be configured with a switch button: show a switch button. (At this point we have no cookie entering this category but we might in the future)

### DIFF
--- a/assembl/static2/js/app/components/cookies/cookieSetter.jsx
+++ b/assembl/static2/js/app/components/cookies/cookieSetter.jsx
@@ -17,7 +17,7 @@ export type CookieObject = {
   cookieType: string
 };
 
-type cookieSetterProps = {
+type CookieSetterProps = {
   handleToggle: Function,
   cookie: CookieObject,
   toggleCookieType: Function,
@@ -27,7 +27,7 @@ type cookieSetterProps = {
 // @$FlowFixMe globalAnalytics is a global variable
 const matomoHost = globalAnalytics.piwik ? globalAnalytics.piwik.host : null;
 
-const cookieSetter = ({ handleToggle, cookie, toggleCookieType, locale }: cookieSetterProps) => {
+const CookieSetter = ({ handleToggle, cookie, toggleCookieType, locale }: CookieSetterProps) => {
   const toggleSwitch = () => {
     const { accepted, cookieType } = cookie;
     const updatedCookie = { ...cookie, accepted: !accepted, cookieType: toggleCookieType(cookieType) };
@@ -69,4 +69,4 @@ const cookieSetter = ({ handleToggle, cookie, toggleCookieType, locale }: cookie
   );
 };
 
-export default cookieSetter;
+export default CookieSetter;
--- a/assembl/static2/js/app/components/cookies/cookieSetter.jsx
+++ b/assembl/static2/js/app/components/cookies/cookieSetter.jsx
@@ -8,7 +8,7 @@ import classnames from 'classnames';
 import Helper from '../common/helper';
 import CookieSwitch from './cookieSwitch';
 
-import { COOKIE_TRANSLATION_KEYS } from '../../constants';
+import { COOKIE_TRANSLATION_KEYS, COOKIES_CATEGORIES } from '../../constants';
 
 export type CookieObject = {
   name: string,
@@ -37,12 +37,12 @@ const cookieSetter = ({ handleToggle, cookie, toggleCookieType, locale }: cookie
   const { name, category, accepted } = cookie;
 
   const cookieName = Object.keys(COOKIE_TRANSLATION_KEYS).includes(name) ? I18n.t(`cookies.${name}`) : name;
-  const cookieIsMatomo = name === 'matomo';
+  const cookieIsMatomo = name === COOKIE_TRANSLATION_KEYS.matomo;
   const matomoOptOutLink = matomoHost ?
     `https://${matomoHost}/index.php?module=CoreAdminHome&action=optOut&language=${locale}`
     : null;
 
-  const required = category === 'essential';
+  const required = category === COOKIES_CATEGORIES.essential;
 
   return (
     <div
--- a/assembl/static2/js/app/components/cookies/cookieSetter.jsx
+++ b/assembl/static2/js/app/components/cookies/cookieSetter.jsx
@@ -24,7 +24,6 @@ type CookieSetterProps = {
   locale: string
 };
 
-// @$FlowFixMe globalAnalytics is a global variable
 const matomoHost = globalAnalytics.piwik ? globalAnalytics.piwik.host : null;
 
 const CookieSetter = ({ handleToggle, cookie, toggleCookieType, locale }: CookieSetterProps) => {
--- a/assembl/static2/js/app/components/cookies/cookieSetter.jsx
+++ b/assembl/static2/js/app/components/cookies/cookieSetter.jsx
@@ -53,6 +53,7 @@ const CookieSetter = ({ handleToggle, cookie, toggleCookieType, locale }: Cookie
         <span className="dark-title-3 ellipsis">{cookieName}</span>
         <Helper
           helperText={I18n.t(`cookies.${name}Helper`)}
+          popOverClass="cookie-helper-popover"
           classname="cookie-helper"
         />
       </div>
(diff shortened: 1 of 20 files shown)


assembl/static2/js/app/components/cookies/cookieSetter.jsx
```diff
@@ -0,0 +1,72 @@
...
+import { I18n } from 'react-redux-i18n';
+import classnames from 'classnames';
+
+// Components
+import Helper from '../common/helper';
+import CookieSwitch from './cookieSwitch';
+
+import { COOKIE_TRANSLATION_KEYS } from '../../constants';
+
+export type CookieObject = {
```
cedrtang: Could it be worth to put in flow/object_types.js if used somewhere else ?

assembl/static2/js/app/components/cookies/cookieSetter.jsx
```diff
@@ -0,0 +1,72 @@
...
+  const toggleSwitch = () => {
+    const { accepted, cookieType } = cookie;
+    const updatedCookie = { ...cookie, accepted: !accepted, cookieType: toggleCookieType(cookieType) };
+    handleToggle(updatedCookie);
+  };
+
+  const { name, category, accepted } = cookie;
+
+  const cookieName = Object.keys(COOKIE_TRANSLATION_KEYS).includes(name) ? I18n.t(`cookies.${name}`) : name;
+  const cookieIsMatomo = name === 'matomo';
```
cedrtang: Is it worth creating a constant in constants.js for 'matomo'?

assembl/static2/js/app/components/cookies/cookieSetter.jsx
```diff
@@ -0,0 +1,72 @@
...
+  const matomoOptOutLink = matomoHost ?
+    `https://${matomoHost}/index.php?module=CoreAdminHome&action=optOut&language=${locale}`
+    : null;
+
+  const required = category === 'essential';
```
cedrtang: Is it worth creating a constant in constants.js for 'essential'?

assembl/static2/js/app/components/cookies/cookieSetter.jsx
MilanBarande: It's only used in this folder

pascalhuynh:
same remarks as Cedric, otherwise LGTM

assembl/static2/js/app/components/cookies/cookieSetter.jsx
```diff
@@ -0,0 +1,72 @@
+// @flow
+/* global globalAnalytics */
```
cedricmessiant: what is this?

assembl/static2/js/app/components/cookies/cookieSetter.jsx
```diff
@@ -0,0 +1,72 @@
...
+};
+
+type cookieSetterProps = {
+  handleToggle: Function,
+  cookie: CookieObject,
+  toggleCookieType: Function,
+  locale: string
+};
+
+// @$FlowFixMe globalAnalytics is a global variable
```
cedricmessiant: you should probably declare the global variable in a new `flow/globals.flow.js` file (see this answer: https://stackoverflow.com/a/44376914/2318507 )

assembl/static2/js/app/components/cookies/cookieSetter.jsx
cedricmessiant: yeah, I agree, if it's a cookie, it should go to constants.
These two lines are a bit strange since you are declaring a `cookieName` const but then use `name` line 40. It looks like an error. If it's not you should move `cookieName` declaration to avoid this confusion

cedricmessiant: oh ok, it seems that `cookieName` may be renamed `cookieLabel` or something like that

assembl/static2/js/app/components/cookies/cookieSwitch.jsx
```diff
@@ -0,0 +1,46 @@
...
+type cookieToggleProps = {
+  cookieIsMatomo: boolean,
+  matomoOptOutLink: ?string,
+  toggleSwitch: Function,
+  accepted: boolean,
+  name: string,
+  required: boolean
+}
+
+const cookieSwitch = ({ cookieIsMatomo, matomoOptOutLink, toggleSwitch, accepted, name, required }: cookieToggleProps) => {
```
cedricmessiant: please use an uppercase for components first letter

assembl/static2/js/app/components/cookies/cookiesSelector.jsx
```diff
@@ -36,36 +36,36 @@ const CookiesSelector = ({
...
-                onClick={() => {
-                  handleCategorySelection(category);
-                }}
-              >
-                <span className={classnames('assembl-icon-right-dir', { 'active-arrow': isActiveKey })} />
-                <Translate value={`cookiesPolicy.${category}`} className="dark-title-3" />
+          Object.keys(cookies).map((category) => {
+            const isActiveKey = category === activeKey;
+            return (
+              <div key={`category-${category}`}>
```
cedricmessiant: you don't need `category-` (key is relative to context)

assembl/static2/js/app/components/cookies/cookieSetter.jsx
MilanBarande: It's a global variable needed to access the matomo_host of the server

ayyazdaniaryan: I know globalAnalytics is used in V1 interface, in fact, for piwik. We use React-Router-Piwik in V2, which is completely different, and this variable becomes redundant.

assembl/static2/js/app/components/cookies/cookieSetter.jsx
```diff
@@ -0,0 +1,72 @@
...
+  name: string,
+  category: string,
+  accepted: boolean,
+  cookieType: string
+};
+
+type cookieSetterProps = {
+  handleToggle: Function,
+  cookie: CookieObject,
+  toggleCookieType: Function,
```
vincentfretin: Please don't use `Function`, specify the complete function signature.

assembl/static2/js/app/components/cookies/cookieSwitch.jsx
MilanBarande: I thought you don't do that for functional components and only use an uppercase when you render it (<CookieSwitch />) but there seems to be a debate :) @pascalhuynh

assembl/static2/js/app/components/cookies/cookieSetter.jsx
MilanBarande: Is this the same thing? This was added by @vincentfretin on analytics_common.jinja2 on the previous cookies branch

assembl/static2/js/app/components/cookies/cookieSwitch.jsx
vincentfretin: Like I said, I don't agree with this style change, https://github.com/assembl/assembl/pull/503#commitcomment-30357255

MilanBarande: Allright 😃

ayyazdaniaryan: @vincentfretin oops, I didn't see your tag on The previous PR. I am 100% on the side of Capitalizing functional components. We do it for class based components, why not for functional? And agreed, I search for components with case sensitivity, it makes finding things quicker. I'll copy/paste this on the other thread.

assembl/static2/js/app/components/cookies/cookieSetter.jsx
MilanBarande: Little issue: this variable is empty on localhost so I can't log its content. @vincentfretin globalAnalytics.host is a string and globalAnalytics.isActive is a boolean, right?

vincentfretin: globalAnalytics.piwik.isActive is a boolean yes, globalAnalytics.piwik.host is a string

PR #513, opened 2018-09-07
Add nginx setup task to fabfile and use sudoer instead of sudo_user

mohamedzayed92:


fabfile.py
```diff
@@ -46,9 +46,8 @@ def running_locally(hosts=None):
 
 
 def sudo(*args, **kwargs):
-    sudoer = env.get("sudo_user", None) or env.get("user")
-    with settings(user=sudoer,
-                  sudo_prefix='sudo -i -S -p \'{}\''.format(env.sudo_prompt)):
+    sudoer = env.get("sudoer", None) or env.get("user")
```
ayyazdaniaryan: So, make a comment here to not use the `sudo_user` env key.

fabfile.py
```diff
@@ -1092,6 +1091,21 @@ def app_compile_nodbupdate():
...
+def setup_nginx_file():
+    """Creates nginx config file from template."""
+    sudo("cp %s/doc/sample_nginx_config/assembl.yourdomain.com /etc/nginx/sites-available/assembl.%s" % (env._projectpath, env.public_hostname))
+    sudo("ln -s /etc/nginx/sites-available/assembl.%s /etc/nginx/sites-enabled/" % env.public_hostname)
+    sudo("/etc/init.d/nginx restart")
+
+
+@task
+def setup_assembl_service():
+    sudo("cp %s/doc/sample_systemd_script/assembl.service /etc/systemd/system/assembl.service" % env.public_hostname)
```
ayyazdaniaryan: You should make the service file a jinja2 template, as the project path, the assembl user, etc are all variables based on the RC files. You should render the template and put the template with the right permissions in the systemd directory.

mohamedzayed92:
@ayyazdaniaryan Closed this pull request. Opened a new one on a new branch. Since there was a lot of changed code on develop.

PR #516, opened 2018-09-12
Front improvement part3

felixsaintgenis:


cedricmessiant:
it would have been nice to add a test or two for `Avatar` component

PR #520, opened 2018-09-18
Extract vote results- WIP

mohamedzayed92:


assembl/views/api2/votes.py
```diff
@@ -349,3 +350,65 @@ def global_vote_results_csv(request):
...
+    extract_list = []
+    ctx = request.context
+    widget = ctx._instance
+    if widget.activity_state != "ended":
+        permissions = get_permissions(user_id, ctx.get_discussion_id())
+        if P_ADMIN_DISC not in permissions:
+            raise HTTPUnauthorized()
+    user_prefs = LanguagePreferenceCollection.getCurrent()
+    db = widget.db
+    fieldnames = ["Nom du contributeur", "Adresse mail du contributeur", "Date/heure du vote", "Proposition"]
```
cedricmessiant: I don't remember but I think we don't need translations for exports? It may be nice to use i18n here if in the future we want to export data with the language of the person who asks for the export (or to add a language field in the export page)

assembl/views/api2/votes.py
```diff
@@ -349,3 +350,65 @@ def global_vote_results_csv(request):
...
+    ideas = widget.db.query(Idea).join(AbstractVoteSpecification).filter(AbstractVoteSpecification.widget_id == widget.id).distinct().all()
+    votes = widget.db.query(AbstractIdeaVote).filter(AbstractVoteSpecification.widget_id==widget.id).all()
+    users = widget.db.query(User).all()
+    vote_specs = widget.db.query(AbstractVoteSpecification).filter(AbstractVoteSpecification.widget_id == widget.id).all()
+    #token_category_specifications = widget.db.query(TokenCategorySpecification).filter(TokenCategorySpecification.widget_id == widget.id).all()
```
cedricmessiant: don't forget to remove this comment if you don't need it before merge

assembl/views/api2/votes.py
```diff
@@ -349,3 +350,69 @@ def global_vote_results_csv(request):
...
     output.seek(0)
     return Response(body_file=output, content_type='text/csv', content_disposition='attachment; filename="vote_results.csv')
+
+
+@view_config(context=InstanceContext, name="extract_csv_voters",
+             ctx_instance_class=VotingWidget, request_method='GET',
+             permission=P_DISC_STATS)
+def extract_voters(request):
+    import assembl.models as m
+    import numpy as np
```
ayyazdaniaryan: Don't use numpy here. It's not an Assembl dependency. You only have it pip installed locally. And you only use it for the range.

assembl/views/api2/votes.py
```diff
@@ -349,3 +350,69 @@ def global_vote_results_csv(request):
...
+        if vote.type == u'token_idea_vote':
+            token_category = m.TokenCategorySpecification.get(vote.token_category_id).name.best_lang(user_prefs).value or ""
+            if token_category not in fieldnames:
+                fieldnames.append(token_category.encode('utf-8'))
+            extract_info.update({token_category : str(vote_value).encode('utf-8')})
+
+        if vote.type == u'gauge_idea_vote':
+            vote_spec = m.AbstractVoteSpecification.get(vote.vote_spec_id)
+            if vote_spec.type == u'number_gauge_vote_specification':
+                options = list(np.arange(vote_spec.minimum, vote_spec.maximum, vote_spec.maximum/vote_spec.nb_ticks))
```
ayyazdaniaryan: Use `range()` instead

assembl/views/api2/votes.py
```diff
@@ -349,3 +350,69 @@ def global_vote_results_csv(request):
...
+            "Nom du contributeur": contributor.encode('utf-8'),
+            "Adresse mail du contributeur": contributor_mail.encode('utf-8'),
+            "Date/heure du vote": str(vote_date).encode('utf-8'),
+            "Proposition": proposition.encode('utf-8'),
+        }
+
+        if vote.type == u'token_idea_vote':
+            token_category = m.TokenCategorySpecification.get(vote.token_category_id).name.best_lang(user_prefs).value or ""
+            if token_category not in fieldnames:
+                fieldnames.append(token_category.encode('utf-8'))
```
ayyazdaniaryan: Hmm, adding columns at runtime will make a debate with multiple votes very hard to read. If you're not adding to the column after this, then fine. It's ugly to read, and if there are multiple token votes with the same category name, it will be very confusing, but for a very primitive first pass, it's okay. These should not be CSVs, instead, they should be excel format, a token specification per sheet.

assembl/views/api2/votes.py
mohamedzayed92: I tried to use `range()` but apparently range only works for integers, in the case of gauge values, it could be floats. I will try to look for something else

ayyazdaniaryan: @mohamedzayed92 https://code.activestate.com/recipes/66472/

assembl/views/api2/votes.py
mohamedzayed92: @ayyazdaniaryan Actually, I raised the ugliness of the CSV with Fred, who raised the issue with christy. They want only one CSV no matter how long it is. the only thing that we modified in the user story is to put blank columns between each vote (whether it is token or gauge) , which I did. During Sprint planning, we agreed not to make several sheets per export, unless you know a way to do it. so it was finally decided to to make two seperate buttons.

assembl/static2/js/app/utils/translations.js
```diff
@@ -764,7 +764,11 @@ const Translations = {
         defaultSectionTitle: "Exporter les données",
         taxonomySectionTitle: "Exporter les taxonomies",
         translateTheMessagesIn: "Traduire l'ensemble des messages en :",
-        anonymous: "Rendre les données anonymes"
+        anonymous: "Rendre les données anonymes",
+        vote: {
+          vote_results_csv: "Exporter les données générales du module de vote",
```
ayyazdaniaryan: I would name these in camelCase

assembl/static2/js/app/components/administration/voteSession/exportVotesSection.jsx
```diff
@@ -0,0 +1,68 @@
...
+          <FormGroup>
+            {options.map(option => (
+              <Radio
+                key={option}
+                checked={this.state.exportRoute === option}
+                name="exportLink"
+                value={option}
+                onChange={this.handleRouteChange}
+              >
+                <Translate value={`administration.export.vote.${option}`} />
```
ayyazdaniaryan: ```${_.camelCase(option)}``` instead

assembl/static2/js/app/utils/translations.js
cedricmessiant: you're right, just did the change

cedricmessiant:
@amensouissi I did the changes and refactored the `ExportSection` component. You can check my 2 last commits if you want

vincentfretin:
I almost completely rewrite the export...
https://github.com/assembl/assembl/commit/a5d7c5ca4a2886de59ba4f297b515aa5149c8b56

PR #524, opened 2018-09-20
Next local only

maparent:


assembl/views/__init__.py
```diff
@@ -659,6 +660,24 @@ def redirector(request):
         'home', discussion_slug=request.matchdict.get('discussion_slug')))
 
 
+def sanitize_next_view(next_view):
+    if next_view and ':/' in next_view:
```
ayyazdaniaryan: The `:/` check is not necessary. A `urlparse` will yield no netloc if it doesn't exist.

assembl/models/preferences.py
```diff
@@ -546,6 +546,17 @@ def get_preference_data_list(self):
             "default": True,
         },
 
+        # Show different messages in case of a failed registration
+        {
+            "id": "generic_auth_errors",
+            "name": _("Generic Auth Errors"),
+            "value_type": "bool",
+            "description": _("Display a generic error message in case of bad authentication"),
```
ayyazdaniaryan: This isn't just for auth, it's being used for auth for now. Can set this to be `generic_errors` and change the description to generalize all errors.

assembl/views/api2/auth.py
```diff
@@ -39,10 +39,11 @@
 from ..auth.views import (
     send_change_password_email, from_identifier, send_confirmation_email,
     maybe_auto_subscribe)
+from assembl.lib import logging
 
 
 _ = TranslationStringFactory('assembl')
-
+generic_error_message = "This username and password combination is not valid."
```
ayyazdaniaryan:
```
generic_error_message = _("The provided input combination is not valid")```

File: assembl/views/api2/auth.py [hunk 4]
```diff
@@ -343,9 +345,15 @@ def reset_password(request):
...
-            raise JSONError(
-                localizer.translate(_("The user does not exist")),
-                code=HTTPNotFound.code)
+            if not discussion.preferences['generic_auth_errors']:
+                raise JSONError(
+                    localizer.translate(_("The user does not exist")),
+                    code=HTTPNotFound.code)
+            else:
+                raise JSONError(
+                    localizer.translate(_(generic_error_message)),
```
ayyazdaniaryan: ```localizer.translate(generic_error_message)```

assembl/views/api2/auth.py
```diff
@@ -343,9 +345,15 @@ def reset_password(request):
...
+                    code=HTTPNotFound.code)
+                logger.error("The user does not exist.")
```
ayyazdaniaryan: Be more specific, please.
```
"[Password Reset] The user with the identifier "%(identifier)s" does not exist." % (identifier,)
```

assembl/views/api2/auth.py
```diff
@@ -354,9 +362,15 @@ def reset_password(request):
...
-                code=HTTPNotFound.code)
+            if not discussion.preferences['generic_auth_errors']:
+                raise JSONError(
+                    localizer.translate(_("This email does not exist")),
+                    code=HTTPNotFound.code)
+            else:
+                raise JSONError(
+                    localizer.translate(_(generic_error_message)),
+                    code=HTTPNotFound.code)
+                logger.error("This email does not exist.")
```
ayyazdaniaryan: Again, follow the above example to be more specific. In a long logger, need to search for this quickly, and in a lot of other information, need clear information of which view the error came from, and what the trigger point was.

assembl/views/api2/auth.py
```diff
@@ -452,21 +475,36 @@ def assembl_register_user(request):
...
                          ErrorTypes.INVALID_EMAIL)
     username = json.get('username', None)
     if username:
         if session.query(Username).filter(
                 func.lower(Username.username) == username.lower()).count():
-            errors.add_error(localizer.translate(_(
-                "We already have a user with this username.")),
-                ErrorTypes.EXISTING_USERNAME,
-                HTTPConflict.code)
+            if not discussion.preferences['generic_auth_errors']:
```
ayyazdaniaryan: You've made the comparison `not discussion.preference['generic_auth_errors']` multiple times. Could create a variable use `generic_error = discussion.preference['generic_auth_errors']` and do a comparison this way, if you wish.

assembl/views/api2/auth.py
```diff
@@ -428,6 +450,7 @@ def assembl_register_user(request):
     localizer = request.localizer
     session = AgentProfile.default_db
     json = request.json
+    logger = logging.getLogger('assembl.auth')
```
ayyazdaniaryan: There is no `assembl.auth` logs, yet. Should switch to `assembl`.

assembl/views/api2/auth.py
```diff
@@ -452,21 +475,36 @@ def assembl_register_user(request):
...
-                HTTPConflict.code)
+            if not discussion.preferences['generic_auth_errors']:
+                errors.add_error(localizer.translate(_(
+                    "We already have a user with this email.")),
+                    ErrorTypes.EXISTING_EMAIL,
+                    HTTPConflict.code)
+            else:
+                errors.add_error(localizer.translate(_(
+                    generic_error_message)),
+                    ErrorTypes.EXISTING_EMAIL,
```
ayyazdaniaryan: This, in fact, reveals a lot to the front-end. Must add a new enumType to `ErrorTypes`, perhaps `GENERIC`. And the front-end will have to deal with that case. I can do the front-end @mohamedzayed92 .

assembl/views/api2/auth.py
```diff
@@ -452,21 +475,36 @@ def assembl_register_user(request):
...
+                    HTTPConflict.code)
+                logger.error("We already have a user with this email %s" % email)
```
ayyazdaniaryan: Give the context, but yah, this is good.

assembl/views/api2/auth.py
```diff
@@ -452,21 +475,36 @@ def assembl_register_user(request):
...
-                HTTPConflict.code)
+            if not discussion.preferences['generic_auth_errors']:
+                errors.add_error(localizer.translate(_(
+                    "We already have a user with this username.")),
+                    ErrorTypes.EXISTING_USERNAME,
+                    HTTPConflict.code)
+            else:
+                errors.add_error(localizer.translate(_(
+                    generic_error_message)),
+                    ErrorTypes.EXISTING_USERNAME,
```
ayyazdaniaryan: Same, use the new enum type created.

configs/base_env.rc
```diff
@@ -18,7 +18,7 @@ _sentry_db_host = localhost
 # _sentry_db_password =
 # Replace with your admin email if necessary
 assembl.admin_email = webmaster@%(public_hostname)s
-
+assembl.generic_auth_errors = false
```
ayyazdaniaryan: Let's rename to `generic_errors` since this could affect more than auth (in the future)

assembl/views/__init__.py
maparent: true, but that way I get a fast path for most common case. urlparse is not nothing.

PR #525, opened 2018-09-23
Packaging

maparent:


assembl/configs/base_env.rc
```diff
@@ -99,6 +122,12 @@ supervisor__autostart_uwsgi = true
 supervisor__autostart_elasticsearch = true
 supervisor__autostart_urlmetadata = true
 
+alembic__script_location = %(code_root)s/assembl/alembic
```
ayyazdaniaryan: Comment the need for this being here.

assembl/scripts/ini_files.py
```diff
@@ -78,9 +93,42 @@ def ensureSection(config, section):
...
+    config['here'] = config.get('here', os.getcwd())
+    if template.endswith('.tmpl'):
+        with open(template) as tmpl:
+            result = tmpl.read() % config
+    elif template.endswith('.jinja2'):
+        from jinja2 import Environment
+        env = Environment()
+        with open(template) as tmpl:
+            tmpl = env.from_string(tmpl.read())
+        # Boolean overloading
```
ayyazdaniaryan: Because you bypassed the Jinja2 read? Why?

assembl/scripts/ini_files.py
```diff
@@ -78,9 +93,42 @@ def ensureSection(config, section):
...
+        for (k, v) in config.items():
+            if str(v).lower() == 'false':
```
ayyazdaniaryan: Why not for True cases?

assembl/scripts/ini_files.py
```diff
@@ -380,6 +438,9 @@ def diff_ini(first, second, diff=None, existing_only=False):
 def compose(rc_filename, random_file=None):
     """Compose local.ini from the given .rc file"""
     rc_info = combine_rc(rc_filename)
+    rc_info['*code_root'] = code_root(rc_info)
+    # Special case: uwsgi does not do internal computations.
+    rc_info['uwsgi__virtualenv'] = venv_path(rc_info)
```
ayyazdaniaryan: USWGI does handoff to the virtualenv to get the proper version of the files?

fabfile.py
```diff
@@ -1,2435 +1 @@
...
-    with settings(warn_only=True), hide('stdout'):
-        checkDatabase = venvcmd('assembl-pypsql -1 -u {user} -p {password} -n {host} "{command}"'.format(
-            command="SELECT 1 FROM pg_database WHERE datname='%s'" % (env.db_database),
-            password=env.db_password, host=env.db_host, user=env.db_user))
-    if not checkDatabase.failed:
-        print(yellow("Cannot connect to database, trying to create"))
-        deleteDatabase = run('PGPASSWORD=%s dropdb --host=%s --username=%s %s' % (
-            env.db_password, env.postgres_db_host, env.db_user, env.db_database))
-        if deleteDatabase.succeeded:
-
```
ayyazdaniaryan: This file should have a big-sh comment at the top as to why it exists. What Fabric does, and a brief history lesson. It needs to be explained the conditions where running it is good and well, conditions where using it crashes the system.

assembl/configs/base_env.rc
```diff
@@ -22,6 +38,10 @@ assembl.generic_errors = false
 memcached_host = localhost
 mysql_host=localhost
 webpack_host=localhost
+nginx_ipv6 = true
+use_letsencrypt = true
+csp_header = default-src 'self'; script-src 'self' 'unsafe-inline' 'unsafe-eval' data: %(piwik_host)s *.facebook.net *.twimg.com *.twitter.com ; frame-src 'self' www.youtube.com *.facebook.com embed.ted.com *.twitter.com twitter.com player.vimeo.com; connect-src 'self' wss://%(public_hostname)s %(sentry_host)s %(piwik_host)s *.twitter.com; style-src 'self' 'unsafe-inline' fonts.googleapis.com *.twimg.com *.twitter.com ; font-src 'self' data: fonts.gstatic.com; img-src 'self' data: %(piwik_host)s *.facebook.com www.gravatar.com *.googleusercontent.com *.twimg.com *.twitter.com framapic.org *.cloudfront.net
```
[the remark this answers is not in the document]
maparent: Unfortunately .rc files expects single-line...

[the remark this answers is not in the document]
maparent: Use word wrap in your text editor :->

assembl/scripts/ini_files.py
maparent: Because jinja does not translate the 'false' string into the False value. It's actually a jinja bug, mentioned here: https://github.com/ansible/ansible/issues/14983

assembl/scripts/ini_files.py
maparent: because the 'true' string is truish anyway.

assembl/scripts/ini_files.py
maparent: yes. BUT no variable substitution in that case, as per comment.

assembl/scripts/ini_files.py
ayyazdaniaryan: I would comment this link here, so that it can be checked, and this code be altered if/when another developer reviews it.

PR #529, opened 2018-10-02
Fix profile page devas 1827

cedricmessiant:
This PR:
- adds a `GlobalErrorBoundary` component that will catch all errors thrown in React components
- fix DEVAS-1827


vincentfretin:
I think there is a few places where we return neither error, nor data, so the throw error won't work and the error will be silent. Can you please do a quick check and add them if missing?

assembl/static2/js/app/components/common/GlobalErrorBoundary.jsx
```diff
@@ -0,0 +1,40 @@
...
+};
+
+type State = {
+  error: ?Error
+};
+
+export default class GlobalErrorBoundary extends React.Component<Props, State> {
+  constructor(props: Props) {
+    super(props);
+    this.state = { error: null };
```
amensouissi: I prefer only state = { error: null }

PR #530, opened 2018-10-03
Adding task to set nginx file before DNS and after DNS.

mohamedzayed92:


assembl/fabfile.py
```diff
@@ -1218,6 +1218,29 @@ def generate_dh_group():
...
+        sudo("rm /etc/nginx/sites-enabled/assembl.%s" % (env.public_hostname))
+    if exists("/etc/nginx/sites-enabled/assembl.%s" % (env.server_ip_address)):
+        sudo("rm /etc/nginx.sites-enabled/assembl.%s" % (env.server_ip_address))
+    # There are two nginx template, one before DNS, the second one for after DNS propagation
+    if not ready_for_production:
+        assembl_nginx_template = jenv.get_template('assembl.yourdomain.com')
+    else:
+        assembl_nginx_template = jenv.get_template('assembl.yourdomain.production-ready')
+    with open('assembl.%s' % (env.public_hostname), 'w') as f:
+        f.write(assembl_nginx_template.render(public_hostname=env.public_hostname))
```
ayyazdaniaryan: Remove nearly everything up until here. After #525, there is a new way to compute templates. You could use the `fill_template` method, or simply call the assembl script: `assembl-ini-files template -o <nginx_conf_name> <configs/client.rc> nginx_default.jinja2'. I feel like fill_template might be easier case for you here. Create a file, fill it with the **env dict, and then put it on the server like you're doing here

PR #532, opened 2018-10-04
Landing page administration bugfixes

MilanBarande:
This PR is linked to [this](https://bluenove.atlassian.net/browse/DEVAS-1777), [this](https://bluenove.atlassian.net/browse/DEVAS-1778) and [this](https://bluenove.atlassian.net/browse/DEVAS-1779) ticket

🐛 The "Add a text & multimedia module" button is now disabled when you haven't checked all the ones you've already created (DEVAS-1777)
🐛 If you create several of these modules they will be numbered (DEVAS-1778)
🐛 The modules in the preview section are numbered as well
🐛 The order issues in the manageModule component has been solved: modules are now presented alphabetically with the Header always first and the Footer always last (DEVAS-1779)

Also updated tests, added flow annotations and refactored the landing page administration.


assembl/static2/js/app/components/administration/landingPage/manageModules.jsx
```diff
@@ -57,6 +85,10 @@ export const DumbManageModules = ({
     const includeFooter = true;
     return displayModal(null, body, includeFooter, footer);
   };
+  const textAndMultimediaIsChecked = enabledModules.some(
+    module => module.getIn(['moduleType', 'title']) === 'Text & Multi-media'
```
vincentfretin: You should use `module.getIn(['moduleType', 'identifier']) === 'INTRODUCTION'` title may change in the future.

MilanBarande:
Can't merge as things stand. Titles have to be updated on the preview column and there is a bug with the title of newly created Text&Multimedia modules...

assembl/static2/js/app/components/administration/landingPage/manageModules.jsx
```diff
@@ -16,16 +18,66 @@ import {
...
+  contact: 'CONTACT',
+  data: 'DATA',
+  news: 'NEWS',
+  partners: 'PARTNERS',
+  introduction: 'INTRODUCTION',
+  footer: 'FOOTER'
+};
+
+export const addCountSuffix = (modules: Array<LandingPageModule>): Array<LandingPageModule> => {
+  // This function counts the occurencies of each module titles
```
cedricmessiant: Your comment is misleading, I prefer something like:

`// Add a suffix to the title of the module if this module appears more than one time. This suffix is the number of times this module appeared in the array.`

Maybe we can change the function name to: `addEnumSuffixToModuleTitles`

assembl/static2/js/app/components/administration/landingPage/manageModules.jsx
```diff
@@ -69,9 +149,18 @@ export const DumbManageModules = ({
...
-            <SelectModulesForm lang={locale} moduleTypes={moduleTypes} modulesById={modulesById} toggleModule={toggleModule} />
+            <SelectModulesForm
+              lang={locale}
+              moduleTypes={updatedModuleTypes}
+              modulesById={modulesById}
+              toggleModule={toggleModule}
+            />
             <div className="margin-xl">
-              <Button className="button-submit button-dark" onClick={displayConfirmationModal}>
+              <Button
```
cedricmessiant: It took me time to realize/remember that this button is here only to add text and multimedia modules. I guess it would be clearer if we had some kind of `AddTextAndMultimediaButton` component with all this logic (modal creation, etc)
We will need to rethink/refactor this component if we need there is another module that we need to add multiple times in the landing page but IMHO this whole UI needs to be totally redesigned.

assembl/static2/js/app/components/administration/landingPage/manageModules.jsx
```diff
@@ -102,9 +191,17 @@ const mapDispatchToProps = dispatch => ({
   moveModuleDown: id => dispatch(moveLandingPageModuleDown(id)),
   moveModuleUp: id => dispatch(moveLandingPageModuleUp(id)),
   toggleModule: id => dispatch(toggleLandingPageModule(id)),
-  createModule: (nextOrder, identifier = 'INTRODUCTION') => {
+  createModule: (nextOrder, numberOfDuplicatesModules, identifier = MODULES_IDENTIFIERS.introduction) => {
     const newId = createRandomId();
-    return dispatch(createLandingPageModules(newId, identifier, nextOrder));
+    return dispatch(
+      createLandingPageModules(
```
cedricmessiant: why is this called `createLandingPageModules`? shouldn't it be `createLandingPageModule`?

assembl/static2/js/app/components/administration/landingPage/manageModules.jsx
```diff
@@ -102,9 +191,17 @@ const mapDispatchToProps = dispatch => ({
...
+        newId,
+        identifier,
+        numberOfDuplicatesModules,
```
cedricmessiant: `numberOfDuplicates`?
If possible, it is better to add the number here before to call the redux action (to avoid an additional param). You also need to check if there is no duplicates and don't add the suffix in this case.
Something like `I18n.t('administration.landingPage.manageModules.textAndMultimedia', { count: numberOfDuplicates })` should do the trick (and you can even add 'administration.landingPage.manageModules.textAndMultimedia_0` to handle the no duplicates case

assembl/static2/js/app/components/administration/landingPage/manageModules.jsx
```diff
@@ -16,16 +18,80 @@ import {
...
+  tweets: 'TWEETS',
+  contact: 'CONTACT',
+  data: 'DATA',
+  news: 'NEWS',
+  partners: 'PARTNERS',
+  introduction: 'INTRODUCTION',
+  footer: 'FOOTER'
+};
+
+export const addEnumSuffixToModuleTitles = (modules: Array<LandingPageModule>): Array<LandingPageModule> => {
```
cedricmessiant: you don't need this function here, it should move to the file in which it is used or in a distinct file (maybe in `utils/` folder but it is probably overkill and not very easy to make it clear and reusable)

assembl/static2/js/app/components/administration/landingPage/manageModules.jsx
```diff
@@ -38,6 +104,9 @@ export const DumbManageModules = ({
 }: Props) => {
   const displayConfirmationModal = () => {
     const body = <Translate value="administration.landingPage.manageModules.confirmationModal" />;
+    const numberOfTextAndMultimediaModules = moduleTypes.filter(
```
cedricmessiant: same code here and below

assembl/static2/js/app/components/administration/landingPage/manageModules.jsx
MilanBarande: Well I can put it in utils because it's used in administration.jsx and this file doesn't have flow nor a test file. Of course we need to add those 2 things but it's not the purpose of this branch. Anyway administration.jsx is already too big

assembl/static2/js/app/components/administration/landingPage/manageModules.jsx
MilanBarande: True

assembl/static2/js/app/components/administration/landingPage/manageModules.jsx
MilanBarande: >  but IMHO this whole UI needs to be totally redesigned.

💯

assembl/static2/js/app/components/administration/landingPage/manageModules.jsx
MilanBarande: But it's impossible not to have a duplicate if you are creating an additional module.

assembl/static2/js/app/components/form/utils.js
```diff
@@ -87,4 +90,20 @@ export function getFileVariable(img: FileValue, initialImg: ?FileValue): FileVar
   // we need to send image: null if we didn't change the image.
   const variab = img && img.externalUrl instanceof File ? img.externalUrl : null;
   return variab;
-}
\ No newline at end of file
+}
+
+export const addEnumSuffixToModuleTitles = (modules: Array<LandingPageModule>): Array<LandingPageModule> => {
```
cedricmessiant: Sorry, I don't like it here. `form/utils` contains only `react-final-form` utils. This code is too specific (really tightly bound to landing page modules). I was thinking about the main `utils` folder but as I said, to move it there, you need to do lots of changes to make it generic. I suggest that you move it in its own file `addEnumSuffixToModuleTitle` in `administration/landingPage/`. Thanks!

cedricmessiant: and don't forget to change the import ;-)

assembl/static2/js/app/components/administration/landingPage/manageModules.jsx
```diff
@@ -135,13 +113,13 @@ export const DumbManageModules = ({
...
-                className="button-submit button-dark"
-                onClick={displayConfirmationModal}
-                disabled={!allTextAndMultimediaAreChecked}
-              >
-                <Translate value="administration.landingPage.manageModules.textAndMultimediaBtn" />
-              </Button>
+              <AddModuleButton
+                numberOfTextAndMultimediaModules={numberOfTextAndMultimediaModules}
+                numberOfEnabledModules={enabledModules.size}
+                closeModal={() => {}}
```
cedricmessiant: is it normal?

assembl/static2/tests/unit/components/administration/landingPage/addModuleButton.spec.jsx
```diff
@@ -0,0 +1,39 @@
...
+    createModule: createModuleSpy,
+    numberOfDuplicatesModules: 3,
+    numberOfEnabledModules: 12,
+    allDuplicatesAreChecked: true,
+    buttonTitleTranslationKey: 'textAndMultimediaBtn'
+  };
+  const wrapper = shallow(<AddModuleButton {...props} />);
+  it('should render a non disabled button', () => {
+    expect(wrapper.find('Button [disabled=false]')).toHaveLength(1);
+  });
```
cedricmessiant: please leave one blank lines between each test (much easier to read)

assembl/static2/js/app/components/administration/landingPage/manageModules.jsx
MilanBarande: absolutely not, oblivion, thanks :)

PR #537, opened 2018-10-04
calculate the global pristine for the admin form

amensouissi:


cedricmessiant:
Did you check if we should fix this also in another admin form?

PR #543, opened 2018-10-08
Fix votes export

cedricmessiant:
Fixes the following in votes export:

- add BOM so that excel knows this is UTF-8 (DEVAS-1792)
- add filename and content disposition for detailed votes export
- fix bug in `global_vote_results_csv` (I had an `AttributeError` because `spec` was used before it was defined (in fact, it was defined in the list comprehension that creates the `template_specs` line 244 so the error was hidden)
- add filename and add BOM for taxonomies extract

### DIFF
--- a/assembl/views/api2/votes.py
+++ b/assembl/views/api2/votes.py
@@ -274,6 +274,8 @@ def global_vote_results_csv(request):
         coltitles.append('Total votes')
 
     output = StringIO()
+    # include BOM for Excel to open the file in UTF-8 properly
+    output.write(u'\ufeff'.encode('utf-8'))
     csvw = csv.writer(output)
     csvw.writerow(coltitles)
     from assembl.graphql.vote_session import get_avg_choice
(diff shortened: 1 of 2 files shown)


assembl/views/api2/votes.py
```diff
@@ -266,14 +266,16 @@ def global_vote_results_csv(request):
         else:
             coltitles.append(u'{title} - moyenne'.format(title=title).encode('utf-8'))
             if isinstance(template_spec, NumberGaugeVoteSpecification):
-                for choice_value in range_float(template_spec.minimum, spec.maximum, spec.nb_ticks):
+                for choice_value in range_float(template_spec.minimum, template_spec.maximum, template_spec.nb_ticks):
```
vincentfretin: good catch, flake8 didn't see that.

assembl/views/api2/votes.py
```diff
@@ -414,4 +417,4 @@ def extract_voters(request):
 
             extract_votes.append(extract_info)
     extract_votes.sort(key=operator.itemgetter('Nom du contributeur'))
-    return csv_response(extract_votes, CSV_MIMETYPE, fieldnames)
+    return csv_response(extract_votes, CSV_MIMETYPE, fieldnames, content_disposition='attachment; filename="detailed_vote_results.csv"')
```
vincentfretin: Thanks for that.

PR #544, opened 2018-10-08
*** flow annotations: Tree component :O ***

amensouissi:


assembl/static2/js/app/components/common/tree.jsx
```diff
@@ -319,34 +117,35 @@ class Tree extends React.Component {
...
-    const { contentLocaleMapping, data, lang, noRowsRenderer } = this.props;
+    const { contentLocaleMapping, data, lang, noRowsRenderer, InnerComponentHeight } = this.props;
     return (
       <WindowScroller>
         {({ height, isScrolling, onChildScroll, scrollTop }) => (
           <AutoSizer
             disableHeight
             onResize={() => {
               this.cache.clearAll();
+              // $FlowFixMe listRef is not null
```
cedrtang: Higher above you set listRef: List | null = null; Couldn't you check listRef not null before calling the function and then would flow pass ?

assembl/static2/js/app/components/common/treeItem.jsx
```diff
@@ -0,0 +1,267 @@
...
+  onScroll = debounce(() => {
+    const holder = this.holder;
+    if (!holder) {
+      return;
+    }
+    const box = holder.getBoundingClientRect();
+    const pageYOffset = window.pageYOffset;
+    const top = box.top + pageYOffset;
+    // visible if the top of the box is in viewport or next page
+    const isVisible = top < pageYOffset + 2 * window.innerHeight && top > pageYOffset;
```
cedrtang: Just wondering why this calculation with "2 * window.innerHeight"

assembl/static2/js/app/components/common/treeItem.jsx
```diff
@@ -0,0 +1,267 @@
...
+      hashid = hash.replace('#', '').split('?')[0];
+      isVisible = hashid === id || isVisible;
+    }
+
+    return (
+      <div
+        className={classnames(`level level-${level}`, {
+          'border-left child-level': level > 0,
+          'no-shift': level > 3,
+          'padding-right': level > 4,
```
cedrtang: Could it be useful to set those integer as constants with a name explaining their usefullness so we could understand why and customize them if necessary ?

assembl/static2/js/app/components/common/treeItem.jsx
vincentfretin: one window.innerHeight in the viewport, two is the viewport of the next page

PR #546, opened 2018-10-08
Add a validation for JSON, and a test

ayyazdaniaryan:
A validation for JSON in the preference. To avoid faulty json to creep past front-end checks / API based calls to Preferences's `extra_json` field.

### DIFF
--- a/assembl/models/preferences.py
+++ b/assembl/models/preferences.py
@@ -280,7 +280,8 @@ class Preferences(MutableMapping, Base, NamedClassMixin):
         elif data_type == "int":
             assert isinstance(value, int), "Not an integer"
         elif data_type == "json":
-            pass  # no check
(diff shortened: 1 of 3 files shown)


assembl/models/preferences.py
```diff
@@ -280,7 +280,8 @@ def validate_single_value(self, key, value, pref_data, data_type):
         elif data_type == "int":
             assert isinstance(value, int), "Not an integer"
         elif data_type == "json":
-            pass  # no check
+            # Will raise a JSONDecodeError if not a valid JSON
```
cedrtang: Are you catching it somewhere ? What do you do then?

ayyazdaniaryan: No, let it raise an error if it fails, if it succeeds, then the validate function proceeds without an error.

PR #548, opened 2018-10-09
Add test for password change from email

victorneuret:


assembl/tests/views_tests/test_view_auth.py
```diff
@@ -268,3 +269,20 @@ def test_autologin_override(
...
+                "password2": "lolo"}
+
+    # Test token
+    user, validity = verify_password_change_token(token)
+    assert validity == Validity.VALID
+    
+    # Test API
+    response = test_app.post_json('/data/AgentProfile/do_password_change', my_json)
+    assert response.status_code == 200
+    assert old_password != participant1_user.password
```
amensouissi: you have to check the password here: participant1_user.check_password("lolo") ?

PR #549, opened 2018-10-09
DEVAS-1847: fix error when loading BM fiction for the first time

pascalhuynh:


assembl/static2/js/app/pages/brightMirrorFiction.jsx
```diff
@@ -126,6 +126,9 @@ export class BrightMirrorFiction extends Component<LocalBrightMirrorFictionProps
 
   componentWillReceiveProps(nextProps: LocalBrightMirrorFictionProps) {
     // Sync state
+    const { loading } = nextProps.brightMirrorFictionData;
```
amensouissi: use static getDerivedStateFromProps(nextProps, prevState) instead of componentWillReceiveProps...

PR #550, opened 2018-10-09
DEVAS 1843: form to write fiction is bigger. Can now save fiction wit…

cedrtang:
- form to write fiction is bigger
- Can now save fiction with either title or content


assembl/static2/js/app/components/debate/common/topPostForm.jsx
```diff
@@ -137,6 +140,9 @@ class TopPostForm extends React.Component<TopPostFormProps, TopPostFormState> {
             this.setState({ submitting: false });
           });
       });
+    } else if (publicationState === PublicationStates.DRAFT && (!subject && bodyIsEmpty)) {
+      displayAlert('warning', I18n.t('debate.brightMirror.fillEitherTitleContent'));
+      this.setState({ submitting: false });
```
ayyazdaniaryan: Functional form, perhaps. https://medium.freecodecamp.org/functional-setstate-is-the-future-of-react-374f30401b6b

assembl/static2/js/app/components/debate/brightMirror/instructionView.jsx
```diff
@@ -52,6 +52,7 @@ const InstructionView = ({
       bodyMaxLength={NO_BODY_LENGTH}
       draftable
       draftSuccessMsgId="debate.brightMirror.draftSuccessMsg"
+      fullscreen
```
ayyazdaniaryan: How would fullscreen be set upstream?

cedrtang: @ayyazdaniaryan Doesn't need to go higher, it s for the TopFormContainer that is already a higher up component

assembl/static2/js/app/components/debate/common/topPostForm.jsx
MilanBarande: Isn't it only useful when you use the previousState (eg to toggle a boolean or increase a counter) ?

PR #555, opened 2018-10-11
Devpi server

mohamedzayed92:


cedricmessiant:
LGTM (but I don't remember exactly how it works). Did you try to install it somewhere and to distribute an egg or a wheel on it (and then to install this wheel on another machine)? (for distribution, see https://pypi.org/project/twine/ )

However, I don't think we should install devpi via fabric. We can in fact install it once on a "dev" server and it should just work. If the team wants to use fabric for this, can you put this code in another file (the `fabfile.py` is huge).

More importantly, we need to take a decision about whether we need devpi or not (related to my comments on DEVAS-1872 )

cedricmessiant:
I guess that we can close this PR

mohamedzayed92:
Closed PR because we moved away from this solution. We are going in the direction of PyPi and not devPi.

PR #558, opened 2018-10-15
Add script to check if Elasticsearch and database are accessible

victorneuret:


script/check_elasticsearch.sh
```diff
@@ -0,0 +1,73 @@
...
+# if curl didn't succeed 
+if [ $? != 0 ]
+then
+    echo "Fail to connect to elatic search"
+    echo "${CURL}"
+    exit 1
+fi
+
+# clear line to keep cluster_name value only
+CLUSTER_NAME=$(echo "${CURL}" | grep "cluster_name")
```
ayyazdaniaryan: This is a good candidate for jq. We'll make sure that the servers have it installed.
Then:
```suggestion
CLUSTER_NAME=$($CURL | jq '.cluster_name' | tr -d "\"" )
```

script/check_elasticsearch.sh
```diff
@@ -0,0 +1,73 @@
...
+CLUSTER_NAME=${CLUSTER_NAME// /}
+CLUSTER_NAME=${CLUSTER_NAME//\"/}
+CLUSTER_NAME=${CLUSTER_NAME//cluster_name:/}
+CLUSTER_NAME=${CLUSTER_NAME//,/}
+
+# clear line to keep version number value only
+VERSION=$(echo "${CURL}" | grep "number")
```
ayyazdaniaryan: Same thing.
```suggestion
VERSION=$($CURL | jq '.version.number' | td -d "\"")
```

script/check_elasticsearch.sh
```diff
@@ -0,0 +1,73 @@
...
+VERSION=${VERSION//,/}
+
+# Check Elasticsearch cluster name and version
+if ([ "$CLUSTER_NAME" != "$elasticsearch_index" ] || [ "$VERSION" != "$elasticsearch_version" ])
+then
+    exit 1
+fi
+
+# Check if database is accessible
+psql -Uassembl assembl -c "SELECT * FROM pg_catalog.pg_tables" &> /dev/null
```
ayyazdaniaryan: Actually, let's make sure that it's not empty, eh? So, we test the connection, could be good, but let's make sure that the result of the psql command return non-empty list of rows?

- So, firstly, add a semicolon at end of command to execute.
- Next, let's clean up the output and only use the parts of the table that are created by Assembl
- Then count the number of lines

In effect, something like this:
```suggestion
psql -Uassembl assembl -c "SELECT * FROM pg_catalog.pg_tables;" | grep public | wc -l | tr -d ' '
```

ayyazdaniaryan: And then you can count to ensure that the len of this number is greater than 1 (so tables are created). We can argue that this number should be over 100, or whatever the current number of tables is, ie 128 tables. But, that's way more hassle than it's worth. If 1 table is created, then all are created.

assembl/fabfile.py
```diff
@@ -2688,3 +2688,16 @@ def set_fail2ban_configurations():
...
             for path in filters_to_file.values():
                 os.unlink(path)
+
+
+@task
+def install_jq():
+    """
+    Install jq
+    """
+    if not exists('/usr/local/bin/jq'):
```
ayyazdaniaryan: In typical linux installs, it's hosted on `/usr/bin/jq`

script/check_availability.sh
```diff
@@ -0,0 +1,61 @@
...
+# parse elasticsearch_version from local.ini
+elasticsearch_version="$(grep "elasticsearch_version" $LOCAL_INI | tr -d ' ' | awk -F "=" '{print $2}')"
+
+ELASTICSEARCH_URL=http://${elasticsearch_host}:${elasticsearch_port}
+
+CURL="$(curl --fail --silent --show-error ${ELASTICSEARCH_URL} 2>&1)"
+
+# if curl didn't succeed 
+if [ $? != 0 ]; then
+    echo "Failed to connect to Elaticsearch server" >> $ASSEMBL_ROOT/var/log/assembl.log
```
ayyazdaniaryan: Actually, since the changes made in a1d811fdf450b591417c1990dce00f4759f43cc3, can simply do a printf statement and we'll have logs of it in the stdout. So, 
```suggestion
printf "Failed to connect to Elaticsearch server with failure: \t %s\n" "$CURL")
```

script/check_availability.sh
```diff
@@ -0,0 +1,61 @@
...
+# Check Elasticsearch cluster name and version
+if ([ "$CLUSTER_NAME" != "$elasticsearch_index" ] || [ "$VERSION" != "$elasticsearch_version" ]); then
+    echo "Wrong Elasticseach cluster name or version" >> $ASSEMBL_ROOT/var/log/assembl.log
+    exit 1
+fi
+
+# Check if database is accessible
+NB_TABLE="$(psql -Uassembl assembl -c "SELECT * FROM pg_catalog.pg_tables;" | grep public | wc -l | tr -d ' ')"
+if [ $NB_TABLE -lt 1 ]; then
+    echo "database isn't accessible" >> $ASSEMBL_ROOT/var/log/assembl.log
```
ayyazdaniaryan: Similar to comments above.

script/check_availability.sh
```diff
@@ -0,0 +1,61 @@
...
+# if curl didn't succeed 
+if [ $? != 0 ]; then
+    echo "Failed to connect to Elaticsearch server" >> $ASSEMBL_ROOT/var/log/assembl.log
+    echo "${CURL}" >> $ASSEMBL_ROOT/var/log/assembl.log
+    exit 1
+fi
+
+# Install jq if necessary
+if [ $(uname) == "Darwin" ]; then
+    fab -c $ASSEMBL_ROOT/assembl/configs/mac.rc install_jq
```
ayyazdaniaryan: I would leave fabric out of this script. You could run into conditions where you're running system fabric, which could be at a different version. To run fabric properly, can ensure you're in virtualenv. Or simply, run the installation commands (apt-get or brew).

assembl/fabfile.py
```diff
@@ -2688,3 +2688,16 @@ def set_fail2ban_configurations():
         finally:
             for path in filters_to_file.values():
                 os.unlink(path)
+
+
+@task
+def install_jq():
```
ayyazdaniaryan: Put this on the main system installation trail. Likely in `install_server_deps`, this should ensure it's installed on future installs. It can also be a single command run to ensure that servers are migrated properly.

ayyazdaniaryan:
@cedricmessiant Sure, no problem. Just make sure you propagate this change to the setup.py (and any parts of the code that calls `script` module)

script/check_availability.sh
```diff
@@ -44,15 +45,14 @@ VERSION=$(echo $CURL | jq '.version.number' | tr -d "\"")
 
 # Check Elasticsearch cluster name and version
 if ([ "$CLUSTER_NAME" != "$elasticsearch_index" ] || [ "$VERSION" != "$elasticsearch_version" ]); then
-    echo "Wrong Elasticseach cluster name or version" >> $ASSEMBL_ROOT/var/log/assembl.log
+    echo "Wrong Elasticseach cluster name or version"
```
ayyazdaniaryan: prinf

cedricmessiant:
@ayyazdaniaryan  I don't understand why renaming `script` has an impact on `setup.py`. Those are scripts that are in assembl root folder (I used this directory for scripts used in the build job)

maparent:
> @ayyazdaniaryan I don't understand why renaming `script` has an impact on `setup.py`. Those are scripts that are in assembl root folder (I used this directory for scripts used in the build job)

Curious: Why not in `assembl/scripts`? This is where all other scripts are.

ayyazdaniaryan:
@cedricmessiant , my mistake. I moved too fast, I thought this was in fact the folder that @maparent was talking about, the `assembl/scripts/` folder. And there are scripts defined in the setup.py that are in that folder.

PR #563, opened 2018-10-22
Fix the thematics Menu

amensouissi:


assembl/static2/js/app/components/administration/survey/configureThematicForm.jsx
```diff
@@ -33,25 +37,42 @@ export function getFieldName(themeId: string, values: ThemesValue, fieldName: st
...
-        component={FileUploaderFieldAdapter}
-        label={I18n.t('administration.tableOfThematics.headerLabel')}
-      />
-    </div>
-  );
-};
+class ConfigureThematicForm extends React.PureComponent<Props> {
+  constructor(props: Props) {
+    super(props);
+    const { values, thematicId, slug } = this.props;
```
MilanBarande: since you have access to 'props' as the constructor's argument here isn't it better to use instead of this.props ? No clue if this makes any performance difference or difference at all

amensouissi: no difference...

PR #566, opened 2018-10-23
Restore predefined dump devas 1875

amensouissi:
Restore and update the latest database. Only for the staging server.

### DIFF
--- a/assembl/configs/base_env.rc
+++ b/assembl/configs/base_env.rc
@@ -15,8 +15,9 @@ piwik_host = localhost
 url_metadata_host = localhost
 changes_websocket_host = localhost
 # AWS S3 config
-aws_access_key_id=
-aws_secret_access_key=
+aws_bucket_name =
+aws_access_key_id =
+aws_secret_access_key =
 # The port to use for the websocket (client frontends will connect to this)
 # In prod, your firewall needs to allow this through or proxy it through nginx
 # UNIQUE_PER_SERVER
--- a/assembl/configs/base_env.rc
+++ b/assembl/configs/base_env.rc
@@ -14,6 +14,9 @@ imap_host = localhost
 piwik_host = localhost
 url_metadata_host = localhost
 changes_websocket_host = localhost
+# AWS S3 config
+aws_access_key_id=
+aws_secret_access_key=
 # The port to use for the websocket (client frontends will connect to this)
 # In prod, your firewall needs to allow this through or proxy it through nginx
 # UNIQUE_PER_SERVER
--- a/assembl/fabfile.py
+++ b/assembl/fabfile.py
@@ -816,7 +816,7 @@ def reset_db():
     Restore and update the latest database
     """
     # Only for the staging server (fro tests)
-    if env.wsginame == 'dev.wsgi':
+    if env.wsginame == 'staging.wsgi':
         exists = False
         # Retrieve the symbolic link of the dump
         with shell_env(
--- a/assembl/fabfile.py
+++ b/assembl/fabfile.py
@@ -801,17 +801,22 @@ def reset_db():
     """
     # Only for the staging server (fro tests)
     if env.wsginame == 'staging.wsgi':
+        print(green('Restore and update the latest database'))
         # Retrieve the symbolic link of the dump
         path = join(env.projectpath, remote_db_path())
         if not exists(path):
             # If the dump don't exist, we create it
+            print(cyan('Create a new dump'))
             execute(database_dump)
         else:
             # Otherwise, we restore the last dump
+            print(cyan('Restore the last dump'))
             execute(database_restore)
 
+        print(cyan('Update the restored db'))
         execute(app_db_update)
         # Create the updated dump for future tests
+        print(cyan('Create the updated dump for future tests'))
         execute(database_dump)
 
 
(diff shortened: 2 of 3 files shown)


assembl/configs/base_env.rc
```diff
@@ -14,6 +14,10 @@ imap_host = localhost
 piwik_host = localhost
 url_metadata_host = localhost
 changes_websocket_host = localhost
+# AWS S3 config
+aws_bucket_name =
+aws_access_key_id =
```
moutard-bluenove: Keys are for developers to connect S3 using their personal account. We shouldn't do that here.  It would be better to use an aws role, (temporary permission)  that allows hosts (like  EC2 or server) to access a s3 with their own credentials (based on IP for instance)

assembl/fabfile.py
```diff
@@ -794,6 +801,155 @@ def update_pip_requirements(force_reinstall=False):
...
+        exists = False
+        # Test if the dump exists on the amazon s3 object storage
+        with shell_env(
+            AWS_ACCESS_KEY_ID=env.aws_access_key_id,
+            AWS_SECRET_ACCESS_KEY=env.aws_secret_access_key
+        ):
+            exists = venvcmd('aws s3 ls s3://%s/%s | wc -l' % (env.aws_bucket_name, get_db_dump_name()))
+            exists = False if exists == '0' else True
+
+        print(green('Restore and update the latest database'))
```
moutard-bluenove: Can we use a real logger. https://docs.python.org/2/library/logging.html
This backup should be run automatically and we would like to be able to debug the logs if they don't run properly.

assembl/fabfile.py
```diff
@@ -666,6 +666,13 @@ def get_db_dump_name():
     return 'assembl-backup.pgdump'
 
 
+def get_versioned_db_dump_name():
+    bumpversion_config = SafeConfigParser()
```
moutard-bluenove: I think the fabfile is already quite big, can we put the backup part in a different file?

assembl/fabfile.py
```diff
@@ -794,6 +801,155 @@ def update_pip_requirements(force_reinstall=False):
...
+    """
+    # Only for the staging server (for tests)
+    if env.wsginame == 'staging.wsgi':
+        exists = False
+        # Test if the dump exists on the amazon s3 object storage
+        with shell_env(
+            AWS_ACCESS_KEY_ID=env.aws_access_key_id,
+            AWS_SECRET_ACCESS_KEY=env.aws_secret_access_key
+        ):
+            exists = venvcmd('aws s3 ls s3://%s/%s | wc -l' % (env.aws_bucket_name, get_db_dump_name()))
```
moutard-bluenove: I tend to prefer .format() more readable, do we have a guideline for that?

assembl/fabfile.py
```diff
@@ -794,6 +801,155 @@ def update_pip_requirements(force_reinstall=False):
...
+        run('rm -f %s' % dump_path)
+
+
+@task
+def database_restore_aws(backup=False):
+    """
+    Restores the database backed up on the amazon s3 object storage
+    """
+    if not backup:
+        assert(env.wsginame in ('staging.wsgi', 'dev.wsgi'))
```
moutard-bluenove: what's the difference between staging and dev I thought there is only one host?

assembl/fabfile.py
amensouissi: I agree with you. the fab file uses this syntax ... Done.

assembl/fabfile.py
amensouissi: je suis pour un éclatement et un nettoyage globale du fabfile... à discuter

assembl/fabfile.py
amensouissi: je suis pour un éclatement et un nettoyage globale du fabfile... à discuter

assembl/fabfile.py
amensouissi: it's for the local dev

moutard-bluenove:
ok sounds good.

assembl/fabfile.py
```diff
@@ -794,6 +801,155 @@ def update_pip_requirements(force_reinstall=False):
...
+        processes = filter_autostart_processes([
```
ayyazdaniaryan: This code is complex to maintain. It's anti-dry and in past, there isn't enough wait time to recover for these services turning off. I suggest instead of this management, turn off supervisor (for now). A restore of DB, under current conditions, will have downtime regardless (until we fully introduce rolling updates). So kill supervisor, wait 30s, then go about your restore. Then restart the `supervisord`.

assembl/fabfile.py
```diff
@@ -794,6 +801,155 @@ def update_pip_requirements(force_reinstall=False):
...
+
+    for process in processes:
+        supervisor_process_stop(process)
+
+    # Kill postgres processes in order to be able to drop tables
+    # execute(postgres_user_detach)
+
+    # Drop db
+    with settings(warn_only=True):
+        dropped = run('PGPASSWORD={} dropdb --host={} --username={} --no-password {}'.format(
```
ayyazdaniaryan: let's use `shell_env` here too.

assembl/fabfile.py
amensouissi: Supervisor is restarted automatically each ~5 seconds

PR #567, opened 2018-10-23
Set a threshold a jest coverage and add coverage check on frontend gitlab job

paolina92:
We have to decide a coverage percent which is acceptable. This value is defined in the jest settings in the `package.json`


paolina92:
I tried to work on this tasks but I'm not sure that's correct. Especially with the yml file.
I decided to check the jest coverage in Travis but we could do that in the pre-commit.
Let me know what you prefer guys.

cedricmessiant:
A few remarks:
- it should go in a gitlab job instead of travis (you can create a new job `check_coverage` in the `test` stage)
- in https://www.npmjs.com/package/jest-coverage-validator it says to use it with `jest-coverage-ratchet` to do exactly what we want (it change the thresholds in the file if coverage is higher). If so, you will need to commit/push the new coverage from the gitlab job
- we want this also for python tests (that's why coveralls may be a good idea)

MilanBarande:
Sorry there seem to be an issue with the linter of my VS Code, causing a lot of noise on the PR. Going to fix that

.pre-commit-config.yaml
```diff
@@ -1,42 +1,47 @@
 - repo: local
   hooks:
-#  - id: pytest
```
amensouissi: only prettier!

.travis.yml
```diff
@@ -3,7 +3,7 @@ sudo: required
 dist: trusty
 addons:
   ssh_known_hosts: localhost
-  postgresql: "9.5"
```
amensouissi: prettier!

assembl/static2/package.json
```diff
@@ -146,6 +148,17 @@
   },
   "//": "See https://facebook.github.io/jest/docs/webpack.html for more info",
   "jest": {
+    "coverageReporters": [
+      "json-summary"
+    ],
+    "coverageThreshold": {
+      "global": {
+        "lines": 53.04,
```
amensouissi: current values! we need to update them! we have to update them manually!

assembl/static2/package.json
```diff
@@ -146,6 +148,17 @@
   },
   "//": "See https://facebook.github.io/jest/docs/webpack.html for more info",
   "jest": {
+    "coverageReporters": [
```
amensouissi: only for front-end!

assembl/static2/package.json
MilanBarande: No, it's done automatically with jest-coverage-ratchet

.travis.yml
MilanBarande: Yes, I need to fix my prettier config :)

assembl/static2/package.json
MilanBarande: Not sure what you mean, this is jest configuration so it obviously is only about frontend. Like I said this PR is only about frontend and it was decided that backend will be taken care of separately. Gitlab is not compatible with coveralls sadly

amensouissi: :+1:

amensouissi: we need to push the new values on git!? it's dangerous...

MilanBarande: Why is that?

assembl/static2/yarn.lock
```diff
@@ -5,12 +5,10 @@
 "@2fd/command@^1.1.2":
   version "1.1.2"
   resolved "https://registry.yarnpkg.com/@2fd/command/-/command-1.1.2.tgz#b831d511bdd3b4e2d88139aaddb06d3d6aa0b08a"
-  integrity sha1-uDHVEb3TtOLYgTmq3bBtPWqgsIo=
```
cedricmessiant: this is strange. Your `yarn` may be out of date. Try to update it, rebase develop, re-run `yarn` and verify that it is fixed

PR #574, opened 2018-10-29
Fix: post translation in survey and thread exports

cedricmessiant:
I would prefer to add a new `TranslationTable` type and specify the locale to which we want to translate to avoid unnecessary translations (here we translate the post to each language of the discussion but we only need the target locale that have been chosen by the user). What do you think about it?

Also, I'm not sure about the change in `assembl/tasks/translate.py`

### DIFF
--- a/assembl/models/generic.py
+++ b/assembl/models/generic.py
@@ -454,15 +454,17 @@ class Content(TombstonableMixin, DiscussionBoundBase):
             log.error("What is this mimetype?" + mimetype)
             return body
 
-    def maybe_translate(self, pref_collection=None):
+    def maybe_translate(self, pref_collection=None, target_locales=None):
         from assembl.tasks.translate import (
-            translate_content, PrefCollectionTranslationTable)
+            translate_content, PrefCollectionTranslationTable, LanguagesTranslationTable)
         service = self.discussion.translation_service()
         if service.canTranslate is not None:
             translations = None
             if pref_collection:
                 translations = PrefCollectionTranslationTable(
                     service, pref_collection)
+            elif target_locales:
+                translations = LanguagesTranslationTable(service, target_locales)
 
             translate_content(
                 self, translation_table=translations, service=service)
--- a/assembl/models/generic.py
+++ b/assembl/models/generic.py
@@ -454,13 +454,16 @@ class Content(TombstonableMixin, DiscussionBoundBase):
             log.error("What is this mimetype?" + mimetype)
             return body
 
-    def maybe_translate(self, pref_collection):
+    def maybe_translate(self, pref_collection=None):
         from assembl.tasks.translate import (
             translate_content, PrefCollectionTranslationTable)
         service = self.discussion.translation_service()
         if service.canTranslate is not None:
-            translations = PrefCollectionTranslationTable(
-                service, pref_collection)
+            translations = None
+            if pref_collection:
+                translations = PrefCollectionTranslationTable(
+                    service, pref_collection)
+
             translate_content(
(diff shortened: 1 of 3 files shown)


assembl/models/generic.py
```diff
@@ -454,13 +454,16 @@ def get_body_as_text(self):
             log.error("What is this mimetype?" + mimetype)
             return body
 
-    def maybe_translate(self, pref_collection):
+    def maybe_translate(self, pref_collection=None):
```
ayyazdaniaryan: Hmm, maybe this is a way to do it. But, I would expect that `pref_collection` to always be passed. In fact, under every request, no matter the caller, there is a `request.lang_prefs`. You can use the `setCurrentFromLocale` with the `has_lang` locale. And then pass this collection that is on the request to `maybe_translate`. It's a bit spaghetti, I know. But I feel that if you need to force a lang_collection, you can force it for the request, and of course, pass it to the translate.

ayyazdaniaryan:
Also, the `TranslationTable` that you seek already exists, in `UserLanguagePreference`. There is a `translate_to` column, which indicates what the specific user wants a language to be translated to. Mind you, there isn't a similar table discussion-wide, however.

assembl/models/generic.py
ayyazdaniaryan: Look in `assembl/models/auth.py` for `LanguagePreferenceCollection`.

maparent: This was actually my suggestion; without the pref_collection, we get the discussion's table, which translates to all discussion languages (which is what he needs for the report.)

ayyazdaniaryan: Hmm, alright fair. So, not biasing against a userpref, but instead all supported languages of a debate.

cedricmessiant: @maparent @ayyazdaniaryan I need only one language in fact (that is chosen by the user that requests the export so I'd prefer to translate only to this locale (like I said in my first comment: https://github.com/assembl/assembl/pull/574#issue-226614165). What would be the best way to do it in your opinion?

maparent: Suggestion: current (admin) user presumably has a target language. Use his prefs.

assembl/tasks/translate.py
```diff
@@ -14,7 +14,26 @@ def languages_for(self, locale_code, db=None):
         return ()
 
 
+class LanguagesTranslationTable(TranslationTable):
```
maparent: Sorry I did not think about it initially, this could have been done with a PrefCollectionTranslationTable using a LanguagePreferenceCollectionWithDefault... I don't think it's worth changing now. And your version handles multiple targets, which is better.

assembl/tasks/translate.py
```diff
@@ -37,8 +57,12 @@ def __init__(self, service, discussion):
         self.base_languages.discard(None)
 
     def languages_for(self, locale_code, db=None):
+        content_locale = set()
         locale_code = self.service.asKnownLocale(locale_code)
-        return self.base_languages - set(locale_code)
```
maparent: is asKnownLocale returning null a real case? When? Then maybe it should be debugged upstream.

maparent:
OK, I have questions about null asKnownLocale, but otherwise LGTM.

assembl/tasks/translate.py
cedricmessiant: yes, it was `zxx` and `und`

assembl/tasks/translate.py
cedricmessiant: Thanks, I'll merge then

PR #575, opened 2018-10-29
One click prod

MilanBarande:
The point here is simply to have the button on gitlab to manually deploy to production servers. As the deployment script is not ready yet, we just set up a for loop to iterate over a fake list of hosts which we fetch from the gitlab environment variables

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -68,7 +68,6 @@ deploy_to_dev:
   stage: deploy_to_dev
   only:
     - tmp_master_for_ci_cd
-    - master
   script:
     # empty script for now
     - sh scripts/echo.sh
@@ -77,7 +76,6 @@ deploy_to_uat:
   stage: deploy_to_uat
   only:
     - tmp_master_for_ci_cd
-    - master
   script:
     # empty script for now
     - sh scripts/echo.sh
@@ -87,7 +85,6 @@ deploy_to_prod:
   environment: PROD
   only:
     - tmp_master_for_ci_cd
-    - master
   script:
     # fake script for now
     - python scripts/deploy_to_prod.py
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -83,7 +83,7 @@ deploy_to_uat:
     - sh scripts/echo.sh
 
 deploy_to_prod:
-  stage: depoy_to_prod
+  stage: deploy_to_prod
   environment: PROD
   only:
     - tmp_master_for_ci_cd
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -86,7 +86,7 @@ deploy_to_prod:
   stage: depoy_to_prod
   environment: PROD
   only:
-    - deploy_to_uat
+    - tmp_master_for_ci_cd
     - master
   script:
     # fake script for now
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -87,4 +87,4 @@ deploy_to_prod:
   script:
     # empty script for now
     - python scripts/deploy_to_prod.py
-  when: manual
+  # when: manual
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -65,26 +65,26 @@ build_wheel:
     - twine upload -u bluenove --skip-existing dist/*
 
 deploy_to_dev:
-  stage: deploy_to_dev
+  stage: tmp_master_for_ci_cd
   only:
-    - develop
+    - master
   script:
     # empty script for now
     - sh scripts/echo.sh
 
 deploy_to_uat:
-  stage: deploy_to_uat
+  stage: tmp_master_for_ci_cd
   only:
-    - develop
+    - master
   script:
     # empty script for now
     - sh scripts/echo.sh
 
 deploy_to_prod:
-  stage: deploy_to_prod
+  stage: tmp_master_for_ci_cd
   environment: PROD
   only:
-    - develop
+    - master
   script:
     # fake script for now
     - python scripts/deploy_to_prod.py
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -8,10 +8,10 @@ stages:
   - deploy_to_uat
   - deploy_to_prod
 
-cache:
-  paths:
-    - assembl/static2/node_modules/
-    - /usr/local/share/.cache/yarn/v2
+# cache:
+#   paths:
+#     - assembl/static2/node_modules/
+#     - /usr/local/share/.cache/yarn/v2
 
 test_frontend:
   stage: test
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -65,25 +65,28 @@ build_wheel:
     - twine upload -u bluenove --skip-existing dist/*
 
 deploy_to_dev:
-  stage: tmp_master_for_ci_cd
+  stage: deploy_to_dev
   only:
+    - tmp_master_for_ci_cd
     - master
   script:
     # empty script for now
     - sh scripts/echo.sh
 
 deploy_to_uat:
-  stage: tmp_master_for_ci_cd
+  stage: deploy_to_uat
   only:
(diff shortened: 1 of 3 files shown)


.gitlab-ci.yml
```diff
@@ -62,3 +62,29 @@ build_wheel:
     # create the wheel and upload to pypi
     - sh scripts/create_assembl_wheel.sh
     - twine upload -u bluenove --skip-existing dist/*
+
+deploy_to_dev:
+  stage: deploy_to_dev
+  only:
+    - develop
```
cedricmessiant: it will be `tmp_master_for_ci_cd` and then master here

.gitlab-ci.yml
```diff
@@ -62,3 +62,29 @@ build_wheel:
...
+  script:
+    # empty script for now
+    - sh scripts/echo.sh
+
+deploy_to_uat:
+  stage: deploy_to_uat
+  only:
+    - develop
```
cedricmessiant: here also

.gitlab-ci.yml
```diff
@@ -62,3 +62,29 @@ build_wheel:
...
+  script:
+    # empty script for now
+    - sh scripts/echo.sh
+
+deploy_to_prod:
+  stage: deploy_to_prod
+  environment: PROD
+  only:
+    - develop
```
cedricmessiant: and here

scripts/deploy_to_prod.py
```diff
@@ -0,0 +1,11 @@
+import os
+import json
+
+prod_clients_json = os.environ['PROD_CLIENTS']
+
+prod_clients = json.loads(prod_clients_json)
+
+for client in prod_clients["prod_clients"]:
+    # We just print a fake list of hostnames for now
+    print(client["hostname"])
```
cedricmessiant: it would be nice here to reuse the script that we use in `deploy_to_dev` and `deploy_to_uat` jobs. Also, you should try to connect on an actual server (via ssh) and run a command on this server (for example `echo $HOSTNAME`)

.gitlab-ci.yml
MilanBarande: makes sense

scripts/deploy_to_prod.py
MilanBarande: We want to use the script that @ayyazdaniaryan is currently writing for prod deployment

cedricmessiant: yeah, I know. I meant use the `scripts/echo.sh` to prove we can do it but I don't think it worth it since in the end it will depend on how the deploy script is done. 
But you could definitely do the ssh part (echo $HOSTNAME) on all servers defined in `prod_clients`

MilanBarande: Allright sure :)

cedricmessiant:
About re-deploy, rollbacks and environments: https://docs.gitlab.com/ee/ci/environments.html
The deploy script should take the assembl version as a param to ensure we deploy the good wheel (it is in the `.bumpversion.cfg` file and `setup.py`). It a more a reminder than a request for this particular PR

ayyazdaniaryan:
@cedricmessiant indeed, the deploy script should take the version. But, determining that, where is the best place for it. Is it in the .bumpversion config or the setup.py now?

moutard-bluenove:
@cedricmessiant I understand your point, however I think we should drive the business here. No client is unique we sell the same software the only difference is config. So deployment is pushed to all. That will force us to be rigourous.
Moreover for the rollback I would wait until we get the paid version, then we could use canary deployment. If I understand correctly it does rollback too.

cedricmessiant:
@ayyazdaniaryan `.bumpversion.cfg` and `setup.py` should always be in sync. If they are not, we screwed! So what I would do is take it from `.bumpversion.cfg` file, it should be much easier

.gitlab-ci.yml
```diff
@@ -62,3 +62,29 @@ build_wheel:
     # create the wheel and upload to pypi
     - sh scripts/create_assembl_wheel.sh
     - twine upload -u bluenove --skip-existing dist/*
+
+deploy_to_dev:
+  stage: tmp_master_for_ci_cd
+  only:
+    - master
```
amensouissi: stage: deploy_to_dev
only: tmp_master_for_ci_cd
???

.gitlab-ci.yml
```diff
@@ -62,3 +62,29 @@ build_wheel:
...
+  script:
+    # empty script for now
+    - sh scripts/echo.sh
+
+deploy_to_uat:
+  stage: tmp_master_for_ci_cd
```
amensouissi: and here

.gitlab-ci.yml
```diff
@@ -62,3 +62,29 @@ build_wheel:
...
+  only:
+    - master
+  script:
+    # empty script for now
+    - sh scripts/echo.sh
+
+deploy_to_prod:
+  stage: tmp_master_for_ci_cd
```
amensouissi: and here

.gitlab-ci.yml
MilanBarande: Oh yeah I misunderstood @cedricmessiant 's comment :) I wasn't fully awake I think

amensouissi: mater and tmp_master_for_ci_cd ?

.gitlab-ci.yml
```diff
@@ -62,3 +62,32 @@ build_wheel:
...
+    - master
+  script:
+    # empty script for now
+    - sh scripts/echo.sh
+
+deploy_to_prod:
+  stage: depoy_to_prod
+  environment: PROD
+  only:
+    - deploy_to_uat
```
amensouissi: deploy_to_uat is a tag, a branche...???

cedricmessiant: :+1: it should be `tmp_master_for_ci_cd` here

MilanBarande: 🤦‍♂️ I think I'm really tired today, sorry for the mess, will fix this

.gitlab-ci.yml
MilanBarande: Well yes isn't it what you said in your review @cedricmessiant ? Or should it be just the tmp branch?

.gitlab-ci.yml
```diff
@@ -62,3 +62,32 @@ build_wheel:
...
+  stage: deploy_to_uat
+  only:
+    - tmp_master_for_ci_cd
+    - master
+  script:
+    # empty script for now
+    - sh scripts/echo.sh
+
+deploy_to_prod:
+  stage: depoy_to_prod
```
amensouissi: deploy_to_prod

.gitlab-ci.yml
```diff
@@ -63,3 +63,32 @@ build_wheel:
     # create the wheel and upload to pypi
     - sh scripts/create_assembl_wheel.sh
     - twine upload -u bluenove --skip-existing dist/*
+
+deploy_to_dev:
+  stage: deploy_to_dev
+  only:
+    - tmp_master_for_ci_cd
+    - master
```
cedricmessiant: @MilanBarande you should remove `master` from now

.gitlab-ci.yml
```diff
@@ -63,3 +63,32 @@ build_wheel:
...
+  script:
+    # empty script for now
+    - sh scripts/echo.sh
+
+deploy_to_uat:
+  stage: deploy_to_uat
+  only:
+    - tmp_master_for_ci_cd
+    - master
```
cedricmessiant: same here

.gitlab-ci.yml
```diff
@@ -63,3 +63,32 @@ build_wheel:
...
+  script:
+    # empty script for now
+    - sh scripts/echo.sh
+
+deploy_to_prod:
+  stage: deploy_to_prod
+  environment: PROD
+  only:
+    - tmp_master_for_ci_cd
+    - master
```
cedricmessiant: and here

.gitlab-ci.yml
amensouissi: plus logique :)

PR #579, opened 2018-11-05
bug fixing: DEVAS-1904

paolina92:
- Create a « Medias » component to manage different medias types
- Rename the « Media » component to « TextAndMedia »
- Write unit tests


assembl/static2/js/app/components/common/medias.jsx
```diff
@@ -0,0 +1,20 @@
...
+import { videoUtilities } from '../../utils/videoUtilities';
+
+export type Props = {
+  path: string
+};
+
+export const Medias = ({ path }: Props) => (
+  <Fragment>
+    {videoUtilities.pathIsVideoFile(path) ? (
+      <video src={path} controls preload="none" width="100%" /> // eslint-disable-line
```
MilanBarande: Why eslint-disable-line :) ?

paolina92: The error message is: "Média elements such as a <audio> ans <video> must have a <track> for captions". Any ideas?

assembl/static2/js/app/components/home/video.jsx
```diff
@@ -3,6 +3,8 @@ import { connect } from 'react-redux';
 import { Translate } from 'react-redux-i18n';
 import { Grid, Row, Col } from 'react-bootstrap';
 
+import Medias from '../common/medias';
+
 class Video extends React.Component {
```
MilanBarande: // @flow :)

assembl/static2/js/app/components/home/video.jsx
```diff
@@ -3,6 +3,8 @@ import { connect } from 'react-redux';
...
   constructor(props) {
```
MilanBarande: constructor not needed, you can do ```state = { isTextHigher: false }```

paolina92: It's an old file habibi. Not in the scope

assembl/static2/js/app/components/home/video.jsx
paolina92: Same here. It's an old file habibi. Not in the scope

assembl/static2/js/app/utils/videoUtilities.js
```diff
@@ -0,0 +1,54 @@
...
+  'f4b'
+];
+
+export const videoUtilities = {
+  pathIsVideoFile: (path) => {
+    if (!path || typeof path !== 'string') return false;
+    const components = path.split('/');
+    const file = components[components.length - 1].split('.');
+    const extension = file[file.length - 1];
+    return videoExtensions.indexOf(extension) > -1;
```
MilanBarande: you can do ```videoExtensions.includes(extension)``` in ES6

assembl/static2/tests/unit/components/common/medias.spec.jsx
```diff
@@ -0,0 +1,32 @@
...
+  beforeEach(() => {
+    wrapper = shallow(<Medias {...MediasProps} />);
+  });
+
+  it('should display a video tag when the path is a mp4', () => {
+    expect(wrapper.find('video')).toHaveLength(1);
+    expect(wrapper.find('iframe')).toHaveLength(0);
+  });
+
+  it('should display a iframe tag when the path an embed video', () => {
```
MilanBarande: ```an iframe tag when the path is an embed video```

maybe you can add one test for the expected result when the path is not a correct video path

assembl/static2/tests/unit/components/common/medias.spec.jsx
```diff
@@ -0,0 +1,32 @@
...
+};
+
+describe('<Medias /> - with shallow', () => {
+  let wrapper;
+
+  beforeEach(() => {
+    wrapper = shallow(<Medias {...MediasProps} />);
+  });
+
+  it('should display a video tag when the path is a mp4', () => {
```
MilanBarande: ```an mp4 file```

au début je n'avais pas compris les 2 tests donc désolé je chipote sur leur nom

assembl/static2/tests/unit/utils/videoUtilities.spec.js
```diff
@@ -0,0 +1,27 @@
+import { videoUtilities } from '../../../js/app/utils/videoUtilities';
+
+describe('should test if the path is a video file', () => {
```
MilanBarande: This is a test assertion, describe should be something like ```pathIsVideoFile function```

assembl/static2/js/app/components/home/video.jsx
MilanBarande: we try to refactor every file we touch to clean the codebase though :/ and this takes like 1 sec but as you wish, not important

assembl/static2/js/app/components/home/video.jsx
MilanBarande: We said we try to add flow on 10 files per dev, there is no ticket for that, it's just something you do when you touch a file where it's missing. At least I would take note that it needs to be done on video.jsx

paolina92: I'm gonna do it now ;) I have time...

assembl/static2/js/app/components/common/medias.jsx
MilanBarande: Ok forget it, this is about adding text info about the video for deaf users, you would have to add a <track> element with a dybnamic src set by the video uploader I guess... so clearly out of scope

amensouissi: Currently, there are 3 supported video formats for the <video> element: MP4, WebM, and Ogg...
https://www.w3schools.com/tags/tag_video.asp

amensouissi: c'est pour l'accessibilité

assembl/static2/js/app/components/home/video.jsx
amensouissi: componentWillReceiveProps => getDerivedStateFromProps

assembl/static2/js/app/utils/videoUtilities.js
```diff
@@ -0,0 +1,54 @@
...
+  'mxf',
+  'roq',
+  'nsv',
+  'f4v',
+  'f4p',
+  'f4a',
+  'f4b'
+];
+
+export const videoUtilities = {
```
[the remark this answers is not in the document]
paolina92: je n'ai pas le mimtype du fichier. Je me contente de parser une url. Ce que tu dis implique un refactoring qui me parait hors scope.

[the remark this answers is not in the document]
ayyazdaniaryan: I think setting the mime-type would be the advanced way. @amensouissi , it's possible to extend the url_metadata service to serve the HEAD request of a URL in order to get the mime-type (if available, whilst avoiding CORS issues). So you can make an API HEAD call of the URL in question to get the mime-type. I think for this bug, this is out of scope (agree with Pauline here). However, there should be a TODO here for this.

assembl/static2/js/app/utils/videoUtilities.js
```diff
@@ -0,0 +1,54 @@
...
+  'svi',
+  '3gp',
+  '3g2',
+  'mxf',
+  'roq',
+  'nsv',
+  'f4v',
+  'f4p',
+  'f4a',
+  'f4b'
```
ayyazdaniaryan: Okay, I would cut this video into 2 types. VIDEO tag supported types, and OTHER. Here are the video tag supported types: https://developer.mozilla.org/en-US/docs/Web/HTML/Supported_media_formats

You can have `videoExtensions = allVideos + videoTagSupported`. And for EDF, to be accurate, have the utility only look at the videoTagSupported list.

assembl/static2/js/app/utils/videoUtilities.js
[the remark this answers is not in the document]
amensouissi: :+1:

assembl/static2/js/app/utils/videoUtilities.js
amensouissi: not here. It's supposed to be independent of the tag

PR #582, opened 2018-11-07
Use sentry-sdk for python and add sentry integration for javascript

cedricmessiant:
TODO:
- [x] remove all references to `raven`
- [x] fix config files
- [x] add breadcrumbs for js and python (probably in another PR). For example: which graphql query/mutation has been called


assembl/templates/sentry.jinja2
```diff
@@ -0,0 +1,5 @@
+<script src="https://browser.sentry-cdn.com/4.2.4/bundle.min.js" crossorigin="anonymous"></script>
```
ayyazdaniaryan: Hmm, can we do better? We're breaking centrality of package definitions by doing this (and not in yarn/package.json. On other hand, putting it on yarn won't work with our future CI/CD approach 😢. Any other suggestions?

cedricmessiant: I agree. I don't know how I can install it and import it in the jinja template. I tried but I didn't want to spend to much time on it. If you have an idea, I'll do the change with pleasure :-)

cedricmessiant:
Refs DEVAS-1880

assembl/lib/sentry.py
```diff
@@ -0,0 +1,21 @@
...
+def is_sentry_enabled():
+    return bool(get('sentry_dsn', ''))
+
+
+def capture_message(*args, **kwargs):
+    if is_sentry_enabled():
+        sentry_sdk.capture_message(*args, **kwargs)
+
+
+def capture_exception(*args, **kwargs):
```
amensouissi: args and kwargs are not used here (where is the error/exception here?). I think you have to get the POST/GET from the request object (filter emails and passwords)

assembl/static/js/app/views/baseMessageList.js
```diff
@@ -731,12 +731,12 @@ return cls.extend({
         return true;
       });
     }).catch(function(e) {
-      if (raven_url) {
+      if (sentry_dsn) {
```
amensouissi: Raven? it's removed?

assembl/static2/js/app/components/common/GlobalErrorBoundary.jsx
```diff
@@ -15,7 +18,13 @@ export default class GlobalErrorBoundary extends React.Component<Props, State> {
 
   componentDidCatch(error: Error, info: { componentStack: string }) {
     this.setState({ error: error });
-    // TODO: log the error to sentry
+    Sentry.withScope((scope) => {
+      Object.keys(info).forEach((key) => {
+        scope.setExtra(key, info[key]);
```
amensouissi: we have to filter passwords and emails here ...

cedricmessiant: I don't think we have emails here (and I'm pretty sure we don't have passwords). `info` is basically the stack trace from react / js.

assembl/templates/sentry.jinja2
ayyazdaniaryan: Sigh, only thing I can think of, which is a huge hassle, is to put this into package.json/yarn.lock, and import it into Assembl and introduce code splitting in webpack. I'm not even sure this would work, as the DOM will be making JS calls on an object that might not yet be loaded. The bundle.min.js on my machine is 91K.

I'm not certain there is any easier way to load this, unless we create an API endpoint to pass the sentry DSN/secret through before react init, and not use the template provided by Pyramid.

assembl/static/js/app/views/baseMessageList.js
cedricmessiant: Raven is the old sentry sdk. After merge, we will still use it for v1 js but not for python and static2. Here it is the variable from the rc/ini files

assembl/static2/js/app/components/common/GlobalErrorBoundary.jsx
amensouissi: et si erreur graphql, nous avons besoin de récupérer les valeurs du POST ou du GET...

assembl/static/js/app/views/baseMessageList.js
amensouissi: cool ok :+1:

assembl/lib/sentry.py
cedricmessiant: oh yes, I should pass args and kwargs to `capture_exception` like it was before. If exception is not given to `sentry_sdk.capture_exception`, it is taken from `sys.exc_info()`.
About adding the POST/GET from the request, I think we should add those as breadcrumbs but I would like to do it in another PR so we can benefit from the update asap

amensouissi: yes :+1:  then, you have to create a ticket :)

assembl/static2/js/app/components/common/GlobalErrorBoundary.jsx
ayyazdaniaryan: You'll get user id (that's what you got with raven), coming from the request. Otherwise, no passwords or requests should be a key here. And yes the main thing that we need to capture, actually, for sentry, is the root of the exception.

ayyazdaniaryan: @amensouissi assembl/graphql/view.py:20-24

cedricmessiant: yes, I would like to at least have the name of the query / mutation

assembl/templates/sentry.jinja2
cedricmessiant: if the sentry dsn is not sensitive data, I could put it in a data attribute in the DOM and retrieve in it React code (to avoid to create an API endpoint for this but even with the API endpoint, this data is exposed).

assembl/static2/js/app/components/common/GlobalErrorBoundary.jsx
cedricmessiant: @ayyazdaniaryan @amensouissi  about `LoggingMiddleware`, it is in fact not used. See lines 98 to 100.

assembl/static2/js/app/client.js
```diff
@@ -27,6 +29,31 @@ const myFragmentMatcher = new IntrospectionFragmentMatcher({
...
+  opts: {
+    credentials: 'same-origin'
+  }
+});
+
+// trace every graphql operation in sentry
+networkInterface.use([
+  {
+    applyMiddleware: function (req, next) {
+      Sentry.addBreadcrumb({
```
amensouissi: oups!! toutes les queries et mutations... Les variables en plus... là vraiment il faut cacher les mots de passe et les mails...

assembl/static2/js/app/client.js
```diff
@@ -27,6 +29,31 @@ const myFragmentMatcher = new IntrospectionFragmentMatcher({
...
 
+const networkInterface = createNetworkInterface({
+  uri: getFullPath('graphql', { slug: getDiscussionSlug() }),
+  opts: {
+    credentials: 'same-origin'
+  }
+});
+
+// trace every graphql operation in sentry
+networkInterface.use([
```
ayyazdaniaryan: Hmm, not choice. This is a choice and perhaps should be discussed. Because the backend is logging all of this in the assembl.log file, line by line, as JSON, as a middleware too. But, there isn't a web-reader available for it. Are we going to log in both places? Also, what about the spamminess of these logs on Sentry. We've usually used Sentry to log for warn level and up.

assembl/templates/includes/javascript.jinja2
```diff
@@ -5,7 +5,7 @@
   var urlArgs = null;
   {% endif %}
   var assembl_locale = '{{locale}}';
-  var raven_url = '{{raven_url}}';
+  var sentry_dsn = '{{sentry_dsn}}';
```
ayyazdaniaryan: I think you can remove this now.

assembl/static2/js/app/client.js
cedricmessiant: If we want clear Graphql errors, we need this here because the error is reported by the js sentry client and not the python one (well in fact, it depends because for mutations decorated with `abort_transaction_on_exception` it is done on the python side).

Without this, we are blind about what cause the error (we don't know the query, we can find the component but it does not help a lot).

My guess is that the best would be to abort transactions also for queries and report the error in the sentry python sdk. We have a big work to do on reporting as the whole. But until then, I think we should merge this to be able to debug errors in the meanwhile.

assembl/templates/includes/javascript.jinja2
cedricmessiant: I think we need it for v1 js reporting (raven)

ayyazdaniaryan: You're right 👍

assembl/static2/js/app/client.js
ayyazdaniaryan: Yeah, I thought about it some more, and I agree with you. Keep this in. We haven't made any decision (ie, putting the stacktrace on sentry for the abort_on_transaction). So, until we do a reporting user story, this stays.

assembl/static2/js/app/client.js
cedricmessiant: after some tests, I think that the only case we could have email in variables is the creation of a user but it is done via the old api

cedricmessiant: and passwords are marked as `[Filtered]` by sentry

PR #585, opened 2018-11-08
DEVAS 1940 bugfix survey admin

paolina92:
Now it’s forbidden to change the module if the phase already exists.

- The select box to set a module is disabled if the phase is marked as _isNew=false
- `isThematicsTable` is set in the frontend, even if it's not yet used in the backend, just to not forget for the future
- The backend don't return orphan ideas

### DIFF
--- a/assembl/static2/js/app/components/administration/discussion/phaseForm.jsx
+++ b/assembl/static2/js/app/components/administration/discussion/phaseForm.jsx
@@ -27,7 +27,8 @@ type PhaseFormProps = {
   handleEndDateChange: Function,
   handleIdentifierChange: Function,
   locale: string,
-  hasConflictingDates: boolean
+  hasConflictingDates: boolean,
+  isNew: boolean
 };
 
 export const DumbPhaseForm = ({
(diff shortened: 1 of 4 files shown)


assembl/static2/js/app/components/administration/discussion/phaseForm.jsx
```diff
@@ -137,12 +145,16 @@ const mapStateToProps = (state, { phaseId }) => {
     start: phase ? phase.get('start') : null,
     end: phase ? phase.get('end') : null,
     hasConflictingDates: phase ? phase.get('hasConflictingDates') : null,
-    locale: state.i18n.locale
+    locale: state.i18n.locale,
+    isNew: phase.get('_isNew')
```
amensouissi: phase && phase.get('_isNew') ?

assembl/static2/js/app/components/administration/discussion/phaseForm.jsx
```diff
@@ -106,6 +113,7 @@ export const DumbPhaseForm = ({
       <div className="margin-m">
         <SplitButton
           className="admin-dropdown"
+          disabled={!isNew}
```
amensouissi: suggestion: we need a tooltip to explain to the user why it is disabled

assembl/static2/js/app/components/administration/discussion/phaseForm.jsx
paolina92: yes

PR #590, opened 2018-11-09
Upgrade nodejs

cedricmessiant:
Update node version to use last LTS, also always upgrade yarn when we upgrade node


assembl/fabfile.py
```diff
@@ -2674,6 +2676,14 @@ def create_add_to_crontab_command(crontab_line):
     return "(crontab -l | grep -Fv '{cron}'; echo '{cron}') | crontab -".format(cron=crontab_line)
 
 
+@task
```
mohamedzayed92: is there a need for this to be a task ?

cedricmessiant: It is to ensure that yarn is upgraded because Vincent told me that the cron is not configured on every server. It will also avoid to have different versions between developers and thus environment issues.
@vincentfretin @ayyazdaniaryan thoughts?

assembl/static/js/package.json
```diff
@@ -61,7 +61,7 @@
     "gulp-uglify": "~1.5.4",
     "gulp-util": "~3.0.7",
     "js-fixtures": "~1.5.3",
-    "jsdoc-rst-template": "https://github.com/assembl/jsdoc-rst-template",
+    "jsdoc-rst-template": "git+https://github.com/assembl/jsdoc-rst-template.git",
```
ayyazdaniaryan: 👍

assembl/fabfile.py
ayyazdaniaryan: Hmm, we already have `upgrade_yarn_crontab`, which is run on the main server installation path.

vincentfretin: @ayyazdaniaryan on existing servers, the cron is not installed everywhere. On staging: 
```
yarn -v
1.6.0
```
is not the last version.

cedricmessiant:
yes, I tested v1 and v2 builds and start

assembl/fabfile.py
```diff
@@ -1401,8 +1404,10 @@ def update_node(force_reinstall=False):
...
         supervisor_process_stop('dev:gulp')
         supervisor_process_stop('dev:webpack')
+        venvcmd("rm -rf venv/lib/node_modules/")
         venvcmd("rm -f venv/bin/npm")  # remove the symlink first otherwise next command raises OSError: [Errno 17] File exists
-        venvcmd("nodeenv --node=6.11.5 --npm=3.10.10 --python-virtualenv assembl/static/js")
+        venvcmd("nodeenv --node=10.13.0 --npm=6.4.1 --python-virtualenv assembl/static/js")
+        execute(upgrade_yarn)
         with cd(get_node_base_path()):
             venvcmd("npm install reinstall -g", chdir=False)
         with cd(get_new_node_base_path()):
```
vincentfretin: instead of 
```
with cd(get_node_base_path()):
    venvcmd("npm install reinstall -g", chdir=False)
with cd(get_new_node_base_path()):
    venvcmd("npm install reinstall -g", chdir=False)  # which is wrong because this doesn't use yarn for static2
```
you should use 
```
update_npm_requirements(True)
```

assembl/fabfile.py
```diff
@@ -1069,6 +1069,8 @@ def compile_stylesheets():
     """
     with cd(env.projectpath):
         with cd('assembl/static/js'):
+            venvcmd('pwd', chdir=False)
+            venvcmd('ls node_modules/.bin', chdir=False)
```
vincentfretin: You can remove those 2 lines as well.

cedricmessiant: Thanks

PR #591, opened 2018-11-12
bugherd: inject script on preprod server

pascalhuynh:
Quick PR on injecting Bugherd script for BM testing on preprod server only. (Note: Bugherd project key will have to be updated with the one purchase by Bluenove)

### DIFF
--- a/assembl/templates/index_react.jinja2
+++ b/assembl/templates/index_react.jinja2
@@ -50,5 +50,18 @@
                 <div id="under-test" class="hidden">V2</div>
             {%- endif %}
         {% endblock %}
+
+        {% if inject_bugherd %}
+            <!-- Bugherd injection -->
+            <script type='text/javascript'>
+            (function (d, t) {
+            var bh = d.createElement(t), s = d.getElementsByTagName(t)[0];
+            bh.type = 'text/javascript';
+            bh.src = 'https://www.bugherd.com/sidebarv2.js?apikey=mtcl06zwfwhfvmxqxb1d5q';
+            s.parentNode.insertBefore(bh, s);
+            })(document, 'script');
+            </script>
+            <!-- End of Bugherd injection -->
+        {% endif %}
     </body>
 {% endblock%}
--- a/assembl/templates/index_react.jinja2
+++ b/assembl/templates/index_react.jinja2
@@ -49,19 +49,19 @@
             {% if under_test %}
                 <div id="under-test" class="hidden">V2</div>
             {%- endif %}
-        {% endblock %}
 
-        {% if inject_bugherd %}
-            <!-- Bugherd injection -->
-            <script type='text/javascript'>
-            (function (d, t) {
-            var bh = d.createElement(t), s = d.getElementsByTagName(t)[0];
-            bh.type = 'text/javascript';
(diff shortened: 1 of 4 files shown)


assembl/templates/index_react.jinja2
```diff
@@ -50,5 +50,18 @@
...
             {%- endif %}
         {% endblock %}
+
+        {% if inject_bugherd %}
+            <!-- Bugherd injection -->
+            <script type='text/javascript'>
+            (function (d, t) {
+            var bh = d.createElement(t), s = d.getElementsByTagName(t)[0];
+            bh.type = 'text/javascript';
+            bh.src = 'https://www.bugherd.com/sidebarv2.js?apikey=mtcl06zwfwhfvmxqxb1d5q';
```
ayyazdaniaryan: I would put both the URL and the apikey in the rc files. I would put it in comments in base_env.rc (there are a number of commented keys noted to be filled in client rc files).

assembl/templates/index_react.jinja2
```diff
@@ -50,5 +50,18 @@
                 <div id="under-test" class="hidden">V2</div>
             {%- endif %}
         {% endblock %}
+
+        {% if inject_bugherd %}
+            <!-- Bugherd injection -->
+            <script type='text/javascript'>
+            (function (d, t) {
+            var bh = d.createElement(t), s = d.getElementsByTagName(t)[0];
```
ayyazdaniaryan: You should get the last element and append it there. You'll be delaying the loading of `bundle.js` with this code. Doesn't make much difference in testing env, but just letting you know, as in the template, the bundle.js is the last `script` tag that is declared in the HTML.

assembl/views/__init__.py
```diff
@@ -333,6 +333,13 @@ def process_export_list(ls):
     theme_name, theme_relative_path = get_theme_info(discussion)
     node_env = os.getenv('NODE_ENV', 'production')
     under_test = bool(config.get('under_test') or False)
+    # Define flag for testing environment
+    wsginame = config.get('wsginame') or None
```
ayyazdaniaryan:
```
wsginame = config.get('wsginame', None)
inject_bugherd = wsginame in ('preprod.wsgi') or False
```

Just a bit cleaner, but it's all your preference.

assembl/templates/index_react.jinja2
ayyazdaniaryan: In the rc file, it could be like:
```
bugherd_url = https://www.bugherd.com/sidebarv2.js?apikey=%(bugherd_api)s
bugherd_api = mtcl06zwfwhfvmxqxb1d5q
```


And here, it would be

```
bh.src = bugherd_url.format(config)
```

assembl/views/__init__.py
```diff
@@ -333,6 +333,13 @@ def process_export_list(ls):
...
+    if wsginame:
+        # wsginame values are defined in *.rc files
+        inject_bugherd = True if wsginame in ('preprod.wsgi')
```
amensouissi: wsginame in ('preprod.wsgi')
2 cases:
1) substring in string ==> parentheses are not needed here
2) string in list/tuple of string ==> you need to add a comma here ('preprod.wsgi', ) or wsginame == 'preprod.wsgi'

PR #593, opened 2018-11-13
FIX - secretsmanager implementation

moutard-bluenove:
Only use AWS when a key aws_secret_id is specified. If not use the default in .rc file so the you can run in your desktop.

### DIFF
--- a/assembl/fabfile.py
+++ b/assembl/fabfile.py
@@ -160,15 +160,22 @@ def populate_secrets():
         # we don't have boto3 yet
         return
 
-    client = boto3.client('secretsmanager') # noqa
-    name = os.path.splitext(os.path.basename(env.rcfile))[0]
-    response = client.get_secret_value(
-        SecretId=name,
-    )
-    env_vars = json.loads(response["SecretString"])
-    # Add env variable overrides here
-    for env_key, env_value in env_vars.iteritems():
-        env[env_key] = env_value
+    # ignore if you don't have secrets it means you are on your desktop.
+    # so we don't need a AWS account to start locally
+    if env.get('aws_secrets_id', None):
+        aws_client = boto3.client('secretsmanager')
+        print("Fetching secrets from AWS Secrets Manager with id %s" % env.aws_secrets_id)
+        response = aws_client.get_secret_value(
+            SecretId=env.aws_secrets_id,
+        )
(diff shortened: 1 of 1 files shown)


assembl/fabfile.py
```diff
@@ -160,15 +160,20 @@ def populate_secrets():
...
-    response = client.get_secret_value(
-        SecretId=name,
-    )
-    env_vars = json.loads(response["SecretString"])
-    # Add env variable overrides here
-    for env_key, env_value in env_vars.iteritems():
-        env[env_key] = env_value
+    # ignore if you don't have secrets it means you are on your desktop.
+    # so we don't need a AWS account to start locally
+    if 'aws_secrets_id' in env and env.aws_secrets_id:
```
amensouissi: env.get('aws_secrets_id', None)

assembl/fabfile.py
```diff
@@ -160,15 +160,20 @@ def populate_secrets():
...
+        print("Fetching secrets from AWS Secrets Manager with id %s" % env.aws_secrets_id)
+        response = aws_client.get_secret_value(
+            SecretId=env.aws_secrets_id,
+        )
+        env_vars = json.loads(response["SecretString"])
+        # Add env variable overrides here
+        for env_key, env_value in env_vars.iteritems():
+            env[env_key] = env_value
+    else:
+        print("No aws_secrets_id defined - using defaults defined in .rc files.")
```
amensouissi: you should raise an error here only if it is necessary for server deployment (?)

assembl/fabfile.py
moutard-bluenove: done

assembl/fabfile.py
moutard-bluenove: how do we know we are on a server deployment ? I don't think relying on localhost is a good idea.

moutard-bluenove: Done. Fair enough, However I am a bit sceptic because we actually don't use wsgi on local but pserve instead, it feels really confusing to use this value.

ayyazdaniaryan: @moutard-bluenove Actually, you do. Waitress is also a WSGI server, just not uWSGI, which is what we use in production to serve python servers.
https://docs.pylonsproject.org/projects/waitress/en/latest/

PR #597, opened 2018-11-14
SUPPORT 30: the links to a post in the announcement do not scroll down to the post

MilanBarande:
This PR is linked to [this support ticket](https://bluenove.atlassian.net/browse/SUPPORT-30)

The issue was that the function that scrolls down to the post was only run in componentDidMount and we did not go through this lifecycle method of the Tree component when clicking on a link. It is now run in componentDidUpdate as well, which solves the issue.

### DIFF
--- a/assembl/static2/js/app/components/common/tree/tree.jsx
+++ b/assembl/static2/js/app/components/common/tree/tree.jsx
@@ -50,8 +50,11 @@ class Tree extends React.Component<Props> {
     }
   }
 
-  componentDidUpdate() {
-    this.triggerScrollToRow(this.props.initialRowIndex);
+  componentDidUpdate(prevProps: Props) {
+    const { initialRowIndex } = this.props;
+    if (prevProps.initialRowIndex !== initialRowIndex) {
+      this.triggerScrollToRow(initialRowIndex);
+    }
   }
 
   triggerScrollToRow = (rowIndex: ?number) => {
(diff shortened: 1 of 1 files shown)


assembl/static2/js/app/components/common/tree/tree.jsx
```diff
@@ -53,6 +50,16 @@ class Tree extends React.Component<Props> {
     }
   }
 
+  componentDidUpdate() {
+    this.triggerScrollToRow(this.props.initialRowIndex);
```
cedricmessiant: I would use `prevProps` (first argument) and compare to `this.props` to check if `initialRowIndex` changed before to scroll

MilanBarande: good point

assembl/static2/js/app/components/common/tree/tree.jsx
```diff
@@ -31,10 +31,7 @@ class Tree extends React.Component<Props> {
     // or to avoid recreating all dom nodes if we go back to the same idea.
     this.cache.clearAll();
     this.prevStopIndex = 0;
-    const { initialRowIndex } = this.props;
-    if (this.listRef && initialRowIndex !== null) {
-      this.listRef.scrollToRow(initialRowIndex);
-    }
+    this.triggerScrollToRow(this.props.initialRowIndex);
```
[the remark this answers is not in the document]
MilanBarande: yep, to keep it one line and simple

PR #607, opened 2018-11-20
Add username in csv extracts

paolina92:
- I added the username in all csv extracts

- I updated the corresponding tests

- I tried to write some tests on the voters extract but my skills are limited

### DIFF
--- a/assembl/models/auth.py
+++ b/assembl/models/auth.py
@@ -950,6 +950,7 @@ class User(AgentProfile):
         if self.username_p:
             size = len(self.username_p)
             return hash(self.username_p, size)
+        return None
 
     def check_password(self, password):
         if self.password:
--- a/assembl/models/auth.py
+++ b/assembl/models/auth.py
@@ -946,6 +946,11 @@ class User(AgentProfile):
             # set the new password
             self.password = hash_password(password)
 
+    def anonymous_username(self):
+        if self.username_p:
+            size = len(self.username_p)
+            return hash(self.username_p, size)
+
     def check_password(self, password):
         if self.password:
             from ..auth.password import verify_password
--- a/assembl/tests/views_tests/test_api2.py
+++ b/assembl/tests/views_tests/test_api2.py
@@ -1088,12 +1088,13 @@ class TestPhase1Export(AbstractExport):
     POST_LIKE_COUNT = 6
     POST_DISAGREE_COUNT = 7
     POST_CREATOR_NAME = 8
-    POST_CREATOR_EMAIL = 9
-    POST_CREATION_DATE = 10
-    SENTIMENT_ACTOR_NAME = 11
-    SENTIMENT_ACTOR_EMAIL = 12
-    SENTIMENT_CREATION_DATE = 13
-    POST_BODY_ORIGINAL = 14
+    POST_CREATOR_USERNAME = 9
+    POST_CREATOR_EMAIL = 10
+    POST_CREATION_DATE = 11
+    SENTIMENT_ACTOR_NAME = 12
+    SENTIMENT_ACTOR_EMAIL = 13
+    SENTIMENT_CREATION_DATE = 14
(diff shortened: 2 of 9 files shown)


assembl/models/auth.py
```diff
@@ -946,6 +946,11 @@ def password_p(self, password):
             # set the new password
             self.password = hash_password(password)
 
+    def anonymous_username(self):
```
ayyazdaniaryan: Not critical, but the pythonic way is to finish this method by adding `return None`. By adding this, the function is explicit that it will always return something.

assembl/tests/views_tests/test_api2.py
```diff
@@ -987,8 +987,10 @@ def test_add_timeline_event(test_app, discussion):
 class AbstractExport(object):
     # HEADER = {}
 
-    def _get(self, app, discussion_id, lang=None, view_name=None):
+    def _get(self, app, discussion_id, widget_id=0, lang=None, view_name=None, votes_export=False):
```
ayyazdaniaryan: You could simplify by just adding `widget_id` in the end instead of `votes_export` and check your condition against only one variable, instead of two.

assembl/tests/views_tests/test_api2.py
```diff
@@ -1180,4 +1186,23 @@ def test_base(self, proposals_with_sentiments, discussion, timeline_phase2_inter
...
+    def test_base(self, test_app, discussion, token_vote_spec_with_votes, gauge_vote_specification_with_votes, number_gauge_vote_specification_with_votes):
+        result = self.get_result(test_app, discussion.id, gauge_vote_specification_with_votes.widget_id, view_name=self.view_name, votes_export=True)
+        header = result[0]
+        # assert header[self.NOM_DU_CONTRIBUTEUR] == b"Nom Du Contributeur"
+        assert header[self.NOM_UTILISATEUR_CONTRIBUTEUR] == b"Nom D'Utilisateur Du Contributeur"
+        assert header[self.EMAIL_DU_CONTRIBUTEUR] == b"Adresse Mail Du Contributeur"
+        assert header[self.DATE_HEURE_DU_VOTE] == b"Date/Heure Du Vote"
+        assert header[self.PROPOSITION] == b"Proposition"
+
+    # TODO: Add more unit tests for votes export API.
```
ayyazdaniaryan: With the `token_vote_spec_with_votes` fixture, you can actually one more test that actually tests your code for the values of each column. That should be enough to test both the fact that export works and business logic is functional.

assembl/tests/views_tests/test_api2.py
```diff
@@ -1180,4 +1186,23 @@ def test_base(self, proposals_with_sentiments, discussion, timeline_phase2_inter
...
+    NOM_DU_CONTRIBUTEUR = 0
+    NOM_UTILISATEUR_CONTRIBUTEUR = 1
+    EMAIL_DU_CONTRIBUTEUR = 2
+    DATE_HEURE_DU_VOTE = 3
+    PROPOSITION = 4
+
+    def test_base(self, test_app, discussion, token_vote_spec_with_votes, gauge_vote_specification_with_votes, number_gauge_vote_specification_with_votes):
+        result = self.get_result(test_app, discussion.id, gauge_vote_specification_with_votes.widget_id, view_name=self.view_name, votes_export=True)
+        header = result[0]
+        # assert header[self.NOM_DU_CONTRIBUTEUR] == b"Nom Du Contributeur"
```
ayyazdaniaryan: Does this make the test fail? Perhaps uncomment this :)

paolina92: @ayyazdaniaryan  Yes... I should check this with you before the merge.

ayyazdaniaryan:
Yeah, it will affect bigdatext's clustering algo, we'll have to update that package as well. I'll take care of that. Otherwise, the changes look good. 👍

PR #610, opened 2018-11-21
Chinese flag support in Admin Flags

ayyazdaniaryan:


ayyazdaniaryan:
I saw it appear on preprod, and thought perhaps some of the changes made on the Chinese language weren't trickled down to this component.

screen shot 2018-11-21 at 1 37 15 pm

vincentfretin:
zh_Hant is not supported by v2, so the real fix is to select the right Chinese in v1 admin.

ayyazdaniaryan:
Yes, I saw that in the V1. But I don't think we should be limiting simply because it's a different type of Chinese.

ayyazdaniaryan:
Ahhhhh, I see that the support for Chinese on the platform is for simplified chinese only :(. The nav drop down doesn't support simplified chinese either. Elasticsearch supports zh_CN, which is closest to zh_Hans. Hmm, alright, so no support for other zh langs. Okay, let's discuss with team to see if we should be more guarded on other zh langs or not.

PR #613, opened 2018-11-22
Thematic of post in search

MilanBarande:
This branch is linked to [this ticket](https://bluenove.atlassian.net/browse/DEVAS-1723)

### DIFF
--- a/assembl/static2/js/app/components/debate/common/post/postView.jsx
+++ b/assembl/static2/js/app/components/debate/common/post/postView.jsx
@@ -175,6 +175,8 @@ class PostView extends React.PureComponent<Props, State> {
     const relatedIdeasTitles = indirectIdeaContentLinks
       ? indirectIdeaContentLinks.map(link => link && link.idea && link.idea.title)
       : [];
+    const hasRelatedIdeas = relatedIdeasTitles.length > 0;
+
     return (
       <div
         ref={(p) => {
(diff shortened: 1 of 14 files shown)


assembl/static2/js/app/components/debate/common/post/postView.jsx
```diff
@@ -229,15 +232,15 @@ class PostView extends React.PureComponent<Props, State> {
 
               <Attachments attachments={attachments} />
 
-              {!multiColumns && (
-                <div>
-                  <RelatedIdeas indirectIdeaContentLinks={indirectIdeaContentLinks} />
+              {!multiColumns ? (
+                <React.Fragment>
+                  <RelatedIdeas relatedIdeasTitles={relatedIdeasTitles} />
```
vincentfretin: You need a condition `hasRelatedIdeasTitles` here before calling RelatedIdeas.

cedricmessiant:
@MilanBarande @vincentfretin @amensouissi it seems that something is missing because this afternoon when we created new posts the idea titles were not updated but it seems strange, maybe it was another bug. Also, did you check with PO that we want it on extracts ? did you also check that this is OK that the idea title we see in search may not be the one that we see on the post (I think it is misleading / strange, maybe you should go up in the hierarchy to find the top thematic)

MilanBarande:
@cedricmessiant  So, PO wants the associated thematic on extracts too. And no problem with the fact that associated thematic on search and postview is not the same apparently. Guess you can justify by saying that you'd rather have top thematic on postview since you already know the closest thematic (you are on the page).

PR #621, opened 2018-11-27
fix the tags unque constraint

amensouissi:


assembl/alembic/versions/de9ade82771c_fix_tags_constraint.py
```diff
@@ -0,0 +1,29 @@
...
+
+def upgrade(pyramid_env):
+    with context.begin_transaction():
+        op.drop_constraint(u'keyword_value_key', 'keyword', type_='unique')
+        op.create_unique_constraint('unq_value_discussion', 'keyword', ['value', 'discussion_id'])
+
+
+def downgrade(pyramid_env):
+    with context.begin_transaction():
+        pass
```
cedricmessiant: don't we need the downgrade?

PR #622, opened 2018-11-27
Attachments in rich text fields in legal contents administration :sparkles: :tada:

cedricmessiant:
Only for legal contents for now.

If we want to handle attachments in other fields in react final form admin, we will need to :
- add the relations in the models with a distinct purpose per field
- handle the variable in the graphql mutations
- update the resolver in the graphql queries so it returns something
- add `withAttachmentButton` on the `<Field>`


assembl/graphql/idea.py
```diff
@@ -197,6 +200,10 @@ def resolve_title_entries(self, args, context, info):
     def resolve_body(self, args, context, info):
         return resolve_langstring(self.body, args.get('lang'))
 
+    def resolve_body_attachments(self, args, context, info):
+        # TODO:
```
pascalhuynh: is something need to be added here ?

assembl/graphql/resource.py
```diff
@@ -47,6 +48,10 @@ def resolve_text(self, args, context, info):
         text = resolve_langstring(self.text, args.get('lang'))
         return text
 
+    def resolve_text_attachments(self, args, context, info):
+        # TODO:
```
pascalhuynh: is something need to be added here ?

assembl/static2/tests/unit/components/form/multilingualRichTextFieldAdapter.spec.jsx
```diff
@@ -0,0 +1,170 @@
...
+        touched: false
+      }
+    };
+    const component = renderer.create(
+      <DummyForm>
+        <MultilingualRichTextFieldAdapter {...props} />
+      </DummyForm>
+    );
+    const tree = component.toJSON();
+    expect(tree).toMatchSnapshot();
```
pascalhuynh: Hey @cedricmessiant what's proving you that the rich text field is rendered here ? Saying that it simply matching a snapshot doesn't mean it appears. And also you have to manually check the snapshot created to make sure that the component is there. Same comment for the other `toMatchSnapshot` tests

assembl/graphql/idea.py
cedricmessiant: when we will add attachments on this field (I added the resolver and the TODO to make it easier)

assembl/static2/tests/unit/components/form/multilingualRichTextFieldAdapter.spec.jsx
cedricmessiant: I don't understand. I checked the snapshot when it has been created the first time and then everytime a test complains that it dosen't match the snapshot anymore, I figure out what happens to determine if something is wrong or if the snapshot needs to be updated.

assembl/static2/js/app/components/form/utils.js
```diff
@@ -71,11 +75,40 @@ export function convertToEntries(valuesByLocale: I18nValue): LangstringEntries {
...
+    const result = await uploadNewAttachments(valuesByLocale[locale], uploadDocument);
+    const { documentIds, contentState } = result;
+    return {
+      documentIds: documentIds,
+      localeCode: locale,
+      value: convertEditorStateToHTML(EditorState.createWithContent(contentState))
+    };
+  });
+
+  const variables = await Promise.all(results);
```
vincentfretin: I hope we don't have db conflicts when uploading multiple attachments with uwsgi.

assembl/static2/tests/helpers/graphql.js
```diff
@@ -1,17 +1,292 @@
 /* Helpers for tests involving graphql/apollo */
 import { ApolloClient } from 'react-apollo';
-import { addMockFunctionsToSchema } from 'graphql-tools';
-import { mockNetworkInterfaceWithSchema } from 'apollo-test-utils';
-import { buildClientSchema } from 'graphql';
+import { mockNetworkInterface } from 'react-apollo/test-utils';
 
-import * as introspectionResult from './schema.json';
```
vincentfretin: You can remove the file static2/tests/helpers/schema.json

assembl/static2/js/app/components/form/utils.js
cedricmessiant: just tested with multiple attachments in multiple fields in different languages and it seems ok but it's `pserve` and not `uwsgi`...

vincentfretin: pserve is using threads, uwsgi is used processes. Be sure to test it once deployed.

PR #623, opened 2018-11-28
Corrections thematic of post in search

MilanBarande:
This is a couple of fixes asked by the PO.

I guess the way I get the url to an idea could be improved. If you have a suggestion

### DIFF
--- a/assembl/static2/js/app/components/search.jsx
+++ b/assembl/static2/js/app/components/search.jsx
@@ -274,7 +274,7 @@ const PostHit = ({ bemBlocks, collapseSearch, locale, result }) => {
   const published = {};
   const ideaTitle = highlightedLSOrTruncatedLS(result, 'idea_title', locale);
   const postUrl = getUrl(result);
-  const ideaUrl = postUrl.replace(postUrl.slice(postUrl.indexOf('#'), postUrl.length), '');
+  const ideaUrl = postUrl.slice(0, postUrl.indexOf('#'));
   return (
     <BaseHit
       bemBlocks={bemBlocks}
@@ -324,7 +324,7 @@ const DumbExtractHit = ({ bemBlocks, collapseSearch, isHarvesting, locale, toggl
   };
   const ideaTitle = highlightedLSOrTruncatedLS(result, 'idea_title', locale);
   const extractUrl = getUrl(result);
-  const ideaUrl = extractUrl.replace(extractUrl.slice(extractUrl.indexOf('#'), extractUrl.length), '');
+  const ideaUrl = extractUrl.slice(0, extractUrl.indexOf('#'));
   return (
     <BaseHit
       bemBlocks={bemBlocks}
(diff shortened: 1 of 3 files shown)


assembl/static2/js/app/components/search.jsx
```diff
@@ -316,12 +323,14 @@ const DumbExtractHit = ({ bemBlocks, collapseSearch, isHarvesting, locale, toggl
     collapseSearch();
   };
   const ideaTitle = highlightedLSOrTruncatedLS(result, 'idea_title', locale);
+  const extractUrl = getUrl(result);
+  const ideaUrl = extractUrl.replace(extractUrl.slice(extractUrl.indexOf('#'), extractUrl.length), '');
```
vincentfretin: same here

assembl/static2/js/app/components/search.jsx
```diff
@@ -270,12 +273,14 @@ const PostHit = ({ bemBlocks, collapseSearch, locale, result }) => {
   const body = highlightedLSOrTruncatedLS(result, 'body', locale);
   const published = {};
   const ideaTitle = highlightedLSOrTruncatedLS(result, 'idea_title', locale);
+  const postUrl = getUrl(result);
+  const ideaUrl = postUrl.replace(postUrl.slice(postUrl.indexOf('#'), postUrl.length), '');
```
vincentfretin: `const ideaUrl = postUrl.slice(0, postUrl.indexOf('#'));`
is simpler no?

assembl/static2/tests/unit/components/__snapshots__/search.spec.jsx.snap
```diff
@@ -65,26 +65,6 @@ exports[`Hits components DumbExtractHit component should render an extract hit 1
...
-      <div
-        className="badges"
-      >
-        <span
-          className="badge"
-        >
-          Interesting thematic
-        </span>
-      </div>
-    </div>
```
vincentfretin: I know the snaphost is wrong here because of the way we import Link. Nothing we can do to fix it.

MilanBarande: Oooh that's why it removed those lines. I first thought it was normal and then by reviewing my code I figured I had wrongly updated the snapshots and was wondering what happend here but now it makes sense

assembl/static2/js/app/components/search.jsx
MilanBarande: 💯

assembl/static2/js/app/components/search.jsx
```diff
@@ -270,12 +273,14 @@ const PostHit = ({ bemBlocks, collapseSearch, locale, result }) => {
   const body = highlightedLSOrTruncatedLS(result, 'body', locale);
   const published = {};
   const ideaTitle = highlightedLSOrTruncatedLS(result, 'idea_title', locale);
+  const postUrl = getUrl(result);
+  const ideaUrl = postUrl.slice(0, postUrl.indexOf('#'));
```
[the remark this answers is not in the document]
vincentfretin: If there is two '#' in the url, indexOf return the index of the first one, so we're good.

PR #627, opened 2018-12-03
BM: add comment helper, bug fixes

pascalhuynh:
Feature added:
* add comment helper
* update some visual aspects
and couple bug fixes here and there


assembl/static2/js/app/components/debate/brightMirror/fictionBodyToolbar.jsx
```diff
@@ -59,7 +96,7 @@ const SentimentMenu = ({ postId, mySentiment, screenWidth, position, sentimentCo
                 return result;
               }, [])}
             </div>
-            <div className="txt" style={{ paddingLeft: `${(count + 2) * 6}px` }}>
+            <div className="txt" style={{ paddingLeft: `${(count + 4) * 6}px` }}>
```
MilanBarande: Don't you get huge padding if count is really high ? How about using count.toString().length ? Because I guess this is about the length of the number ?

assembl/static2/js/app/components/debate/brightMirror/sideComment/sideCommentMenu.jsx
```diff
@@ -28,6 +28,17 @@ type State = {
 };
 
 class SideCommentMenu extends React.Component<Props, State> {
+  static getDerivedStateFromProps(nextProps: Props) {
+    // Required when switching from displaying comment to submitting to force a refresh of component
+    const { displaySubmitBox } = nextProps;
+    if (displaySubmitBox) {
```
MilanBarande: `return displaySubmitBox ? { commentBoxDisplayed: false } : null; `
No ?

MilanBarande: Isn't there a problem if you return `null` as your new state btw?

assembl/static2/js/app/utils/globalFunctions.js
```diff
@@ -296,7 +296,5 @@ export function getIconPath(icon: string, color: string = '') {
   return color ? `${ICONS_PATH}/${color}/${icon}` : `${ICONS_PATH}/${icon}`;
 }
 
-export const getRandomPictureUrl = () => {
-  const pictureId = Math.floor(Math.random() * PICTURES_LENGTH) + 1;
-  return `${PICTURE_BASE_URL}${pictureId}${PICTURE_EXTENSION}`;
-};
\ No newline at end of file
+// We `pictureId + 1` because there is no image in the S3 bucket with 0 as an id
```
MilanBarande: Then why does the id starts at 0?

pascalhuynh: the picture id comes from the position of the fiction in an array: 
array[0] = first fiction => picture id should be 1 instead of 0 because no file on S3 with id 0
array[1] = second fiction and so on...

assembl/static2/js/app/components/debate/brightMirror/fictionBodyToolbar.jsx
pascalhuynh: If it gets messy we will fix it: shouldn't cause any problem though 👍

PR #628, opened 2018-12-03
Landing page objectives

ayyazdaniaryan:
- All new header administration using react-final-form
- new DatePicker Field Adapter for final-form
- Header now completely de-coupled from `extra_json`


assembl/static2/js/app/components/administration/landingPage/customizeHeader.jsx
```diff
@@ -1,137 +1,174 @@
...
+type State = {
+  startDate: DateTime,
+  endDate: DateTime,
+  startDateConflict: boolean,
+  endDateConflict: boolean
+};
+
+const loading = <Loader />;
+
+export const DumbCustomizeHeader = ({ client, editLocale, data: { discussion: { startDate, endDate } } }: Props) => {
```
pascalhuynh: @ayyazdaniaryan I don't think you need to pass the structure of data (`{ discussion: { startDate, endDate }`). Just `data` should be enough. Unless you get a Flow error...

pascalhuynh: But well, that way data is already destructured... so leave it that way.

assembl/static2/js/app/components/common/participateButton.jsx
```diff
@@ -8,14 +8,15 @@ import { getIsDebateStarted } from '../../utils/timeline';
 type Props = {
   displayPhase: Function,
   timeline: Timeline,
-  btnClass: string
+  btnClass: string,
+  btnLabel: string | null
```
pascalhuynh: @ayyazdaniaryan `btnLabel?: string`. It means optional value.

assembl/static2/js/app/components/form/datePickerFieldAdapter.jsx
```diff
@@ -0,0 +1,98 @@
...
+  { time: moment().utc() }
+
+  Errors are managed by a Mutator on the Form itself. As a result, to show errors
+  on DatePickerFieldAdapter, must add the `setFieldTouched` from final-form-set-field-touched
+  along with passing the form <FormApi> prop
+*/
+import DatePicker from 'react-datepicker';
+import 'react-datepicker/dist/react-datepicker.css';
+import * as React from 'react';
+import { type FieldRenderProps } from 'react-final-form';
```
pascalhuynh: @ayyazdaniaryan could you regroup every similar import altogether: component with component, type with type and so on. Thanks !

assembl/static2/js/app/components/form/datePickerFieldAdapter.jsx
```diff
@@ -0,0 +1,98 @@
...
+  required,
+  hasConflictingDate,
+  onDateChange,
+  children,
+  form,
+  ...rest
+}: Props) => {
+  const onLocalizedChange = (e: DateTime): void => {
+    if (e !== value.time) {
+      if (form) {
```
pascalhuynh: `if (e !== value.time && form) {`

assembl/static2/js/app/utils/globalFunctions.js
```diff
@@ -296,5 +298,13 @@ export function getIconPath(icon: string, color: string = '') {
   return color ? `${ICONS_PATH}/${color}/${icon}` : `${ICONS_PATH}/${icon}`;
 }
 
+
 // We `pictureId + 1` because there is no image in the S3 bucket with 0 as an id
-export const getPictureUrl = (pictureId: number) => `${PICTURE_BASE_URL}${pictureId + 1}${PICTURE_EXTENSION}`;
\ No newline at end of file
+export const getPictureUrl = (pictureId: number) => `${PICTURE_BASE_URL}${pictureId + 1}${PICTURE_EXTENSION}`;
+
+export const getRandomPictureUrl = () => {
```
pascalhuynh: This function should be deleted. I think it comes from a rebase. Only getPictureUrl should be kept.

assembl/static2/js/app/components/administration/landingPage/customizeHeader.jsx
```diff
@@ -1,137 +1,174 @@
...
+  const state: State = {
```
MilanBarande: Why having a functional component and then have a custom made state ? The whole state.startDate = e logic is confusing. I think the whole component would be much clearer if it was a class and using setState

assembl/static2/js/app/components/administration/landingPage/customizeHeader.jsx
MilanBarande: In the props we keep it to 1 level of destructuring, I would put it the way Pascal said and destructure data where it's needed

assembl/static2/js/app/components/administration/landingPage/customizeHeader.jsx
```diff
@@ -1,137 +1,174 @@
...
+    startDate: convertISO8601StringToDate(startDate),
+    endDate: convertISO8601StringToDate(endDate),
+    startDateConflict: false,
+    endDateConflict: false
+  };
+
+  const onStartChange = (e: DateTime): void => {
```
paolina92: You should check dates in `validate.js`

assembl/static2/js/app/components/administration/landingPage/header/validate.js
```diff
@@ -0,0 +1,27 @@
...
+import { type DateTime } from '../../../form/types.flow';
+import type { DatePickerValue, DatePickerValidation } from './types.flow';
+
+export const validStartDate = (startDate: DateTime | null, endDate: DateTime | null): boolean => {
+  if (!startDate || !endDate) { return true; }
+  return startDate <= endDate;
+};
+
+export const validEndDate = (startDate: DateTime | null, endDate: DateTime | null): boolean => {
+  if (!startDate || !endDate) { return true; }
```
MilanBarande: null is a valid date ?

assembl/static2/js/app/components/common/participateButton.jsx
MilanBarande: otherwise if it is not optional but can be null you can write ```btnLabel: ?string``` instead of ```| null```

assembl/static2/js/app/components/administration/landingPage/customizeHeader.jsx
```diff
@@ -1,137 +1,174 @@
 // @flow
```
paolina92: You should rename `customizeHeader` to `index` and put it in the `header` folder

assembl/static2/js/app/components/form/types.flow.js
```diff
@@ -12,4 +13,19 @@ export type FileVariable = null | 'TO_DELETE' | File;
...
 export type MutationsPromises = Array<() => Promise<*>>;
 
-export type SaveStatus = 'OK' | 'KO' | 'PENDING';
\ No newline at end of file
+export type SaveStatus = 'OK' | 'KO' | 'PENDING';
+
+export type DateTime = Moment;
+
+export type DatePickerInput = {
+  time: DateTime | null
```
MilanBarande: ?DateTime

assembl/static2/js/app/components/home/header.jsx
```diff
@@ -24,40 +27,51 @@ class Header extends React.Component {
...
+    } = this.props;
     return (
       <section className="home-section header-section">
         <Grid fluid className="max-container">
           <div className="header-content">
-            {debateData.headerLogoUrl ? <img className="header-logo" src={debateData.headerLogoUrl} alt="logo" /> : null}
+            {logoImage && logoImage.externalUrl ? <img className="header-logo" src={logoImage.externalUrl} alt="logo" /> : null}
             <div className="max-text-width">
-              {debateData.topic && <h1 className="light-title-1">{debateData.topic.titleEntries[locale]}</h1>}
+              {title && <h1 className="light-title-1">{title}</h1>}
```
MilanBarande: ```{title ? <h1 className="light-title-1">{title}</h1> : null}``` is now the preferred syntax

assembl/static2/js/app/pages/landingPageAdmin.jsx
```diff
@@ -198,9 +198,11 @@ class LandingPageAdmin extends React.Component<Props, State> {
   render() {
     const { editLocale, header, section, timelineModuleId } = this.props;
     const saveDisabled = !this.dataHaveChanged();
+    // TODO: Remove this crap after migrating all of landing page to react-final-form
```
MilanBarande: forgot this?

assembl/static2/js/app/utils/globalFunctions.js
```diff
@@ -1,8 +1,10 @@
 // @flow
+import moment from 'moment';
+import type Moment from 'moment';
 import { type Map } from 'immutable';
 
 import { getDisplayedPhaseIdentifier } from './timeline';
-import { HARVESTABLE_PHASES, ICONS_PATH, PICTURE_BASE_URL, PICTURE_EXTENSION } from '../constants';
```
MilanBarande: You can remove PICTURES_LENGTH

assembl/static2/js/app/utils/translations.js
```diff
@@ -903,7 +904,13 @@ const Translations = {
           headerDescription:
             "L'image doit avoir une hauteur de 450px de haut et une largeur de 1280px. Le poids ne doit pas dépasser 1Mo.",
           logoDescription:
-            "Le logo doit avoir une hauteur maximale de 78px et une largeur maximale de 200px. Le poids ne doit pas dépasser 1Mo. Le fond doit être transparent."
+            "Le logo doit avoir une hauteur maximale de 78px et une largeur maximale de 200px. Le poids ne doit pas dépasser 1Mo. Le fond doit être transparent.",
+          startDate: "Du",
+          endDate: "À",
```
MilanBarande: If we are talking about dates you would say "Du 12 Janvier **au** 15 Février" so you probably want "Au" here

assembl/static2/js/app/utils/translations.js
```diff
@@ -903,7 +904,13 @@ const Translations = {
...
+          timePlaceholder: "Optionnel : Entrez les dates du débat",
+          dateDescription: "Pour configurer les dates du debat. Si vous n'avez pas sélectionné des dates, les dates sont déduites des timelines.",
```
MilanBarande: sélectionné **de** dates

assembl/static2/js/app/components/administration/landingPage/customizeHeader.jsx
ayyazdaniaryan: Sure, I structured it at depth of one now.

assembl/static2/js/app/components/form/datePickerFieldAdapter.jsx
ayyazdaniaryan: Ooops, this logic was flawed. Instead of putting form up there, instead, I shifted the logic below under the same conditional check as the form

assembl/static2/js/app/utils/globalFunctions.js
ayyazdaniaryan: Indeed, must have crept in from a rebase. I'll remove it

assembl/static2/js/app/components/administration/landingPage/customizeHeader.jsx
ayyazdaniaryan: Done. Converted to a class-based component.

assembl/static2/js/app/components/administration/landingPage/customizeHeader.jsx
ayyazdaniaryan: In this form, two `Field`s' pristine state depend on one another. Hence, I have to create local state, and pass these onChange handlers which will compare and validate between the two fields. At the core of the callbacks, the "check logic", is imported from validate.js.

assembl/static2/js/app/components/administration/landingPage/customizeHeader.jsx
ayyazdaniaryan: Moved it to be more consistent with the other components.

assembl/static2/js/app/pages/landingPageAdmin.jsx
ayyazdaniaryan: Nope. The code below should be killed off when the rest of the landing page is converted to React Final Form

assembl/static2/js/app/utils/globalFunctions.js
ayyazdaniaryan: Bad rebase. Done.

assembl/static2/js/app/components/home/header.jsx
MilanBarande: No, we probably should but they are a few on which people disagree. I think this one is not a stylistic choice but has a reason in terms of code logic

assembl/static2/js/app/components/common/manageErrorAndLoading.jsx
```diff
@@ -29,6 +32,10 @@ const manageErrorAndLoading = (props: Props) => (WrappedComponent: React.Compone
 
   if (loading || (data && data.loading)) {
     if (props.displayLoader) {
+      if (props.customLoader) {
+        return props.customLoaderProps ? <customLoader {...props.customLoaderProps} /> : <customLoader />;
```
vincentfretin: customLoader is for me undefined here. It should have been `<props.customLoader`
I think your `DumbHeaderPlaceholder` is actually never rendered, so this is equivalent to `displayLoader: false`

assembl/static2/js/app/components/form/datePickerFieldAdapter.jsx
```diff
@@ -0,0 +1,97 @@
...
+
+type Props = {
+  editLocale: string,
+  picker?: DatePickerType,
+  placeHolder: string,
+  showTime: boolean,
+  hasConflictingDate: boolean,
+  input: {
+    name: string,
+    onChange: (SyntheticInputEvent<*> | any) => void,
```
vincentfretin: `SyntheticInputEvent<*> | any` is the same as `any`. Could you have avoided the `|any` here?

assembl/static2/js/app/components/common/manageErrorAndLoading.jsx
vincentfretin: Ah you actually used `displayLoader: false` with `customLoader` in `Header` component. So this code is actually never reached and is completely broken. I removed the code in https://github.com/assembl/assembl/commit/ea52a8025a38de465202e2e2e6c4b1d3479124ed

assembl/static2/js/app/components/form/datePickerFieldAdapter.jsx
vincentfretin: I changed the flow annotation in https://github.com/assembl/assembl/commit/bc037b77900c91216796bd0bbb171a601970e8df

assembl/alembic/versions/d813ab110e00_discussion_start_end_dates.py
```diff
@@ -0,0 +1,78 @@
...
+            if not success:
+                # Try to use phases
+                print("Migrating Discussion %s's dates to model from phases" % slug)
+                dates = db.query(sa.sql.func.min(m.TimelineEvent.start),
+                                 sa.sql.func.max(m.TimelineEvent.end)
+                                ).filter_by(discussion_id=d_id).first()
+                if dates:
+                    (start_date, end_date) = dates
+                    if (start_date and end_date and start_date < end_date):
+                        db.execute(d.update().where(d.c.id == d_id).values(
```
vincentfretin: `d` may be undefined here. I moved the `d = m.Discussion.__table__` line 
in https://github.com/assembl/assembl/commit/e1f73c2f5c9f0f37bd679a336565fb5b8dd0fbaa

assembl/static2/js/app/components/form/utils.js
```diff
@@ -49,6 +57,25 @@ export function convertEntriesToI18nRichText(entries: RichTextLangstringEntries)
   return convertEntriesToI18nValue(convertEntriesToEditorState(entries));
 }
 
+export function convertISO8601StringToDateTime(_entry: string): DatePickerInput {
+  if (_entry) {
+    const t = moment(_entry, moment.ISO_8601).utc();
+    if (t._isValid) { // eslint-disable-line no-underscore-dangle
```
vincentfretin: Why do you access a private api `t._isValid`? I modified it in https://github.com/assembl/assembl/commit/e259dbb57f83c7d030fd7f1d7ab0bf349fbde75f to use `t.isValid()`

PR #636, opened 2018-12-21
Review timeline ordering structures

maparent:
The linked list structure (previous_event) was a bit fragile in reordering operations; I had fixed that with post_updates earlier, but it required acrobatics around deletes. Not healthy. I repaired with passive_deletes, which seems to work reliably.
I noted while there that there's a new field order, which is not kept in sync with the linked list. Why? Were you aware of the linked list field?
I also noted that creating a new timeline item with the old interface fails because the JS wants to send an int in the json, which the float field refuses. That's a separate fix, which may have wide-ranging consequences. I could also be more precise here.


vincentfretin:
We don't use the old interface and don't use the linked chain. We currently use the new interface with the order field. We should remove the previous_event field.

PR #641, opened 2019-01-17
Legal contents validation modal Part 2

MilanBarande:
Here is [the associated Jira Ticket](https://bluenove.atlassian.net/browse/DEVAS-1998)

The point here is to add a checkbox in the legal contents administration to activate or deactivate the modal that forces users to accept legal contents when they logged int through an SSO

TODO:

- [x] Styling


assembl/graphql/docstrings.py
```diff
@@ -173,6 +172,7 @@ class UpdateLegalContents:
     cookies_policy_attachments = "The list of the Attachments used in cookies policy entries."
     privacy_policy_attachments = "The list of the Attachments used in privacy policy entries."
     user_guidelines_attachments = "The list of the Attachments used in user guidelines entries."
+    mandatory_legal_contents_validation = """A boolean flag to activate or not mandatory validation of legal contents after SSO login."""
```
pascalhuynh: you can remove the `or not` :)

assembl/graphql/docstrings.py
```diff
@@ -144,6 +143,7 @@ class LegalContents:
     cookies_policy_attachments = Default.string_entry % ("Attachments for cookies policy.")
     privacy_policy_attachments = Default.string_entry % ("Attachments for privacy policy.")
     user_guidelines_attachments = Default.string_entry % ("Attachments for user guidelines.")
+    mandatory_legal_contents_validation = """A boolean flag to activate or not mandatory validation of legal contents after SSO login."""
```
pascalhuynh: you can remove the `or not` here also :)

MilanBarande: True!

PR #648, opened 2019-02-04
bugherd-46 | delete the display flex that break the gauge style

felixsaintgenis:


vincentfretin:
ok, you somehow reverted https://github.com/assembl/assembl/commit/8195a62fef49f3bc8518f3b4c2ad6e964c5ffb78

PR #649, opened 2019-02-04
bugherd-48 | delete useless title input in the multicolumn post modification

felixsaintgenis:


assembl/static2/js/app/components/debate/common/editPostForm.jsx
```diff
@@ -43,7 +43,8 @@ type EditPostFormProps = {
   bodyMaxLength?: number,
   draftable?: boolean,
   draftSuccessMsgId?: string,
-  isDebateModerated: boolean
+  isDebateModerated: boolean,
+  multiColumns: boolean
```
vincentfretin: The EditPostForm is used as several places. It should be `multiColumns?: boolean` and add a default prop `multiColumns: false`

felixsaintgenis: Sure

PR #653, opened 2019-02-05
refacto the footer links and make the cookie page always here

felixsaintgenis:


assembl/static2/js/app/components/common/footer.jsx
```diff
@@ -30,11 +29,17 @@ const Footer = ({
...
   hasTermsAndConditions,
-  hasCookiesPolicy,
   hasPrivacyPolicy,
   hasUserGuidelines,
   lang
 }: Props) => {
+  const legalPages = {
+    terms: hasTermsAndConditions,
+    legalNotice: hasLegalNotice,
+    cookiesPolicy: true,
```
MilanBarande: Why is this one always true ? We have a hasCookiesPolicy boolean in the tabscondition query

assembl/static2/js/app/components/common/footer.jsx
```diff
@@ -76,46 +81,14 @@ const Footer = ({
...
-              )}
-              {hasUserGuidelines && (
-                <div className="user-guidelines">
-                  {hasPrivacyPolicy || hasCookiesPolicy ? <span className="small-hyphen-padding"> &mdash; </span> : null}
-                  <Link to={`${get('userGuidelines', debateSlug)}`}>
-                    <Translate value="footer.userGuidelines" />
-                  </Link>
-                </div>
+            <div className="legal-pages">
+              {Object.keys(legalPages).map(
```
MilanBarande: 👏 great! Just for a cleaner JSX I'd have this in a renderFooterLinks function and simply do {renderFooterLinks()} here

felixsaintgenis: the cookie page needs to be always here because of the cookie configuration section

assembl/static2/js/app/components/common/footer.jsx
felixsaintgenis: the cookie page needs to be always here because of the cookie configuration section

felixsaintgenis:
Je me suis mis une note pour tester le footer sur IE 11 quand on fera un déploiement vu qu'il y a des doutes avec le flex wrap

PR #660, opened 2019-02-08
Fix bug window size react

felixsaintgenis:
hotfix for the assise debate

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -55,8 +55,11 @@ cache:
 services:
  - redis-server
 before_install:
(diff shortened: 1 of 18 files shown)


mohamedzayed92:
This change has already been merged into develop. This branch was taken out of master, using cherry-picks to apply the commits in order to fix the window size bug. Thus it is not to be merged in develop.

vincentfretin:
@mohamedzayed92 Even if the commits are already on develop, you need to merge this hotfix branch on develop so that master is always behing develop. You need to hard reset master before your merge and redo the hotfix by following our procedure.

PR #663, opened 2019-02-13
remove the max length on richtext forms

MilanBarande:


assembl/static2/tests/unit/components/debate/thread/__snapshots__/answerForm.spec.jsx.snap
```diff
@@ -102,7 +102,7 @@ exports[`AnswerForm component should render a answer form 1`] = `
             }
           }
           handleInputFocus={[Function]}
-          maxLength={3000}
+          maxLength={0}
```
ayyazdaniaryan: `NO_BODY_LENGTH` instead. Same thing, but let's use the constants we've created.

assembl/static2/js/app/components/debate/thread/answerForm.jsx
```diff
@@ -154,7 +153,6 @@ export class DumbAnswerForm extends React.PureComponent<Props, State> {
                 <RichTextEditor
                   editorState={this.state.body}
                   handleInputFocus={this.handleInputFocus}
-                  maxLength={BODY_MAX_LENGTH}
```
ayyazdaniaryan: Is this required for Bright Mirror and/or admin views as well? If so, then you should remove the logic of `maxLength` in RichTextEditor class.

ayyazdaniaryan: Not sure if it is part of the scope of the bug or not

MilanBarande: I think it's just the thread

MilanBarande: Update: it's everywhere :--) will remove

assembl/static2/tests/unit/components/debate/thread/__snapshots__/answerForm.spec.jsx.snap
MilanBarande: mmh I didn't notice there was a defaultProps to 0, weird choice. Anyway, this is going

PR #672, opened 2019-02-18
Bugherd 125 126

felixsaintgenis:


assembl/static2/css/components/posts.scss
```diff
@@ -313,8 +315,15 @@ $deleted-post-border: 1px solid $light-grey !default;
...
     .assembl-icon-check-box-declined {
-      font-size: 18px;
-      padding-left: 3px;
+      font-size: 20px;
+      padding-left: 2px;
+    }
+
+    @-moz-document url-prefix() {
+      /* adapt padding to be align on firefox */
+      .assembl-icon-check-box-declined {
```
felixsaintgenis: I'm going to reimport the svg with a better size. It will resolve the problem. Cleaner than specify a mozilla rule for 1px in padding

PR #673, opened 2019-02-18
S3 attachments

ayyazdaniaryan:
Putting greater control over where the attachments of Assembl are hosted. Could be local hashfs (like before), or an s3 bucket.

### DIFF
--- a/assembl/lib/attachment_service.py
+++ b/assembl/lib/attachment_service.py
@@ -68,7 +68,6 @@ class AmazonAttachmentService(AttachmentService):
         key = self.computeHash(filename)
         if not self.exists(key):
             self.bucket.upload_file(filename, key, {
-                'ACL': 'authenticated-read',
                 'ContentType': mimetype or 'application/binary'
             })
         return key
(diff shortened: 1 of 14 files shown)


assembl/lib/attachment_service.py
```diff
@@ -0,0 +1,109 @@
...
+        self.s3 = boto3.resource('s3', region)
+        self.s3c = self.s3.meta.client
+        self.bucket = self.s3.Bucket(self.bucket_name)
+
+    def put_file(self, dataf, mimetype=None):
+        key = self.computeHash(dataf)
+        if not self.exists(key):
+            if hasattr(dataf, 'read'):
+                self.bucket.upload_fileobj(dataf, key, {
+                    'ContentType': mimetype or 'application/binary'
```
vincentfretin: Don't you want "application/octet-stream" as a more standard default value instead of "application/binary"?

vincentfretin: https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Complete_list_of_MIME_types

PR #676, opened 2019-02-21
Bugherd 137 fix opengraph functions

mohamedzayed92:


assembl/views/__init__.py
```diff
@@ -577,7 +580,10 @@ def get_topic(request):
         else:
             topic_dict = topic_dict["titleEntries"]
             locale = discussion.preferences["preferred_locales"][0]
-            return adapt_to_html_content(topic_dict.get(opengraph_locale, topic_dict[locale]))
+            if locale in topic_dict:
+                return adapt_to_html_content(topic_dict.get(opengraph_locale, topic_dict[locale]))
+            else:
+                return "No topic available in the extra json."
```
ayyazdaniaryan: No no, try to pass whatever the original language's description is. And if there is None, you return empty string. Apply to the other scenario as well.

assembl/views/__init__.py
```diff
@@ -559,8 +563,12 @@ def get_description(request):
             locale = discussion.preferences['preferred_locales'][0]
             if locale in objectives_dict:
                 return adapt_to_html_content(objectives_dict.get(opengraph_locale, objectives_dict[locale]))
+            elif len(objectives_dict) != 0:
```
ayyazdaniaryan: Hm, okay, you could traverse the the list of `preferred_locales` and see which key is in the `objectives_dict` before going to pick the first key from the dictionary. Extend this logic to the other function as well 👍

PR #677, opened 2019-02-21
Optimize vote session query

vincentfretin:
- [x] Bottleneck is on `voteResults {numParticipants} `, optimize the numParticipants sql query
- [x] We retrieve 400 participants on each proposal, to calculate vote session numParticipants on the frontend. => calculate vote session numParticipants on backend, don't retrieve participants
- [ ] create a script to populate idea_vote table with more than 10 000 votes (around 30 proposals, 3 vote specs, 400 participants)
- [ ] maybe create a dedicated index to not do a `Seq Scan` (full table scan) https://use-the-index-luke.com/sql/clustering/index-only-scan-covering-index
- [ ] try to do a single get_voter_ids_query() for the vote session with group by idea_id instead of doing the query for each proposal
- [ ] create a seperate VoteSessionResults query to get the results, VoteSession and VoteSessionResults will be launched at the same time, but VoteSession will finish earlier to show something on the page. And we don't care about results in the administation.
- [ ] maybe create a separate VoteSessionAdmin query, for the participant view, we don't care about all langstring entries and modules at the vote session level


vincentfretin:
With 17545 entries in the idea_vote table, some tombstoned, 137 users, 20 proposals.
Before: 28
After: 2s

so for the sql part, I think it's ok now.
I will look into creating a separate VoteSessionResults query though.

vincentfretin:
Actually the split VoteSession/VoteSessionResults is not worth it, it my tests doing a separate VoteSessionResults query we wil probably reduce the VoteSession query by 500-800ms, but to really do it I need to a complex merging of the two queries, I may introduce bugs. Once the vote session is closed, we need the results right away, so actually this is silly to do a separate query.

vincentfretin:
For the idea of writing a single sql query with group by `vote_class.idea_id`, I thought about it, it's actually no really feasible because of the filter `vote_class.vote_spec_id.in_`...
Ok I guess I will merge the PR now.

vincentfretin:
FYI, it's not doing a Seq Scan anymore, but an Index Scan:
```
[(u'Unique  (cost=9.32..9.33 rows=1 width=4)',),
 (u'  ->  Sort  (cost=9.32..9.33 rows=1 width=4)',),
 (u'        Sort Key: voter_id',),
 (u'        ->  Index Scan using ix_public_idea_vote_vote_spec_id on idea_vote  (cost=0.29..9.31 rows=1 width=4)',),
 (u'              Index Cond: (vote_spec_id = 49)',),
 (u'              Filter: (tombstone_date IS NULL)',)]
```

vincentfretin:
I added to this PR the vote session counters participants and participations on IdeaPreview component and giant menu.

PR #682, opened 2019-02-26
Fix BM side comment permission 

cedrtang:
To test, need to add the correct permission to participants


assembl/graphql/post.py
```diff
@@ -764,11 +764,13 @@ class Input:
...
     @abort_transaction_on_exception
     def mutate(root, args, context, info):
-        require_cls_permission(CrudPermissions.CREATE, models.Extract, context)
         discussion_id = context.matchdict['discussion_id']
-
         user_id = context.authenticated_userid or Everyone
 
+        permissions = get_permissions(user_id, discussion_id)
+        if P_ADD_SIDE_COMMENT not in permissions and require_cls_permission(CrudPermissions.CREATE, models.Extract, context):
+            raise HTTPUnauthorized()
```
ayyazdaniaryan: We try to use the graphql `permission_helpers` as much as possible. For a start, there is a more descriptive error that is translated across multiple languages that is passed through. Also, you can use the `require_instance_permission` I suspect too (double check that).

cedrtang: @ayyazdaniaryan `require_instance_permission` is used for existing instances obviously so can only be used for updating/deleting, not creating. `permissions_helpers` doesn't handle single permissions, only class and instances. I could add a new function for this.

ayyazdaniaryan: You're right. Nah, this is simple enough to check. If the use-case grows over the APIs, we'll make it a function.

assembl/models/preferences.py
```diff
@@ -807,12 +809,14 @@ def get_preference_data_list(self):
...
                     P_SEND_SYNTHESIS,
                     P_VOTE,
+                    P_ADD_SIDE_COMMENT,
                 ],
                 R_PARTICIPANT: [
                     P_ADD_POST,
                     P_DELETE_MY_POST,
                     P_EDIT_MY_POST,
                     P_VOTE,
+                    P_ADD_SIDE_COMMENT,
```
ayyazdaniaryan: Cool, since the base Participant will have this perm, I suggest the the `R_CATCHER` to also have this perm as well.

PR #687, opened 2019-03-26
WIP Invoke

ayyazdaniaryan:
Migration to AWS infra is built on a series of branches, which are topped by invoke. All changes should be visible here. This is a rolling branch, so whilst AWS production machines are currently on this branch, we would like to move away from this and back to develop. Things that are done:

- [x] initial implementation of CICD in fabric to dev-assembl
- [x] break themes into buckets in order to namespace them (done) - Check https://github.com/bluenove/assembl2-client-themes/pull/2, the accompanying changes
- [x] Migrate assembl's fonts to S3 as well 
- [x] Instead of one connection pool for DB, there is now a read_only pool, and a write pool of connection sessions. 
- [x] Use boto3 to connect to RDS (AWS database) via IAM roles instead of db_user/db_password, and refresh the session token on demand
- [x] Rewrite check_availability scripts in python instead of bash
- [x] Add logging to cloudwatch
- [x] Migrate core fabric functionality to Invoke, use these scripts in CICD + deployment
- [x] Create a standardized wheelname and logic to push the wheel and wheelhouse to an S3 bucket (used by CICD server only)
- [x] Lock database creation and elasticsearch indexing to the startup of a single machine in a cluster 
- [ ] Correct logic of configuration updates on AWS
- [x] Add nginx to cloudwatch
- [ ] Update and install celery script from latest assembl wheel
- [ ] Documentation
- [ ] Remove dead code in fabric related to first implementation of CICD on OVH infrastructure


.gitlab-ci.yml
```diff
@@ -30,65 +31,134 @@ test_frontend:
...
+    GIT_STRATEGY: clone
   artifacts:
     paths:
       - dist/
   before_script:
-    - sh scripts/install_build_dependencies.sh
-    - sh scripts/install_node_and_yarn.sh
-    - sh scripts/configure_assembl_remote.sh
-    - eval $(ssh-agent -s) && sh scripts/configure_ssh.sh
-    - sh scripts/clear_apt_cache.sh
```
vincentfretin: In the scripts folder, install_build_dependencies.sh install_node_and_yarn.sh configure_assembl_remote.sh configure_ssh.sh clear_apt_cache.sh create_assembl_wheel.sh are unused it seems. There were replaced by inv commands I guess. Please delete those scripts if we're not using them anymore. I see bump_assembl_version_and_push.sh has been deleted, but not the others scripts.

assembl/tasks/common.py
```diff
@@ -0,0 +1,321 @@
...
+    pre = list(kwargs.pop('pre', args))
+    pre.append(base_task(setup_ctx))
+    return base_task(pre=pre, **kwargs)
+
+
+def exists(c, path, sudo=False):
+    """
+    Return True if given path exists on the current remote host. Taken from patchwork.
+    """
+    cmd = '{}test -e "$(echo {})"'.format('sudo ' if sudo else '', path)
```
MathieuVDP: Why not use built-in python utils to check if file exists (such as os.path.exists) ?

assembl/__version__.py
```diff
@@ -0,0 +1,30 @@
...
+                # We're on the tag
+                _cached_version = Version(parts[0])
+            else:
+                base = Version(parts[0]).next_patch()
+                _cached_version = Version('%s.dev%s+%s' % (base, parts[1], parts[2][1:]))
+        else:
+            # from wheel
+            _cached_version = Version(pkg_resources.get_distribution("assembl").version)
+            assert str(_cached_version) == __version__
+    return _cached_version
```
vincentfretin: Is this `version()` function used somewhere? I didn't find any import related to `version()` elsewhere and we have a similar `auto_inc_version` function in `setup.py` The `git describe --tags` in `auto_inc_version` doesn't work on travis probably because we are not in the right folder. In the `version()` implementation here you `cd` into the folder first.

assembl/__init__.py
```diff
@@ -36,6 +35,7 @@
 
 def main(global_config, **settings):
     """ Return a Pyramid WSGI application. """
+    settings = dict(global_config, **settings)
     settings['config_uri'] = global_config['__file__']
```
MathieuVDP: No need to set config_uri key as it is already set in previous line with all other keys from global_config.

assembl/fabfile.py
```diff
@@ -507,19 +551,27 @@ def migrate_local_ini(backup=False):
...
+        retries -= 1
     # Another supervisor, upstart, etc may be watching it, give it more time
-    sleep(5)
+    sleep(6)
+
+
+@task
+def supervisor_restart():
+    """Restart supervisor itself."""
+    execute(supervisor_shutdown)
```
vincentfretin: We tested `supervisorctl status` before to only run `supervisord` if it was not up. But now we shudown supervisor, so for me this "no such file" check is not needed anymore.

assembl/fabfile.py
```diff
@@ -817,6 +871,10 @@ def install_url_metadata_source():
         print cyan("Cloning git repository")
         with cd("%(projectpath)s/.." % env):
             run('git clone git://github.com/assembl/url_metadata.git')
+    else:
+        print cyan("Url Metadata service being updated...")
+        with cd("%(projectpath)s/.." % env):
```
vincentfretin: It's wrong, it should be `with cd("%(projectpath)s/../url_metadata" % env):`

assembl/fabfile.py
```diff
@@ -876,6 +934,59 @@ def is_db_updated():
...
+            if exists(get_upload_dir()):
+                venvcmd('aws s3 cp %s' % cmd)
+            else:
+                print(red("There is no uploads folder to sync with Amazon S3!"))
+
+
+@task
+def sync_with_remote_upload_folder():
+    """
+    Syncronize the uploads folder on the host with S3 bucket
```
vincentfretin: Syncronize => Synchronize

assembl/fabfile.py
```diff
@@ -876,6 +934,59 @@ def is_db_updated():
...
+    """
+    _sync_uploads_folder()
+
+
+@task
+def sync_with_local_upload_folder():
+    """
+    Syncronize a remote uploads folder, on S3 bucket, with the local uploads folder
```
vincentfretin: Syncronize => Synchronize

assembl/fabfile.py
```diff
@@ -876,6 +934,59 @@ def is_db_updated():
...
+    """
+    _sync_uploads_folder(local=True)
+
+
+@task
+def reset_upload_folder():
+    if env.wsginame == 'staging.wsginame':
```
vincentfretin: I'm pretty sure it has to be "staging.wsgi". We have "staging.wsgi" in other places.

assembl/fabfile.py
```diff
@@ -876,6 +934,59 @@ def is_db_updated():
...
+        # Ensure an uploads folder
+        remote_uploads_exists = False
+        local_upload_exists = exists(get_upload_dir())
+        with shell_env(
+            AWS_ACCESS_KEY_ID=env.aws_access_key_id,
+            AWS_SECRET_ACCESS_KEY=env.aws_secret_access_key
+        ):
+            remote_uploads_exists = venvcmd('aws s3 ls s3://{}/uploads | wc -l'.format(env.aws_bucket_name))
+            remote_uploads_exists = False if exists == '0' else True
```
vincentfretin: I think you wanted `remote_uploads_exists == '0'` here.

assembl/fabfile.py
```diff
@@ -965,18 +1078,19 @@ def database_restore_aws(backup=False):
...
-    if not backup:
-        assert(env.wsginame in ('staging.wsgi', 'dev.wsgi'))
-        processes = filter_autostart_processes(_processes_to_restart_without_backup)
-    else:
-        processes = filter_autostart_processes(_processes_to_restart_with_backup)
+    # if not backup:
+    #     assert(env.wsginame in ('staging.wsgi', 'dev.wsgi'))
+    #     processes = filter_autostart_processes(_processes_to_restart_without_backup)
+    # else:
+    #     processes = filter_autostart_processes(_processes_to_restart_with_backup)
```
vincentfretin: remove the commented code if you really want to delete it

assembl/fabfile.py
```diff
@@ -1472,19 +1610,114 @@ def generate_dh_group():
...
+        return content
+    except s3.exceptions.NoSuchKey:
+        return None
+
+
+@task
+def get_aws_localrc():
+    assert running_locally()
+    import requests
+    r = requests.get('http://169.254.169.254/latest/meta-data/iam/info')
```
vincentfretin: hard coded ip?

assembl/fabfile.py
```diff
@@ -1472,19 +1610,114 @@ def generate_dh_group():
...
+        match = extends_re.search(ex_content)
+    if not content:
+        raise RuntimeError("local_%s.rc was not defined in S3" % account)
+
+
+@task
+def setup_aws_default_region():
+    assert running_locally()
+    import requests
+    r = requests.get('http://169.254.169.254/latest/meta-data/placement/availability-zone')
```
vincentfretin: Same ip as above. may add a constant?

assembl/fabfile.py
```diff
@@ -1472,19 +1610,114 @@ def generate_dh_group():
...
+    region = r.content[:-1]
+    run('aws configure set region ' + region)
+
+
+@task
+def aws_instance_startup():
+    """Operations to startup a fresh aws instance from an assembl AMI"""
+    if not env.projectpath:
+        # Assumption: Started from project directory
+        env.projectpath = os.cwd()
```
vincentfretin: `os.cwd()` doesn't exist! It should be `os.getcwd()`

assembl/fabfile.py
```diff
@@ -1472,19 +1610,114 @@ def generate_dh_group():
...
+
+
+@task
+def aws_server_startup_from_local():
+    """Update files that depend on local.rc and restart nginx, supervisor"""
+    create_local_ini()
+    venvcmd('assembl-ini-files populate %s' % (env.ini_file))
+    fill_template('assembl/templates/system/nginx_default.jinja2', env, 'var/share/assembl.nginx')
+    if is_supervisord_running():
+        supervisor_restart()
```
vincentfretin: Shouldn't you use `execute(supervisor_restart)` ?

assembl/fabfile.py
```diff
@@ -1528,7 +1765,8 @@ def webservers_start():
     if env.uses_nginx:
         # Nginx
         if exists('/etc/init.d/nginx'):
-            sudo('/etc/init.d/nginx start')
+            # Have to ensure that the env.user has visudo rights to call this
+            run('/etc/init.d/nginx start')
```
vincentfretin: even if the user has visudo right, you should prefix the command by `sudo` right? Like it's done for the reload.

assembl/auth/util.py
```diff
@@ -494,12 +505,32 @@ def add_user(name, email, password, role, force=False, username=None,
...
+    status_in_discussion = None
+    if resend_if_not_logged_in and discussion and not (created_user or created_localrole):
+        status_in_discussion = user.get_status_in_discussion(discussion.id)
+    if send_password_change and (
+            created_user or created_localrole or (
+                resend_if_not_logged_in and (
+                    status_in_discussion is None or
+                    not status_in_discussion.first_visit))):
+        from assembl.views.auth.views import send_change_password_email
+        from assembl.models import Discussion
```
MathieuVDP: Why Discussion is needed here ?

assembl/fabfile.py
```diff
@@ -1031,15 +1040,12 @@ def database_dump_aws():
 
 
 _processes_to_restart_without_backup = [
-    "dev:pserve", "celery", "changes_router",
-    "celery_notify", "source_reader"]
+    "dev:pserve" "celery", "changes_router",
```
MathieuVDP: Missing a comma between "dev:pserve" and "celery" ?

assembl/tasks/deploy.py
```diff
@@ -0,0 +1,583 @@
...
+
+
+@task()
+def supervisor_restart(c):
+    """Restart supervisor itself."""
+    supervisor_shutdown(c)
+    with venv(c, True):
+        result = c.run("supervisorctl status")
+        if "no such file" in result.stdout:
+            c.run("supervisord")
```
vincentfretin: same comment as in fabfile.py
if you shutdown supervisor, you are pretty sure that it is not started I think, so the `supervisorctl status` is for me not needed.

assembl/fabfile.py
```diff
@@ -3296,45 +3304,56 @@ def secure_sshd_fail2ban():
...
-        run('mkdir -p %s' % private_config_path)
-
+    if exists(path):
         # Ensure that the server's random.ini survives (under migration). New servers will generate own random.ini
-        remote_random_path_destination = normpath(os.path.join(private_config_path, 'random.ini'))
+        basename = os.path.basename(path)
+        remote_random_path_destination = normpath(to, basename)
         if not exists(remote_random_path_destination):
-            run('cp %s %s' % (remote_random_path, remote_random_path_destination))
+            run('cp -r %s %s' % (path, remote_random_path_destination))
```
MathieuVDP: If, by any chance, the random path already exists, this command won't fail and you have no clue that the copy command wasn't executed.

assembl/tasks/sudoer.py
```diff
@@ -0,0 +1,307 @@
...
+    c.run('pip install -r requirements-build.txt')
+
+
+@task()
+def install_node_and_yarn(c):
+    """Compiles the necessary packages in order for transpilers to build static asset code. It will often
+    be used in the CI/CD context.
+
+    This is different than updating node inside of a virtual environment, and nodeenv is not used. However,
+    the node version's MUST be mactching.
```
vincentfretin: mactching => matching

assembl/tasks/sudoer.py
```diff
@@ -0,0 +1,307 @@
...
+    c.run('curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | sudo apt-key add -')
+    c.run('echo "deb https://dl.yarnpkg.com/debian/ stable main" | sudo tee /etc/apt/sources.list.d/yarn.list')
+    c.run('wget -qO- https://deb.nodesource.com/setup_%s.x | sudo bash -' % node_major_version)
+    c.sudo('apt-get install -yqq nodejs yarn')
+
+
+@task()
+def install_chrome_dependencies(c):
+    """
+    Installs the requirements for chrome + pupeteer to be run. Assumes to be run on Debian based machines
```
vincentfretin: should be infinitive, Install and Assume. see https://www.python.org/dev/peps/pep-0257/
pupeteer => puppeteer

assembl/fabfile.py
```diff
@@ -3296,45 +3304,56 @@ def secure_sshd_fail2ban():
...
-    remote_random_path = os.path.join(env.projectpath, 'random.ini')
-    private_config_path = normpath(os.path. join(env.projectpath, '..', 'global_configs'))
-    if exists(remote_random_path):
-        run('mkdir -p %s' % private_config_path)
-
+    if exists(path):
         # Ensure that the server's random.ini survives (under migration). New servers will generate own random.ini
-        remote_random_path_destination = normpath(os.path.join(private_config_path, 'random.ini'))
+        basename = os.path.basename(path)
+        remote_random_path_destination = normpath(to, basename)
```
MathieuVDP: variable should be renamed as it doesn't refer to random.ini anymore, and above comment should also be removed.

assembl/views/__init__.py
```diff
@@ -140,20 +216,33 @@ def get_bundle_css_hash(href):
...
+            regex = r'/build/themes/default/theme_' + re.escape(theme_name) + r'_web\.(.*)\.js$'
+        else:
+            regex = r'/build/themes/' + re.escape(bucket_name) + '/[^/]+/theme_' + re.escape(theme_name) + r'_web\.(.*)\.js$'
+        return get_resource_hash(regex, href)
 
     soup = BeautifulSoup(source)
     bundle = soup.find(src=get_bundle_hash)
     bundle_css = soup.find(href=get_bundle_css_hash)
     theme_css = soup.find(href=get_theme_css_hash)
+    theme_js = soup.find(src=get_theme_js_hash)
```
vincentfretin: You don't have theme_js in the jinja template in production. This is only for development.

assembl/fabfile.py
```diff
@@ -1363,7 +1370,9 @@ def get_robot_machine():
 
 def is_integration_env():
     # Centralize checking whether in CI/CD env
-    if getenv('TRAVIS_COMMIT', None):
+    # Travis (https://docs.travis-ci.com/user/environment-variables/)
+    # Gitlab (https://docs.gitlab.com/ee/ci/variables/)
+    if getenv('CI', None):
         return True
```
MathieuVDP: Can be refactor to `return getenv('CI', False)`

assembl/lib/read_write_session.py
```diff
@@ -0,0 +1,34 @@
...
+class ReadWriteSession(orm.Session):
+    """A session that can divert read queries to a different engine
+    Inspired by https://gist.github.com/adhorn/b84dc47175259992d406
+    """
+
+    def __init__(self, bind=None, autoflush=False,
+                 read_bind=None, reading=False, **options):
+        self.read_bind = read_bind
+        self.reading = reading
+        orm.Session.__init__(
```
MathieuVDP: Use super() instead of orm.Session

assembl/lib/read_write_session.py
```diff
@@ -12,23 +13,31 @@ class ReadWriteSession(orm.Session):
...
-            self.reading = reading
+            self.readonly = readonly
+
 
-# TODO: reentrant context manager.
+@contextmanager
+def readonly(session):
+    was_readonly = session.readonly
+    try:
+        session.set_readonly(True)
```
MathieuVDP: Default value is True, so can be refactored into `session.set_readonly()`

assembl/lib/read_write_session.py
```diff
@@ -15,6 +15,7 @@ class ReadWriteSession(orm.Session):
     def __init__(self, bind=None, autoflush=False,
                  read_bind=None, readonly=False, **options):
         self.read_bind = read_bind
+        read_bind.is_readonly = True
```
MathieuVDP: Should be **self**.real_bind.is_readonly = True ?

assembl/lib/read_write_session.py
vincentfretin: `super().__init__` is for python 3. For python 2, you will need super(ReadWriteSession, self).__init__

assembl/lib/rds_token_url.py
```diff
@@ -0,0 +1,70 @@
...
+    def __init__(
+            self, drivername, region=None, username=None, host=None,
+            session_name=None, port=None, database=None, query=None):
+        super(RdsTokenUrl, self).__init__(
+            drivername, username=username, host=host, port=port or 5432,
+            database=database, query=query)
+        self.aws_region = region
+
+    def get_rds_client(self):
+        if not getattr(self, 'rds_client', None):
```
MathieuVDP: Could be False as default value, since we are testing the existence of the attribute

assembl/lib/rds_token_url.py
```diff
@@ -0,0 +1,70 @@
...
+            port=port, database=database, query=query)
+        self.iam_role = iam_role
+        self.sts_session_name = session_name or 'assembl'
+        self.sts_client = boto3.client('sts', self.aws_region)
+
+    def _needs_sts_role(self):
+        role = getattr(self, 'sts_role', None)
+        if role:
+            expiry = role['Credentials']['Expiration']
+            return (datetime.now() - expiry) < timedelta(minutes=10)
```
MathieuVDP: Should `return False`, in case role is not found as an attribute (Explicit is better than implicit)

assembl/alembic/env.py
```diff
@@ -62,15 +62,16 @@ def run_migrations_offline():
     script output.
 
     """
-    url = aws_connection_url(config.get_section('app:assembl')
+    url = connection_url(config.get_section('app:assembl')
         ) or config.get_main_option("sqlalchemy.url")
```
vincentfretin: Is this `or` still valid? This is the only place where you do this. Everywhere else in the code you are doing:
```
url = connection_url(settings)
if url:
    settings['sqlalchemy.url'] = url
```

assembl/tasks/mac.py
```diff
@@ -0,0 +1,106 @@
+import os
```
mohamedzayed92: Creating this Mac file was a choice taken by Raphael to separate the production file from developer file. Its purpose was to avoid the `if (running_locally())` and if env.mac in all functions. It is not complete. Are we still going to keep it ?

assembl/tasks/sudoer.py
```diff
@@ -0,0 +1,307 @@
...
+    is_integration_env,
+    delete_foreign_tasks,
+    get_venv_site_packages,
+    exists as sudo_exists)
+
+# try:
+#     # invoke 0.11
+#     from invoke import ctask as task
+# except ImportError:
+#     from invoke import task
```
mohamedzayed92: is this still necessary ?

assembl/tasks/sudoer.py
```diff
@@ -0,0 +1,307 @@
...
+
+python_dependencies = [
+    'python-dev',
+    'python-pip',
+    'build-essential'
+]
+
+
+@task
+def install_python(c):
```
mohamedzayed92: Is this task necessary, I thought I saw you install python manually with apt-get in your scripts

assembl/templates/system/assembl.service.jinja2
```diff
@@ -0,0 +1,27 @@
+# Sample supervisord service file
```
mohamedzayed92: why is this file a Jinja2 template, there is nothing to be changed in it. The values are all set

tasks/assembl.py
```diff
@@ -0,0 +1,79 @@
...
+    project_prefix = getattr(c.config, '_project_home', c.config._project_prefix[:-1])
+    assembl_prefix = project_prefix + '/assembl/configs/'
+    current = c.config._project
+    target = current.get('_extends', None)
+    while target:
+        if isabs(target):
+            if exists(target):
+                data = c.config._load_yaml(assembl_prefix + target)
+            else:
+                raise RuntimeError("Cannot find " + target)
```
MathieuVDP: Could raise FileNotFoundError, which is more specific than RuntimeError.

assembl/templates/system/system_user_information.jinja2
```diff
@@ -0,0 +1,5 @@
+hostname = {{ hostname }}
```
mohamedzayed92: This template is only used in the fabfile in a function called `create_user`. This function used to create a user and take care of all the .ssh configuration. Since you handled that in the scripts, I don't think we need that function anymore and hence we don't need this template

assembl/templates/system/urlmetadata.service.jinja2
```diff
@@ -0,0 +1,22 @@
+# Sample service file for direct urlmetadata startup
```
mohamedzayed92: Does this file need to be a template ? All values are default.

assembl/templates/system/uwsgicloudwatch.service.jinja2
```diff
@@ -0,0 +1,25 @@
...
+
+[Service]
+# These are the lines you have to modify for your environment. This assumes Assembl is running under the assembl_user user.
+Environment=VENV_PATH=/home/assembl_user/venvpy3
+User=assembl_user
+Group=www-data
+
+Type=simple
+Restart=on-failure
+ExecStart=/bin/bash -c "/home/assembl_user/venvpy3/bin/uwsgi-cloudwatch http://localhost:3002 --namespace assembl/uwsgi --frequency 60 --region `curl http://169.254.169.254/latest/meta-data/placement/availability-zone|sed -E 's/(.*).$/\1/'`"
```
mohamedzayed92: What is this IP dress ? Shouldn't that be templated ?

ci_cd_scripts/check_availability.sh
```diff
@@ -1,56 +0,0 @@
-#!/bin/sh
-
-# This script try to connect to elastic search and access the DB
-# return 0 if success / return 1 if error
-
-ASSEMBL_ROOT=$VIRTUAL_ENV/..
-LOCAL_INI=$ASSEMBL_ROOT/local.ini
-
-elasticsearch_host="$(grep "elasticsearch_host" $LOCAL_INI | tr -d ' ' | awk -F "=" '{print $2}')"
```
mohamedzayed92: Not sure it is relevant to this PR but I will mention it anyway. Often I had to do a reindex-all-contents for elastic search on production servers. Since we are not supposed to have SSH access to production instances in the future, we won't be able to do that. just raising a flag @vincentfretin

tasks/assembl.py
```diff
@@ -70,9 +91,204 @@ def create_var_dir(c):
...
+        if not is_supervisord_running(c):
+            return
+        c.run("supervisorctl shutdown")
+        for i in range(10):
+            sleep(6)
+            result = c.run("supervisorctl status", warn=True)
+            if not result.failed:
+                break
+    # Another supervisor, upstart, etc may be watching it, give it more time
+    sleep(6)
```
MathieuVDP: Here, you unsure if supervisor is down or not. Maybe raise an exception if after all retries it is still up ? Or kill the process ?

doc/sample_systemd_script/assembl.service
```diff
@@ -15,8 +15,8 @@ Group=www-data
 
 Type=simple
 Restart=on-failure
-ExecStart=/bin/bash -c '. $ASSEMBL_PATH/venv/bin/activate; exec supervisord -c $ASSEMBL_PATH/supervisord.conf --nodaemon'
-ExecStartPost=/bin/bash -c 'sleep 5; . $ASSEMBL_PATH/venv/bin/activate; supervisorctl -c $ASSEMBL_PATH/supervisord.conf start prod:uwsgi >/tmp/supervisordebug.txt'
+ExecStart=/bin/bash -c 'test -f $ASSEMBL_PATH/local.ini && source $ASSEMBL_PATH/venv/bin/activate && exec supervisord -c $ASSEMBL_PATH/supervisord.conf --nodaemon'
```
mohamedzayed92: This file has a template, shouldn't that file be removed before the merge ?

mohamedzayed92: Actually I think it should be removed , since all the templates are now in templates/system

doc/sample_systemd_script/urlmetadata.service
```diff
@@ -0,0 +1,22 @@
+# Sample service file for direct urlmetadata startup
```
mohamedzayed92: I think this file also should be templated and put in templates/system

ci_cd_scripts/check_availability.sh
vincentfretin: Please read all the PR. The reindexing is now handled if the mapping changed.

assembl/configs/aws.yaml
```diff
@@ -15,7 +15,7 @@ attachment_bucket: "assembl-data-%(aws_client)s"
 elasticsearch_index: "%(aws_client)s:assembl-es"
 elasticsearch_port: 80
 elasticsearch_version: 5.6.8
-aws_secrets_id: assembl_secrets
+aws_secrets_ids: arn:aws:secretsmanager:eu-west-1:900208097337:secret:assembl_secrets-1eqvUR
```
MathieuVDP: Is this real secret ?

assembl/tasks/common.py
ayyazdaniaryan: There were a few reasons. But I think the only one of the reasons that it remains is the sudo aspect, allowing `sudo` checking if a file exists. `os.path.exists` does not have those kinds of rights.

assembl/fabfile.py
ayyazdaniaryan: I'll make this a constant. The IP address is fixed from AWS. https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html

assembl/fabfile.py
ayyazdaniaryan: Lol, good catch. This code was never tested, and in fact the logic was recreated by invoke.

assembl/fabfile.py
ayyazdaniaryan: As I understood it, this was the first pass at migrating to cloud using fabric. The idea was to use fabric ON the machine itself. It was meant to only be run locally (on the machine). In any case, part of the dead code...

assembl/fabfile.py
ayyazdaniaryan: Well, from running reload/stop/start, it seems like it doesn't make much difference. If the user has visudo rights, they can call the command. I've made them all use sudo so it's consistent.

assembl/auth/util.py
ayyazdaniaryan: Lol, bad copy/pasta. It was already imported at the top of the function and used. Hence the linter not complaining about it.

assembl/fabfile.py
ayyazdaniaryan: Already resolved. You were likely looking commit by commit?

assembl/lib/sqla.py
```diff
@@ -1819,10 +1853,31 @@ def configure_engine(settings, zope_tr=True, autoflush=True, session_maker=None,
...
+    if read_url:
+        settings = dict(settings)
+        settings['sqlalchemy.url'] = read_url
+        read_engine = engine_from_config(settings, 'sqlalchemy.', **engine_kwargs)
+    else:
+        read_engine = None
+    session_maker.configure(bind=engine, read_bind=read_engine)
     global db_schema, _metadata, Base, TimestampedBase, ObsoleteBase, TimestampedObsolete
-    if settings['sqlalchemy.url'].startswith('virtuoso:'):
+    if str(settings['sqlalchemy.url']).startswith('virtuoso:'):
```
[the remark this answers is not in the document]
ayyazdaniaryan: Done.

assembl/static2/css/variables.scss
```diff
@@ -66,8 +66,7 @@ $margin-section-xxs: 70px;
 $margin-section-xxxs: 60px;
 $margin-section-tiny: 30px;
 
-$fonts-dir: '/static2/fonts';
-$img-dir: '/static2/img';
+$fonts-dir: '//s3-eu-west-1.amazonaws.com/bluenove-assembl-fonts';
```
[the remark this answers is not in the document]
ayyazdaniaryan: Yes, I should update the docs for fonts as well so that it's clear for front-ends.

assembl/scripts/lambda_create_db_aws_user.py
```diff
@@ -0,0 +1,38 @@
...
+def lambda_handler(input, ctx):
+    sm = boto3.client('secretsmanager')
+    response = sm.get_secret_value(SecretId='db_password_assembl')
+    if 'SecretString' in response:
+        info = response['SecretString']
+    else:
+        info = base64.b64decode(response['SecretBinary'])
+    info = json.loads(info)
+    database = input.get('database', os.environ.get('database', 'assembl'))
+    cnx = pg8000.core.Connection(
```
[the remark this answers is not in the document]
ayyazdaniaryan: This script is not loaded in `assembl`. The script is in fact managed by AWS Lambda. pg8000 is used because it's a dependency-less DBAPI implementation in python, which can be used easily in Lambda.

assembl/static2/webpack.config.js
```diff
@@ -13,12 +13,20 @@ var _ = require('lodash');
 
 function theme_entries() {
     var entries = {},
-        paths = glob.sync('./css/themes/**/*_web.scss'),
-        i, path, parts, name;
+        paths = glob.sync('./css/themes/**/**/*_web.scss'),
```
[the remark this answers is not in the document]
ayyazdaniaryan: Right. Reverting back.

assembl/tasks/sudoer.py
ayyazdaniaryan: Cool. Done.

assembl/views/__init__.py
ayyazdaniaryan: Yeah, it's right. Copy/pasta that I didn't remove. But just to be clear, the webpack-dev-server only serves the js file as is, with no hash, right?

assembl/lib/rds_token_url.py
ayyazdaniaryan: You could, won't change anything. But you're looking for a key in an object, basically. If the key isn't there, you either throw an exception, or return `None` to indicate nothing was found. It's falsy enough for the condition clause it's used in.

assembl/tasks/build.py
```diff
@@ -0,0 +1,714 @@
...
+
+@task(install_chrome_dependencies)
+def prepare_integration_tests(c):
+    """
+    Prepares the environment for running Assembl's integration tests. Fully assumes to be run in CI/CD env
+    """
+    c.run('git clone https://github.com/bluenove/assembl-tests')
+    yarn = c.run('which yarn')
+    with c.cd('assembl-tests'):
+        c.run('{}'.format(yarn))
```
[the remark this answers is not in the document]
ayyazdaniaryan: To ensure that the production boxes are as up to date with this dependency as front-end developers are in their daily workflow

assembl/tasks/build.py
```diff
@@ -0,0 +1,714 @@
...
+    if is_integration_env(c):
+        # This has to be added to actually checkout a branch because gitlab runs its runner
+        # in detached HEAD state. This is mostly fine, however, building the wheel names requires
+        # access to the branch's history. This is a workaround instead of setting hard depths on
+        # shallow cloning.
+        c.run('git checkout -B "$CI_BUILD_REF_NAME" "$CI_BUILD_REF"')
+
+
+@task()
+def start_deploy_on_client(c, client_id, region=None):
```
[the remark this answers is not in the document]
ayyazdaniaryan: Hmm, it's a case of naming, really. The `deploy` module is meant to scripts that are run **on** the production machine itself. This function is used from a CICD machine to push latest code to a deployment group on CodeDeploy. Could change the name to `push_state_to_client` or something...

assembl/tasks/build.py
```diff
@@ -0,0 +1,714 @@
...
+        print status
+
+
+@task()
+def get_deployment_clients(c):
+    """
+    Fetches the list of accounts available to deploy to from a remote bucket
+    Assume CI/CD Environment at all times
+    """
+    get_s3_file('bluenove-assembl-deployments', 'clients.json', 'clients.json', c.config.aws_shared_region)
```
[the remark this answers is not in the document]
ayyazdaniaryan: Yeah, it's a bit harder to understand without context. This code will be executed on a CICD machine. This machine needs to know the account id of every account it needs to deploy in the final part of the deployment job. The clients.json is is pushed on the bucket mentioned here, and we're fetching that file in the CICD machine prior to doing deployment.

assembl/views/__init__.py
vincentfretin: Yes, that's right.

assembl/tasks/common.py
```diff
@@ -0,0 +1,321 @@
...
+        pass
+
+
+def get_secrets_from_manager(c, aws_secret_ids, cache=True):
+    base = {}
+    if aws_secret_ids:
+        import boto3
+        secret_clients = {}
+
+        def get_secret_client(region=None):
```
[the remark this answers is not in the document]
ayyazdaniaryan: Because it's mildly abstracting our the logic in the outer function. The intent is this function to not be used outside of the for loop of the secret ids.

assembl/tasks/common.py
```diff
@@ -0,0 +1,321 @@
...
+    print venv
+    assert exists(c, venv)
+    activate = c.config.get('_internal', {}).get('activate', 'activate')
+    if cd:
+        return nested(c.cd(project_prefix), c.prefix('source %s/bin/%s' % (venv, activate)))
+    else:
+        return c.prefix('source %s/bin/%s' % (venv, activate))
+
+
+def venv_py3(c):
```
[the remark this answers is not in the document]
ayyazdaniaryan: In `assembl.tasks.deploy:update_url_metadata` it's used.

assembl/tasks/common.py
```diff
@@ -0,0 +1,321 @@
...
+        platform_info = f.read()
+        if 'amazon' not in platform_info:
+            return None
+    import requests
+    try:
+        r = requests.get('http://169.254.169.254/latest/dynamic/instance-identity/document', timeout=2)
+        if r.ok:
+            return r.json()['accountId']
+    except:
+        pass
```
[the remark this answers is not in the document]
ayyazdaniaryan: Changed to `None` to be more consistent

assembl/tasks/mac.py
ayyazdaniaryan: I think so, yeah. There will be a duplication of logic, but with the right abstraction, can be used in both places equally well. We'll keep it.

assembl/tasks/sudoer.py
ayyazdaniaryan: Nope, ctask, which is basically task with `contextualized=True` (means passing the context as first param) became the default in Invoke 0.13 onwards. We're on Invoke 1.2
https://github.com/pyinvoke/invoke/issues/114

assembl/tasks/sudoer.py
ayyazdaniaryan: No, original intent was to use this to create our AMIs. I think, for centralizing, it should still be as part of our longer term goal. Here is the line that this inv command would replace: 
https://gitlab.com/assembl/terraform/blob/dev/Infrastructure/Packer/BluenoveGolden/scripts/golden_bootstrap.sh#L7

assembl/templates/system/assembl.service.jinja2
ayyazdaniaryan: They're jinja2 files because for the moment they are static, but in near future, they will be variable. Trying to keep things consistent in case changes are required.

assembl/templates/system/system_user_information.jinja2
ayyazdaniaryan: I'll be removing these. They're outdated code from the `add_deploy_job` branch, which was the first go at CICD. The assumptions there are no longer relevant.

assembl/templates/system/uwsgicloudwatch.service.jinja2
ayyazdaniaryan: It's the IP address that Amazon provides for an EC2 box to get metadata information about itself.

assembl/configs/aws.yaml
ayyazdaniaryan: No, these are keys to fetch from SecretManager on AWS. No secrets shared on the filesystem

doc/sample_systemd_script/urlmetadata.service
ayyazdaniaryan: I think this file can be deleted, in fact. I'll go ahead and do that. Refer to here: 
https://gitlab.com/assembl/terraform/blob/dev/Infrastructure/Packer/UrlMetadata/scripts/urlmetadata_ubuntu_bootstrap.sh

PR #688, opened 2019-03-26
be sure get the basename for a document uploaded on Windows

vincentfretin:


assembl/graphql/document.py
```diff
@@ -49,7 +58,12 @@ def mutate(root, args, context, info):
 
         uploaded_file = args.get('file')
         if uploaded_file is not None:
+            # Because the server is on GNU/Linux, basename will only work
+            # with path using "/".
             filename = os.path.basename(context.POST[uploaded_file].filename)
+            # we need to remove "\" used by Windows too.
+            filename = filename.split('\\')[-1]
```
ayyazdaniaryan: Easier solution:
```
import ntpath
filename = ntpath.basename(filename)
```

https://docs.python.org/2/library/os.path.html
You're forcing the windows based paths to be selected. I just tested it quickly on the following two paths:

`C:\\file.txt`, `/home/user/aryan/file.txt`
And the result of both was `file.txt`. You could do a swap. I did see an issue with this example `C:\\somepath\file.txt`, it returned `somepath\x0cile.txt`, but obviously `C:\\somepath\\file.txt` returned `file.txt` as expected. YMMV

ayyazdaniaryan: Could propagate this to the query resolver as well

vincentfretin: Cool.
You need to use a raw string when you test it, because `\` has a special meaning in Python.
```
>>> ntpath.basename(r"C:\\somepath\file.txt")
'file.txt'
```

PR #689, opened 2019-03-28
fix travis on invoke branc

vincentfretin:


setup.py
```diff
@@ -35,7 +35,10 @@ def parse_reqs(req_files, links=False):
 
 
 def auto_inc_version():
-    tag = check_output('git describe --tags', shell=True).strip()
+    # On travis, the repo is cloned with --depth=100, there may not be enough
+    # commits to include any tags, the option --always will fallback to
+    # the latest commit hash
+    tag = check_output('git describe --tags --always', shell=True).strip()
```
ayyazdaniaryan: Hmmm, it seems quite harmless, and it's good to fallback on a hash which makes Travis continue. But, I see a problem downstream. The version name now will be inconsistent. Two options:

1. This is Travis, who cares, just make the test pass and that's it, and under gitlab/dev you won't hit the case of the git clone depth.
2. We keep the code and version consistent with production, even in travis, and set clone depth to `false`, and do a git fetch in the `pre-install` to get the full code state

Which do you prefer (no judgement)?

Info on clone depth (https://docs.travis-ci.com/user/customizing-the-build/#git-clone-depth)

vincentfretin: We actually have `depth: 100` in `.travis.yml`. We can increase it or disable it, but it will increase the time to get the full repo. With depth=100, it takes 27s, with depth false, it takes 1m23s. I prefer option 1.

ayyazdaniaryan: Alright, we'll be paying this cost in GitLab when CICD is fully up, you realize that, right? For Travis, I don't mind option 1. So, I'm okay with option 1, knowing full well that we cannot use git clone depth on GitLab (because we'll be generating a wheel, requiring the tag name)

ayyazdaniaryan: Perhaps a comment, and then you're good to merge to invoke

PR #695, opened 2019-04-17
refacto the addIframeForMindManager function

felixsaintgenis:


assembl/static2/js/app/utils/linkify.js
```diff
@@ -25,8 +25,10 @@ export function addProtocol(url: string): string {
   return hasProtocol ? url : `https://${url}`;
 }
 
-export function addIframeForMindManager(html: string): string {
-  const url = /(<a href="(https:\/\/share.mindmanager.com[^\s]+)".*<\/a>)/gi;
+export function addIframeForMindMapping(html: string): string {
+  const url = html.includes('mindmanager')
```
ayyazdaniaryan: This is a good first effort. But it's exclusively binary. What if tomorrow there's another mapping tool they want to embed. You'll have to change this logic completely. I suggest you build an array of regex's, then do a reduce on which mapping fits the given html. And finally, apply the mapping the works.

vincentfretin:
Your code seems really complicated. The following code should work the same:
```
function addIframeForMindManager(html: string): string {
  const url = /(<a href="(https:\/\/(share.mindmanager.com|embed.coggle.it)[^\s]+)".*?<\/a>)/gi;  // ? means the smallest string needed before a </a>
  return html.replace(url, '<div class="iframed"><iframe src="$2"></iframe></div>');
}

const html='<p>bla bla <a href="https://share.mindmanager.com/123">mindmap</a> and another <a href="https://embed.coggle.it/456">mindmap</a></p>'
addIframeForMindManager(html)
"<p>bla bla <div class=\"iframed\"><iframe src=\"https://share.mindmanager.com/123\"></iframe></div> and another <a href=\"https://embed.coggle.it/456\">mindmap</a></p>"
```

ayyazdaniaryan:
Arrf, I tested a non-greedy way, but it didn't work :(. This seems to work, however. Will update for the next deployment.

PR #701, opened 2019-05-13
Bugherd fixes

felixsaintgenis:
Every commit is one bugherd ticket.

### DIFF
--- a/assembl/static2/css/components/posts.scss
+++ b/assembl/static2/css/components/posts.scss
@@ -82,7 +82,12 @@
     }
 
     .actions {
+      display: inline-flex;
       margin: 15px;
+    }
+
+    .assembl-icon-check-box-declined:before {
+      font-size: 20px;
     }
   }
 
--- a/assembl/static2/css/components/video.scss
+++ b/assembl/static2/css/components/video.scss
@@ -1,8 +1,3 @@
-.video-wrapper {
-  width: 900px;
-  margin: auto;
-}
-
 .video-container {
   height: 0;
   margin: 20px 0 0;
@@ -14,10 +9,11 @@
   iframe,
   object,
   embed {
-    height: 100%;
-    left: 0;
     position: absolute;
     top: 0;
+    left: 0;
     width: 100%;
+    height: 100%;
+    border: 0;
   }
 }
--- a/assembl/static2/css/components/video.scss
+++ b/assembl/static2/css/components/video.scss
@@ -1,3 +1,8 @@
+.video-wrapper {
+  width: 900px;
+  margin: auto;
+}
+
 .video-container {
   height: 0;
   margin: 20px 0 0;
--- a/assembl/static2/js/app/components/common/urlPreview/youtubeEmbed.jsx
+++ b/assembl/static2/js/app/components/common/urlPreview/youtubeEmbed.jsx
@@ -6,18 +6,16 @@ type Props = {
 };
 
 const YoutubeVideo = ({ id }: Props) => (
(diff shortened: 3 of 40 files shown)


assembl/static2/css/components/video.scss
```diff
@@ -1,3 +1,8 @@
+.video-wrapper {
```
vincentfretin: Is it responsive? Did you try on mobile?

assembl/static2/js/app/pages/question.jsx
```diff
@@ -81,7 +81,7 @@ export function DumbQuestion(props: Props) {
               </div>
               <div className="center">
                 <h3 className="collapsed-title">
-                  <span>{`${params.questionIndex}/ ${title}`}</span>
+                  {numPosts > 1 ? <span>{`${params.questionIndex}/ ${title}`}</span> : <span>{`${title}`}</span>}
```
vincentfretin: No, revert this. The issue you need to fix here it to remove the question index if there is only one question (not numPosts), but we will need to a do a query of the thematic to know how many questions we have. We avoid it until now, we added the questionIndex in the url to avoid the query, but here we will need to do it.

assembl/static2/js/app/components/common/urlPreview/youtubeEmbed.jsx
```diff
@@ -13,7 +13,7 @@ const YoutubeVideo = ({ id }: Props) => (
       type="text/html"
       width="640"
       height="360"
-      src={`https://www.youtube.com/embed/${id}?autoplay=1`}
+      src={`https://www.youtube.com/embed/${id}?autoplay=0`}
```
vincentfretin: bugherd 196

assembl/static2/js/app/components/debate/survey/post.jsx
```diff
@@ -262,6 +262,7 @@ class Post extends React.Component<Props> {
     let userCanDeleteThisMessage = false;
     const userIsModerator = connectedUserIsModerator();
     const isPendingForModerator = isPending && userIsModerator;
+    const deleteModalMessage = isPendingForModerator ? 'debate.confirmRefusalBody' : null;
```
vincentfretin: bugherd 222

assembl/static2/css/components/posts.scss
```diff
@@ -82,8 +82,13 @@
     }
 
     .actions {
+      display: inline-flex;
       margin: 15px;
     }
+
+    .assembl-icon-check-box-declined:before {
+      font-size: 20px;
+    }
```
vincentfretin: bugherd 221

assembl/static2/js/app/components/debate/survey/question.jsx
```diff
@@ -130,7 +130,7 @@ export class Question extends React.Component<Props, State> {
               <RichTextEditor
                 editorState={this.state.postBody}
                 onChange={this.updateBody}
-                placeHolder={I18n.t('debate.survey.txtAreaPh')}
+                placeholder={I18n.t('debate.survey.txtAreaPh')}
```
vincentfretin: Good catch. It's silly we have a different case placeHolder for the datepicker. We should lowercase the prop for datepicker field too.

assembl/static2/css/components/video.scss
felixsaintgenis: You're right, I changed the coden ow I'm using the bootstrap classes embed-responsive embed-responsive-16by9

PR #705, opened 2019-05-14
Invoke setup ctx

MathieuVDP:
I DON'T KNOW IF IT BREAKS ANYTHING, PLEASE REVIEW
Fix the issue with updating account_data.yaml in AWS and not getting the right configuration with invoke. The issue came from the fact that setup_ctx doesn't update the configuration if a file changed between the start of the invoke process and the actual actions. I needed to make "invoke.yaml" the target to ensure correct inheritance after the first call to setup_ctx.

### DIFF
--- a/assembl/tasks/common.py
+++ b/assembl/tasks/common.py
@@ -15,6 +15,7 @@ local_code_root = dirname(dirname(realpath(_local_file)))
 
 
 def task(*args, **kwargs):
+    import pdb; pdb.set_trace()
     pre = list(kwargs.pop('pre', args))
     pre.append(base_task(setup_ctx))
     return base_task(pre=pre, **kwargs)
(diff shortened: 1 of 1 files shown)


assembl/tasks/common.py
```diff
@@ -165,7 +176,11 @@ def setup_ctx(c):
         if not data:
             break
         target = data.get('_extends', None)
-        current = rec_update(data, current)
+        temp_config = rec_update(data, temp_config, excluded_keys=['_extends'])
+
+    if 'DEBUG_END' in os.environ:
+        import pdb; pdb.set_trace()
```
mohamedzayed92: @MathieuVDP  Please remove pub

ayyazdaniaryan:
This PR is poorly merged/rebased. Could you rebase this with the latest of `develop` in order to make it easier to digest? Thanks

PR #706, opened 2019-05-14
Prologation session

mohamedzayed92:


assembl/auth/upgradable_session.py
```diff
@@ -13,8 +13,12 @@ def __init__(self, request, elevated=False,
         self.elevated_expires = elevated_expires
         self.basic_expires = cookie_expires
         self.elevated = False
+        if hasattr(request, 'registry'):
+            timeout = request.registry.settings['session_timeout']
+        else:
```
vincentfretin: Please add a comment to say the `else` if for the tests

PR #709, opened 2019-05-17
Modifying the functions reset_password and send_change_password_email…

mohamedzayed92:
Following the security audit, it has been reported that if a user tries to reset his password when he does not have an account. The requested behaviour is to have a message stating that an email was sent to reset the password (no matter if the user has an account or not). See VLN08

### DIFF
--- a/assembl/views/api2/auth.py
+++ b/assembl/views/api2/auth.py
@@ -39,7 +39,7 @@ from . import (
 from assembl.lib.sqla import ObjectNotUniqueError
 from ..auth.views import (
     send_change_password_email, from_identifier, send_confirmation_email,
-    maybe_auto_subscribe)
+    maybe_auto_subscribe, maybe_send_email)
 from assembl.lib import logging
 
 
@@ -363,13 +363,7 @@ def reset_password(request):
     elif identifier:
         user, account = from_identifier(identifier)
         if not user:
-            if not discussion.preferences['generic_errors']:
-                raise JSONError(
-                    localizer.translate(_("This email does not exist")),
-                    code=HTTPNotFound.code)
-            else:
-                send_change_password_email(request, user, email, discussion=discussion)
-                logger.error("This email does not exist.")
+            maybe_send_email(request, user, email, discussion=discussion)
         if account:
             email = account.email
     else:
@@ -381,18 +375,12 @@ def reset_password(request):
         else:
             email = None
         if not email:
-            if not discussion.preferences['generic_errors']:
-                error = localizer.translate(_("This user has no email"))
-            else:
-                logger.error("This user has no email.")
-                send_change_password_email(request, user, email, discussion=discussion)
+            logger.error("This user has no email.")
+            maybe_send_email(request, user, email, discussion=discussion)
     if not isinstance(user, User):
-        if not discussion.preferences['generic_errors']:
-            error = localizer.translate(_("This is not a user"))
-        else:
-            logger.error("This is not a user.")
-            send_change_password_email(request, user, email, discussion=discussion)
-    send_change_password_email(request, user, email, discussion=discussion)
+        logger.error("This is not a user.")
+        maybe_send_email(request, user, email, discussion=discussion)
+    maybe_send_email(request, user, email, discussion=discussion)
     return HTTPOk()
 
 
--- a/assembl/views/api2/auth.py
+++ b/assembl/views/api2/auth.py
@@ -368,9 +368,7 @@ def reset_password(request):
                     localizer.translate(_("This email does not exist")),
                     code=HTTPNotFound.code)
             else:
-                raise JSONError(
-                    localizer.translate(_(generic_error_message)),
-                    code=HTTPNotFound.code)
+                send_change_password_email(request, user, email, discussion=discussion)
                 logger.error("This email does not exist.")
         if account:
             email = account.email
@@ -378,21 +376,22 @@ def reset_password(request):
         error = localizer.translate(_("Please give an identifier"))
         raise JSONError(error)
     if not email:
-        email = user.get_preferred_email()
+        if user:
+            email = user.get_preferred_email()
+        else:
+            email = None
         if not email:
             if not discussion.preferences['generic_errors']:
                 error = localizer.translate(_("This user has no email"))
             else:
-                error = localizer.translate(_(generic_error_message))
                 logger.error("This user has no email.")
-            raise JSONError(error, code=HTTPPreconditionFailed.code)
+                send_change_password_email(request, user, email, discussion=discussion)
     if not isinstance(user, User):
         if not discussion.preferences['generic_errors']:
             error = localizer.translate(_("This is not a user"))
         else:
-            error = localizer.translate(_(generic_error_message))
             logger.error("This is not a user.")
-        raise JSONError(error, code=HTTPPreconditionFailed.code)
+            send_change_password_email(request, user, email, discussion=discussion)
     send_change_password_email(request, user, email, discussion=discussion)
     return HTTPOk()
 
--- a/assembl/views/api2/auth.py
+++ b/assembl/views/api2/auth.py
@@ -39,7 +39,7 @@ from . import (
 from assembl.lib.sqla import ObjectNotUniqueError
 from ..auth.views import (
     send_change_password_email, from_identifier, send_confirmation_email,
-    maybe_auto_subscribe, maybe_send_email)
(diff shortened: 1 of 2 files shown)


assembl/views/api2/auth.py
```diff
@@ -368,31 +368,30 @@ def reset_password(request):
...
             email = account.email
     else:
         error = localizer.translate(_("Please give an identifier"))
         raise JSONError(error)
     if not email:
-        email = user.get_preferred_email()
+        if user:
+            email = user.get_preferred_email()
+        else:
+            email = None
```
MathieuVDP: Could be refactored to:
```python
if user and not email:
    email = user.get_preferred_email()
```

assembl/views/api2/auth.py
```diff
@@ -368,31 +368,30 @@ def reset_password(request):
                     localizer.translate(_("This email does not exist")),
                     code=HTTPNotFound.code)
             else:
-                raise JSONError(
-                    localizer.translate(_(generic_error_message)),
-                    code=HTTPNotFound.code)
+                send_change_password_email(request, user, email, discussion=discussion)
```
vincentfretin: Why do you call `send_change_password_email` if this will do nothing?

assembl/views/api2/auth.py
```diff
@@ -368,31 +368,30 @@ def reset_password(request):
...
+        else:
+            email = None
         if not email:
             if not discussion.preferences['generic_errors']:
                 error = localizer.translate(_("This user has no email"))
             else:
-                error = localizer.translate(_(generic_error_message))
                 logger.error("This user has no email.")
-            raise JSONError(error, code=HTTPPreconditionFailed.code)
+                send_change_password_email(request, user, email, discussion=discussion)
```
vincentfretin: This doesn't make sense to me to call `send_change_password_email` if this does nothing.
Better `return HTTPOk()` ?

assembl/views/api2/auth.py
```diff
@@ -368,31 +368,30 @@ def reset_password(request):
...
     if not isinstance(user, User):
         if not discussion.preferences['generic_errors']:
             error = localizer.translate(_("This is not a user"))
         else:
-            error = localizer.translate(_(generic_error_message))
             logger.error("This is not a user.")
-        raise JSONError(error, code=HTTPPreconditionFailed.code)
+            send_change_password_email(request, user, email, discussion=discussion)
```
vincentfretin: use `return HTTPOk()` instead? and revert your change in `send_change_password_email`

assembl/views/api2/auth.py
vincentfretin: If you `return` early line 371, you shouldn't need this change.

assembl/views/api2/auth.py
vincentfretin: `return HTTPOk()` instead of calling a function that does nothing

mohamedzayed92: the requested change is to show the message 'We have sent you an email with a temporary connection link. Please use that link to log in and change your password.' So yes it does nothing in the case the email does not exist in the database, and it shows the message. so this is coherent with the logic of the task.

assembl/views/api2/auth.py
mohamedzayed92: Again, it is not just about HTTPOk, the idea is to enter in the `send_change_password_email` and it is this function which decides wether to send an email or not, and in any case it shows the message to the user interface

assembl/views/api2/auth.py
ayyazdaniaryan: I agree with @vincentfretin here. Instead of removing the error and raising another Exception downstream in the `send_change_password_email`, which would fail quite early in the same thread, causing an exception error to be shown (exact opposite of what you're trying to do). In these circumstances, you should consume the error, log it as an error in Assembl, so that we can trace people who tried to bang at this API.

assembl/views/auth/views.py
```diff
@@ -1126,53 +1126,54 @@ def send_change_password_email(
...
-        data.update(dict(
-            discussion_topic=discussion.topic,
-            discussion_url=discussion.get_url()))
-        sender_name = sender_name or discussion.topic
-    if sender_name:
-        sender_name = normalize_email_name(sender_name)
-        sender = '"%s" <%s>' % (sender_name, sender_email)
-        sender_name = Header(sender_name, 'utf-8').encode()
-        if len(sender) > 255:
+    if profile:
```
ayyazdaniaryan: I see what you're trying to do, and it's a good idea to be more defensive in this function against the profile not existing, etc. However, perhaps what you really need is another function, one that either sends the email, or returns an `HTTPOk`, according to the error that is passed into it. The function, if the profile doesn't exist, will do nothing, which is dangerous if a future code change breaks sending the profile, and then nothing will happen, and tracking emails being sent/not-sent is not trivial at all.

assembl/views/api2/auth.py
```diff
@@ -363,37 +363,24 @@ def reset_password(request):
...
     if not isinstance(user, User):
-        if not discussion.preferences['generic_errors']:
-            error = localizer.translate(_("This is not a user"))
-        else:
-            error = localizer.translate(_(generic_error_message))
-            logger.error("This is not a user.")
-        raise JSONError(error, code=HTTPPreconditionFailed.code)
-    send_change_password_email(request, user, email, discussion=discussion)
+        logger.error("This is not a user.")
+        maybe_send_email(request, user, email, discussion=discussion)
```
vincentfretin: We do a check `not isinstance(user, User)` but we lose completely the reason why we shouldn't send the mail in maybe_send_email because we check only if profile is not None there. For example if it's just an AgentProfile instance and not a User instance.
Really I'm not sure about your `maybe_send_email` idea @ayyazdaniaryan 
I would just log the error here and `return HTTPOk()`

assembl/views/api2/auth.py
```diff
@@ -363,37 +363,24 @@ def reset_password(request):
...
+        else:
+            email = None
         if not email:
-            if not discussion.preferences['generic_errors']:
-                error = localizer.translate(_("This user has no email"))
-            else:
-                error = localizer.translate(_(generic_error_message))
-                logger.error("This user has no email.")
-            raise JSONError(error, code=HTTPPreconditionFailed.code)
+            logger.error("This user has no email.")
```
ayyazdaniaryan: In the logs "This user" has absolutely no meaning. Put the identifier there so that we have tracing on the error.

assembl/views/api2/auth.py
```diff
@@ -363,37 +363,24 @@ def reset_password(request):
...
+            maybe_send_email(request, user, email, discussion=discussion)
     if not isinstance(user, User):
-        if not discussion.preferences['generic_errors']:
-            error = localizer.translate(_("This is not a user"))
-        else:
-            error = localizer.translate(_(generic_error_message))
-            logger.error("This is not a user.")
-        raise JSONError(error, code=HTTPPreconditionFailed.code)
-    send_change_password_email(request, user, email, discussion=discussion)
+        logger.error("This is not a user.")
```
ayyazdaniaryan: Same here

assembl/views/auth/views.py
```diff
@@ -1119,6 +1120,20 @@ def send_confirmation_email(request, email, immediate=False):
...
+        subject=None, text_body=None, html_body=None,
+        discussion=None, sender_name=None, welcome=False,
+        immediate=False):
+        if profile:
+            send_change_password_email(request, profile, email,
+            subject, text_body=text_body, html_body=html_body, discussion=discussion,
+            sender_name=sender_name, welcome=welcome, immediate=immediate)
+            log.info('An email has been sent to the profile %s' %  email)
+        else:
+            log.error("This profile does not exist")
```
ayyazdaniaryan: Also, you should definitely pass the identifier to this function, not just the profile, so that your error log can be more clear.

assembl/views/api2/auth.py
gary-bluenove: I agree in this instance that the return HTTPOk() avoids the need to descend into the maybe_send_email

assembl/views/api2/auth.py
```diff
@@ -363,37 +363,24 @@ def reset_password(request):
     elif identifier:
         user, account = from_identifier(identifier)
         if not user:
-            if not discussion.preferences['generic_errors']:
```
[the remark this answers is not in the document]
mohamedzayed92: @ayyazdaniaryan  is it not used any more ? I see it is still used in `assembl_register_user`.

PR #715, opened 2019-05-24
Check date at creation post

MathieuVDP:
Add check when a mutation createPost is received: the date should be between the start and the end dates of the phase associated to the idea.


assembl/graphql/post.py
```diff
@@ -327,6 +327,11 @@ def mutate(root, args, context, info):
         idea_id = int(Node.from_global_id(idea_id)[1])
         in_reply_to_idea = models.Idea.get(idea_id)
 
+        phase = in_reply_to_idea.get_associated_phase()
+        if not (phase.start < datetime.now() < phase.end):
+            error = _('It looks like you do not have the right to do this action. If you think it is an error, please reconnect to the platform and try again.')
```
mohamedzayed92: why not raise an error stating that the phase is closed and you can not participate ?

assembl/tests/fixtures/graphql.py
```diff
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 from datetime import datetime
 import pytest
+from freezegun import freeze_time
```
mohamedzayed92: if you are using a new package please add to the requirements

assembl/tests/views_tests/test_graphql.py
```diff
@@ -738,11 +797,13 @@ def test_mutation_create_post_without_subject(graphql_request, thematic_and_ques
...
 
 
+@freeze_time("2018-3-1")
 def test_mutation_create_post_on_column(graphql_request,
                                         test_session,
-                                        idea_message_column_positive):
-    idea_id = to_global_id('Idea', idea_message_column_positive.idea_id)
-    classifier = idea_message_column_positive.message_classifier
+                                        idea_message_column_positive_on_thread_phase):
+    idea_id = to_global_id('Idea', idea_message_column_positive_on_thread_phase.idea_id)
```
mohamedzayed92: how is it possible to have a positive column on a thread positive ? a column is only possible in multicolumn phase

assembl/tests/fixtures/idea_message_columns.py
```diff
@@ -29,6 +29,18 @@ def fin():
     return column
 
 
+@pytest.fixture(scope="function")
+def idea_message_column_positive_on_thread_phase(request, idea_message_column_positive, subidea_1, subidea_thread_phase_1, test_session):
+    idea_message_column_positive.idea = subidea_thread_phase_1
```
mohamedzayed92: why do it this why and do flush after ? why not define it in the fixture ?

mohamedzayed92: Not sure if it will make a problem in teardown but I would rather create a new `IdeaMessageColumn` from scratch like in `idea_message_column_positive`.

MathieuVDP: @ayyazdaniaryan suggested to me to reuse fixtures in these cases (couple other fixtures are added like this one)

assembl/tests/fixtures/graphql.py
MathieuVDP: Done

assembl/tests/views_tests/test_graphql.py
MathieuVDP: There's no constraint in the backend

PR #716, opened 2019-05-25
Add a script to add semantic analysis section to all discussions

cedrtang:


cedrtang:
@ayyazdaniaryan @pascalhuynh The best solution in my opinion would have been to add the tab if not present when an admin activate the semantic analysis in the administration tool. Then, when deactivated, we just hide it (still present in DB) and don't perform semantic analysis.

Also, as Aryan said, adding a button to launch the watson analysis on all not analyzed past posts would be useful.

PR #717, opened 2019-05-27
Password reinforcement

felixsaintgenis:


felixsaintgenis:
Add a helper next to the password input in the signup component with the password requirements

assembl/static2/js/app/components/login/signupForm.jsx
```diff
@@ -162,16 +170,37 @@ class SignupForm extends React.Component<Props, State> {
...
                         onChange={this.handleInput}
                         required={field.required}
                       />
-                      {field.identifier === 'PASSWORD2' ? (
-                        <p className="annotation no-margin">{I18n.t('login.passwordRequirement')}</p>
+                      {field.identifier === 'PASSWORD' ? (
+                        <Helper
+                          classname="title"
+                          helperText={
+                            <reactFragment>
```
ayyazdaniaryan: `<React.Fragment>` no?

assembl/static2/js/app/components/login/signupForm.jsx
```diff
@@ -146,13 +147,20 @@ class SignupForm extends React.Component<Props, State> {
...
     } = this.props;
     const { privacyPolicyIsChecked, termsAndConditionsIsChecked, userGuidelinesIsChecked } = this.state;
-
     const legalContentsType = {
       termsAndConditions: 'termsAndConditions',
       privacyPolicy: 'privacyPolicy',
       userGuidelines: 'userGuidelines'
     };
 
+    const passwordRequirements = [
```
ayyazdaniaryan: So, THIS is configurable by a discussion preference. This should be fetched by the API that populates `debateData` in the store. And updated accordingly.

ayyazdaniaryan: Good first pass for default case, but can be overwritten, and then this would be wrong. The tricky part is that you need to expose this part of the API to the public. Meaning, the API will be called independent of logged in or not.

assembl/static2/js/app/components/login/signupForm.jsx
felixsaintgenis: OMGey

assembl/static2/js/app/components/login/signupForm.jsx
felixsaintgenis: We decided to merge this default configuration and add the "reste à faire" in a new ticket : https://bluenove.atlassian.net/browse/DEVAS-2130
I will talk with you to make a more precise description of this ticket

assembl/static2/js/app/pages/changePassword.jsx
```diff
@@ -75,14 +77,27 @@ class ChangePassword extends React.Component {
...
-              <FormGroup className="margin-m">
+              <FormGroup className="margin-m align-flex">
                 <FormControl
                   type="password"
                   name="password1"
                   required
                   placeholder={I18n.t('login.newPassword')}
                   onChange={this.handleChangePassword}
                 />
+                <Helper
```
ayyazdaniaryan: This should become a component of its own, taking css classes, children nodes.

assembl/static2/js/app/constants.js
```diff
@@ -106,6 +106,14 @@ export const MESSAGE_VIEW = {
   brightMirror: 'brightMirror'
 };
 
+export const defaultPasswordRequirements = [
```
ayyazdaniaryan: Okay, fair enough. We'll create an endpoint to fetch the specific passwords configured on the page.

assembl/static2/js/app/pages/changePassword.jsx
felixsaintgenis: I understand that putting 3 times the same code in different components is bad, but the helper is already a modularize component that I put around an input. Not sure what you want me to do.

ayyazdaniaryan: Already discussed, in person.

assembl/static2/js/app/utils/translations.js
```diff
@@ -229,8 +229,12 @@ const Translations = {
       username: "Email ou nom d'utilisateur",
       fullName: "Nom complet",
       password2: "Répéter le mot de passe",
-      passwordRequirement:
-        "Votre mot de passe doit contenir au moins une lettre minuscule, une lettre majuscule, un nombre et un caractère spécial.",
+      passwordRequirementIntro: "Votre mot de passe doit contenir au moins:",
```
vincentfretin: Il y a un espace avant les `:` en Français.

PR #719, opened 2019-05-28
Dont show api error messages

mohamedzayed92:
Following our security audit, it turned out that the API show sometimes verbose messages. This PR is to set a default generic message in case an SQLAlchemy exception is raised.


vincentfretin:
I rewrote it in #724 with the French translation.

PR #721, opened 2019-05-28
Create get_assembl_version function, call it in react_base_view

mohamedzayed92:
This PR is to hide the assembl version number from the index_react.jinja2 if the user is not an administrator.

### DIFF
--- a/assembl/views/discussion/views.py
+++ b/assembl/views/discussion/views.py
@@ -38,6 +38,8 @@ from .. import (
     get_locale_from_request, get_theme_info, get_resources_hash, sanitize_next_view)
 from ...nlp.translation_service import DummyGoogleTranslationService
 from ..auth.views import get_social_autologin, get_login_context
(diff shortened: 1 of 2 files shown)


assembl/views/discussion/views.py
```diff
@@ -215,6 +217,20 @@ def react_view(request, required_permission=P_READ):
...
 
 
+def get_assembl_version(discussion):
+    request = get_current_request()
+    user_id = request.authenticated_userid or Everyone
+    if discussion:
+        permissions = get_permissions(
+            user_id, discussion.id)
+        if P_ADMIN_DISC in permissions:
+            return pkg_resources.get_distribution("assembl").version
```
MathieuVDP: After this line, you can just do `return ""` and remove the `else`

mohamedzayed92: done

PR #722, opened 2019-05-29
File upload extension check

MathieuVDP:


assembl/graphql/document.py
```diff
@@ -56,9 +60,14 @@ def mutate(root, args, context, info):
 
         uploaded_file = args.get('file')
         if uploaded_file is not None:
+
             # Because the server is on GNU/Linux, os.path.basename will only work
             # with path using "/". Using ntpath works for both Linux and Windows path
             filename = ntpath.basename(context.POST[uploaded_file].filename)
+            extension = filename.split('.')[~0]
```
ayyazdaniaryan: Such a fancy way to say `-1`

assembl/graphql/document.py
```diff
@@ -56,9 +60,14 @@ def mutate(root, args, context, info):
...
+            if extension not in get_config()['allowed_extensions']:
+                error = _('Sorry, this file type is not allowed.')
```
ayyazdaniaryan: Please also add an info log here that an extension of type X was added. We can aggregate these logs in future to track what types people are adding that perhaps we should incorporate by default. Also, put the current extension type in the error sent back to the client. The client might not even understand what an extension type is.

assembl/graphql/document.py
ayyazdaniaryan: Just a note, file extension is NOT reliable enough to determine the type of file. The `context.POST[uploaded_file].mime_type` isn't much better either unfortunately, and can differ depending on the client. You could make the check that the extension and the broader category match (jpg in `image/jpeg`) for example. But this will add enormous complexity to code, as you'll have to have fallbacks and put extensions into groups (browser determined mime-types are unreliable too).

assembl/configs/base_env.yaml
```diff
@@ -178,3 +178,19 @@ maxretry: 10
 # Can be either a URL or a local filesystem path
 # IF URL, there should be no query strings
 theme_base: ""
+
+# The extensions that are allowed for uploaded files
+# on thread module
+# To be completed if needed
+allowed_extensions:
```
ayyazdaniaryan: What about `csv`, `tsv`, `txt`, `avi`, `mpeg`, `mpg`, `mp4`, `mp3`, `wav`, etc etc. For images, what about `jpe`, `svg`, `bmp`, or google's new `webp`? What if people want to share zip files (which inside could have malicious things inside)? This list is far too limiting as a default. Think about the full mime-type list.

If this list becomes unbearably long, inverting the logic may be a better option. 

https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Complete_list_of_MIME_types
https://www.sitepoint.com/mime-types-complete-list/ 😱

assembl/graphql/document.py
MathieuVDP: Yes, basicly file extension is not enough. But the web server try to guess the engine to use based on it, so if you send a php script call `image.jpg`, the webserver will try to render it as an image and will never call the php engine (if any).
I'll see if I can find a python lib for checking magic numbers.

assembl/configs/base_env.yaml
MathieuVDP: Yes, I agree that this list may be too limited, we should discuss what we want to include. We must stick with the whitelist approach, because blacklist is not a security best practice: what if you block `.php` but the attacker send `.php5` file?
The list of mime-types is really huge, but I don't think we will need more than 10-15 extensions at most. It's in the config, so we can change it easily.

assembl/graphql/document.py
MathieuVDP: There is a lib called `filemagic` which does that, it checks the filetype based on the first bytes of a file.

vincentfretin: We use `python-magic` in https://github.com/assembl/url_metadata to detect the file type. See https://github.com/assembl/url_metadata/blob/8e3ed811a7a2ee0a54eac6bf40b9044b902a6443/url_metadata/__init__.py#L153

assembl/configs/base_env.yaml
vincentfretin: Maybe you should whitelist allowed mimetypes, and not extensions, this way you can allow `image/*` or `video/*` if you want.

assembl/graphql/document.py
vincentfretin: Please use `-1` instead of `~0`.

assembl/configs/base_env.yaml
MathieuVDP: Done with possibility to use `image/*` and also specific like `application/pdf`.

assembl/graphql/document.py
MathieuVDP: Done.

assembl/graphql/document.py
MathieuVDP: Done with `python-magic`.

assembl/configs/base_env.rc
```diff
@@ -167,4 +167,5 @@ maxretry = 10
 
 # To isolate themes, each have been put in a bucket. Declare where the bucket is for access
 # ie. https://<bucket-url>, or on local machine: /opts/assembl_themes/<bucket>/
-theme_base =
\ No newline at end of file
+theme_base =
+allowed_extensions = pdf doc docx ppt pptx xls xlsx jpeg png gif jpg
```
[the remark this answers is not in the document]
MathieuVDP: Done.

PR #725, opened 2019-06-05
Add in CSRF Protection to React

gary-bluenove:
1. Add in 2 new feature flags - to redirect traffic to stargate and to turn on csrf_protection so that we can be default run with it turned 'off'
2. Before each POST request from assembl. graphql or data, make a call to stargate, which in turn calls bond (2 micro services) to generate a CSRF token to be sent.
3. Get the CSRF token from the _csrf cookie set by stargate (no directly accessible in the code or across the wire) and send in the custom http header, X-XSRF-TOKEN
4. stargate validates on all POST requests that the csrf token exists (if validation turned on in stargate as well) and that the csrf token exists still in the redis store. bond the service which was responsable for generating the token has a configurable expire set to 5 minutes so that no token is around for too long to be hijacked and re-used.
5. Fix dependencies in package.json for best practice (adding co for flow control, js-cookie to get at the cookie information and superagent as a much nicer way to make http requests using await / async etc rather than xhr...becuase it's not longer the 90s)


assembl/static2/js/app/utils/httpRequestHandler.js
```diff
@@ -1,24 +1,45 @@
...
   Object.keys(obj)
     .map(k => `${encodeURIComponent(k)}=${encodeURIComponent(obj[k])}`)
     .join('&');
 const getResponseContentType = xhr =>
   // eslint-disable-line no-unused-vars
   xhr.getResponseHeader('Content-Type').split(';')[0];
+
+const useCSRFProtection = document.getElementById('useCSRFProtection')
+  ? document.getElementById('useCSRFProtection').value
+  : null;
```
MathieuVDP: Use false

assembl/static2/js/app/utils/csrf.js
```diff
@@ -0,0 +1,21 @@
...
+export const basePath = () => {
+  const useStargate = document.getElementById('useStargate') ? document.getElementById('useStargate').value : null;
+
+  let path = `${window.location.protocol}//${window.location.host}`;
+
+  if (useStargate === 'true') {
+    path = `${window.location.protocol}//localhost:3000`;
+  }
+
+  return path;
```
MathieuVDP: basePath is not related to csrf, but to the use of stargate or not. Should it be in a separated file, like utils/stargate.js?

assembl/static2/js/app/client.js
```diff
@@ -28,6 +31,9 @@ const myFragmentMatcher = new IntrospectionFragmentMatcher({
 // http://dev.apollodata.com/react/query-splitting.html
 
 const dataIdFromObject = o => o.id;
+const useCSRFProtection = document.getElementById('useCSRFProtection')
```
pascalhuynh: It's a bit confusing here for two reasons:
- the returned value is a string
- the values are 'true' and 'false' formatted as string

Shouldn't we type useCSRFProtection as boolean and map 'true' to TRUE and 'false' to FALSE so that the condition below can be `if (useCSRFProtection)` instead of `if (useCSRFProtection === 'true')`

What do you think ?

assembl/static2/js/app/client.js
```diff
@@ -40,16 +46,28 @@ const networkInterface = createNetworkInterface({
...
-        data: {
-          variables: req.request.variables
-        },
-        level: 'info'
-      });
+      if (!req.options.headers) {
+        req.options.headers = {}; // Create the header object if needed.
+      }
+      co(async () => {
+        if (useCSRFProtection === 'true') {
```
pascalhuynh: depending what you decide on my comment above, this line might have to be updated

assembl/static2/js/app/utils/httpRequestHandler.js
```diff
@@ -1,24 +1,46 @@
...
+
 const convertToURLEncodedString = obj =>
   Object.keys(obj)
     .map(k => `${encodeURIComponent(k)}=${encodeURIComponent(obj[k])}`)
     .join('&');
 const getResponseContentType = xhr =>
   // eslint-disable-line no-unused-vars
   xhr.getResponseHeader('Content-Type').split(';')[0];
+
+const useCSRFProtection = document.getElementById('useCSRFProtection')
```
pascalhuynh: same comment as in `assembl/static2/js/app/client.js`. Since it is the exact same condition as above, we can refactore some code here and import the required utility function here and in `assembl/static2/js/app/client.js` instead, what do you think ?

assembl/static2/js/app/utils/httpRequestHandler.js
```diff
@@ -1,24 +1,46 @@
...
     const xhr = new XMLHttpRequest();
-    const url = obj.url;
+    const url = urljoin(basePath(), obj.url);
     let payload = obj.payload;
     xhr.open(obj.method, url);
     if (obj.method.toLowerCase() === 'post') {
       obj.headers = obj.headers || {}; // eslint-disable-line
+
+      // Go and fetch a CSRF token for the POST request if activated
+      if (useCSRFProtection === 'true') {
```
pascalhuynh: same comment as in assembl/static2/js/app/client.js

assembl/static2/js/app/utils/server.js
```diff
@@ -0,0 +1,16 @@
+/** *
+ * Use the standard assembl server location for API calls or Stargate
+ *
+ * @return String path - the path to use for the API call
+ * */
+export const basePath = () => {
+  const useStargate = document.getElementById('useStargate') ? document.getElementById('useStargate').value : 'false';
```
pascalhuynh: same comment as in `assembl/static2/js/app/client.js`: it's a bit confusing having a boolean as a string

gary-bluenove: It's a field in HTML - I thought it was safer to pull as a string rather than boolean

assembl/static2/js/app/utils/httpRequestHandler.js
gary-bluenove: See below

assembl/static2/js/app/client.js
gary-bluenove: I would normally but the config comes from the HTML field and I thought it 'safer' to work using strings rather than assuming booleans would be preserved in the <... value=true>

assembl/static2/js/app/client.js
```diff
@@ -40,16 +46,28 @@ const networkInterface = createNetworkInterface({
...
+          const responseToken = await getCSRFToken();
+          if (responseToken.text === 'ok') {
+            req.options.headers['X-XSRF-TOKEN'] = Cookies.get('_csrf');
```
ayyazdaniaryan: Maybe `X-CSRF-TOKEN` name instead? No preference though...

assembl/static2/js/app/utils/csrf.js
```diff
@@ -0,0 +1,11 @@
+import request from 'superagent';
```
ayyazdaniaryan: Why not just use `fetch`? It's an extra dependency for only a subset of calls made in non-graphql.

assembl/static2/js/app/utils/httpRequestHandler.js
```diff
@@ -1,24 +1,46 @@
+import urljoin from 'url-join';
+import Cookies from 'js-cookie';
```
ayyazdaniaryan: So we have a few helper functions `cookie` and `getCookieItem` from `js/app/utils/globalFunctions.js`. My thinking is, either use them, or we do a small refacto and use `js-cookie` everywhere. This package looks really small and heavily tested across all browsers, so I'm not opposed to it. The front-end code is hairy as is, and this just adds another layer of confusion for a front-end dev to decide what to use to access/manipulate cookies.

assembl/static2/js/app/utils/httpRequestHandler.js
```diff
@@ -1,24 +1,46 @@
...
 /*
   A global async method that returns a Promisified ajax call
   @params payload [Object] The object that will be sent
   @params isJson [Boolean] Pass a flag if the object is JSON. Default is form header.
   @retuns [Promise]
 */
+
 export const xmlHttpRequest = obj =>
-  new Promise((resolve, reject) => {
+  new Promise(async (resolve, reject) => {
```
ayyazdaniaryan: Could just remove the `new Promise` and make xmlHttpRequest an `async` func and await the result/error then

assembl/static2/js/app/utils/csrf.js
```diff
@@ -0,0 +1,11 @@
...
+import urljoin from 'url-join';
+
+import { basePath } from './server';
+
+export async function getCSRFToken() {
+  // Get CSRF token, note the response contains nothing as the cookie is set directly!!
+  const responseToken = await request.get(urljoin(basePath(), '/v1/generateToken')).withCredentials();
```
ayyazdaniaryan: Is the logic for `/v1/generateToken` hosted on gitlab? Possible open a PR for that, or at least link to the project so that we can CR that too. I think it's important to look at that code too.

assembl/static2/js/app/utils/server.js
```diff
@@ -0,0 +1,16 @@
...
+
+  let path = `${window.location.protocol}//${window.location.host}`;
+
+  if (useStargate === 'true') {
+    path = `${window.location.protocol}//${window.location.hostname}:3000`; // move to config ?
```
ayyazdaniaryan: The port should be hard-coded? Pass as config as the comment suggests (pull from html).

assembl/static2/js/app/client.js
```diff
@@ -40,16 +46,28 @@ const networkInterface = createNetworkInterface({
...
+          }
+        }
 
-      next();
+        Sentry.addBreadcrumb({
```
ayyazdaniaryan: Nah, keep the `next()` above, and put Sentry in the next index position. They are different middlewares, CSRF and Sentry, but right now, they're both one big middleware.

assembl/static2/js/app/client.js
```diff
@@ -40,16 +46,28 @@ const networkInterface = createNetworkInterface({
...
-          variables: req.request.variables
-        },
-        level: 'info'
-      });
+      if (!req.options.headers) {
+        req.options.headers = {}; // Create the header object if needed.
+      }
+      co(async () => {
+        if (useCSRFProtection === 'true') {
+          const responseToken = await getCSRFToken();
```
ayyazdaniaryan: Okay, this will make `bond` ms a bottleneck, because every graphQL call is a POST, and you're fetching a new token for every query and mutation. You should peek internally into the request to see if it's a mutation or a query, and only `getCSRFToken()` when it's a mutation.

assembl/static2/js/app/client.js
```diff
@@ -1,10 +1,13 @@
 import { ApolloClient, toIdValue, IntrospectionFragmentMatcher } from 'react-apollo';
 import { createNetworkInterface } from 'apollo-upload-client';
 import * as Sentry from '@sentry/browser';
+import Cookies from 'js-cookie';
+import co from 'co';
```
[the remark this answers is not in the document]
gary-bluenove: You can't use async / await inside a regular callback - it has to be itself wrapped inside something that is async - this block of code is a middleware `applyMiddleware: function (req, next) {`so you cannot yield or async / await, and so it's important to wait for the token before advancing otherwise javascript just keeps on processing. Try for yourself, you can't add an async - you can possible add a Promise but this is cleaner IMHO

assembl/static2/js/app/client.js
gary-bluenove: Ripped this from elsewhere wither CSRF or XSRF seems to be the standard - you want me to change before commit ?

assembl/static2/js/app/utils/csrf.js
gary-bluenove: You're right!

assembl/static2/js/app/utils/httpRequestHandler.js
gary-bluenove: I personally think the cookie stuff we're going in global functions is hairy and 'old school':

`export function setCookieItem(name: string, value: any, sensitive: boolean = false) {
  const date = new Date();
  date.setMonth(date.getMonth() + 13);
  document.cookie = cookie(name, value, date.toString(), sensitive);
}``

I don't see the need to create proprietary functions for things that have been simplified with packages, unless there's something I'm missing - I HATE it's accessing the document in our react code - I've never done this or felt the need to in my life - it's a hack.

assembl/static2/js/app/utils/csrf.js
gary-bluenove: It's stargate and committed to master

assembl/static2/js/app/utils/server.js
gary-bluenove: You're right, not too happy about pulling from HTML but it should not be hard-coded, will change

assembl/static2/js/app/client.js
gary-bluenove: Didn't realise I had removed this!

gary-bluenove: I moved it to below the sentry - you sure?

assembl/static2/js/app/client.js
gary-bluenove: I agree but I'm not sure if that covers what we're trying to achieve. For example the OWASP standards on CSRF don't care if it's graphql and a query or mutation, it cares only that POST requests are not susceptible. Also, we don't have a choice, in order to add CSRF protection we will have a slow down somewhere for the extra round-trips. Happy to make a change but I would suggest we release this as a first version and see what the penalty looks like.

assembl/static2/js/app/utils/httpRequestHandler.js
gary-bluenove: I think that changes upstream code. For example, the code using this function is written like:

```
  return xmlHttpRequest({
    method: 'POST',
    url: route,
    isJson: true,
    payload: payload
  });
```

We would have to change it all for await xmlHttpRequest

assembl/static2/js/app/utils/server.js
gary-bluenove: Se should chat about this, not happy about people 'seeing' our port in the html code - it negates the fact it's hidden behind the react  - let's discuss

assembl/static2/js/app/client.js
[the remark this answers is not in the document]
ayyazdaniaryan: Okay fair. Se queda.

assembl/static2/js/app/client.js
ayyazdaniaryan: Nah, if you're happy with this option, we can keep it. I have no hard ops on this.

assembl/static2/js/app/utils/httpRequestHandler.js
ayyazdaniaryan: Not sure if accessing `document` from react is a hack at all. Even the js-cookie accesses it. I just read the source of `js-cookie`. Let's keep it, because it makes further checks on what's in the cookie, which we would have to implement in the future should we keep our code. However, might not be a bad idea to write a TODO comment in the `globalFunctions` to migrate these funcs to use `js-cookie`.

assembl/static2/js/app/utils/httpRequestHandler.js
ayyazdaniaryan: Okie

assembl/static2/js/app/utils/csrf.js
ayyazdaniaryan: Super, I'll take a look at stargate code as well. Anyway to open up a PR there? Or at least, I can look at it and give you my impressions. It has no bearing on this PR though.

assembl/static2/js/app/utils/server.js
ayyazdaniaryan: Okay, alternatives:

1. Stargate API must have an API that returns the port, this API call made by Assembl and is awaited before any rendering (this must be the first API call of the application)
2. We pass a js variable in jinja2 which will contain our `configs`. We'll have to post-process our HTML prior to sending to client be minified/obfuscated (slows down time to first byte)
3. Do nothing on HTML. It's not the place of the html to be secure. It's the place of Stargate to be hardened and defend against attacks at this port

I like (3)

assembl/static2/js/app/client.js
ayyazdaniaryan: Yeah, it's a middleware array. You're adding another middleware, so you want your middleware func to follow the pattern

assembl/static2/js/app/client.js
ayyazdaniaryan: Hmm, yeah OWASP is covering the base HTTP protection level. Okay, let's pay the price for now and see if we can win the time back elsewhere. We'll keep a close watch on response times. We're already quite terrible (+2s). There are many other places to optimize the response speed (code-splitting, anyone?)

assembl/static2/js/app/utils/server.js
ayyazdaniaryan: Or:

4. Put a path under Assembl, so you benefit from TLS encryption and have nginx redirect requests to port 3000, wherever this machine is hosted.

assembl/static2/js/app/utils/csrf.js
gary-bluenove: Please do take a look at Stargate - no review because I looked around and asked myself who here is better than me at javasccript :-)

assembl/static2/js/app/client.js
gary-bluenove: OK

assembl/static2/js/app/utils/server.js
gary-bluenove: Can't do 4, makes me want to puke.
Let's do 3

PR #726, opened 2019-06-05
Add a share button on survey posts

tdesvenain:
Fixes DEVAS 1756


assembl/static2/js/app/components/debate/survey/posts.jsx
```diff
@@ -24,12 +24,15 @@ type Props = {
   posts: {
     edges: Array<PostNode>
   },
+  params: Object,
```
pascalhuynh: `Object` is a generic type, can you be more specific here ?

assembl/static2/js/app/components/debate/survey/post.jsx
```diff
@@ -291,7 +305,28 @@ class Post extends React.Component<Props> {
       tagList: formatedTagList(post.tags),
       suggestedTagList: formatedSuggestedTagList(post.keywords)
     };
-
+    const slug = getDiscussionSlug();
+    if (!slug) {
+      return <></>;
```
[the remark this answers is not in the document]
vincentfretin: This is the short version of `<React.Fragment></React.Fragment>`. @tdesvenain You can just return `null` too.

assembl/static2/js/app/components/debate/survey/posts.jsx
tdesvenain: now i have created a RouterParams type, yes, i guess i can

assembl/static2/js/app/components/debate/survey/post.jsx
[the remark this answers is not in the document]
tdesvenain: ok, i fix it. vincent already told me you don't do that (as a typescript developer i sometimes feel bad about returning null)

assembl/static2/js/app/components/debate/survey/post.jsx
```diff
@@ -376,6 +416,11 @@ class Post extends React.Component<Props> {
           </div>
           {screenWidth < SMALL_SCREEN_WIDTH && (
             <div className="actions">
+              {!isPending ? (
```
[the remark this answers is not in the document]
tdesvenain: for the moment, i don't think we should: this would not be consistent with the other developements
by the way, i think that a const shareButton should be static, and should not contain logic, and its type should not be "or null"
imho, this makes sense to have if / else logic about wether or not displaying content in the template part of the render

assembl/static2/js/app/components/debate/survey/posts.jsx
tdesvenain: i encounter some issues to use the RouterParams type there, 
a a beginner in flow, i will ask for vincent'help when he comes back
for the moment, i merge the request

tdesvenain: (when @pascalhuynh will have reviewed my last commit)

vincentfretin:
I see you created `js/app/type.flow.js`. For types used in several places, you can put the types in `flow/object_types.js` (no need to import the type)

assembl/static2/js/app/components/debate/survey/posts.jsx
vincentfretin: I just tested putting `params: RouterParams,` here. I have no errors.

vincentfretin:
I see you didn't import `type.flow.js` but it indeed works. Why did we import the types for the other `types.flow.js then? :thinking:

PR #730, opened 2019-06-06
Fixes on votes

tdesvenain:
fix bugherd #274

### DIFF
--- a/assembl/static2/js/app/pages/voteSession.jsx
+++ b/assembl/static2/js/app/pages/voteSession.jsx
@@ -71,7 +71,6 @@ type Props = {
   proposals: Array<Proposal>,
   propositionsSectionTitle: string,
   randomProposals: Array<Proposal>,
-  refetchVoteSession: Function,
   seeCurrentVotes: boolean,
   subTitle: string,
   title: string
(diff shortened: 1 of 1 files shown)


assembl/static2/js/app/pages/voteSession.jsx
```diff
@@ -56,23 +56,24 @@ export type Proposal = {|
 |};
 
 type Props = {
-  loading: boolean,
-  title: string,
-  subTitle: string,
-  seeCurrentVotes: boolean,
+  addGaugeVote: Function,
```
pascalhuynh: Could you be more specific on the function type ?

assembl/static2/js/app/pages/voteSession.jsx
```diff
@@ -56,23 +56,24 @@ export type Proposal = {|
...
+  addTokenVote: Function,
```
pascalhuynh: Same here

tdesvenain: this is not related with the pr, this is just a move for alphabetic order

assembl/static2/js/app/pages/voteSession.jsx
tdesvenain: i improved the return type, this is impossible to improve the parameter types

PR #737, opened 2019-06-07
Replace standard optimizer which hangs on 92% and sucks the life out …

gary-bluenove:
…of machines with Terser

### DIFF
--- a/assembl/static2/webpack.config.js
+++ b/assembl/static2/webpack.config.js
@@ -9,7 +9,7 @@ var UglifyJsPlugin = require('uglifyjs-webpack-plugin');
 var HtmlWebpackPlugin = require('html-webpack-plugin');
 var glob = require('glob');
 var _ = require('lodash');
-
+const TerserPlugin = require('terser-webpack-plugin');
 
 function theme_entries() {
     var entries = {},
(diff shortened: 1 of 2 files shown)


assembl/static2/webpack.config.js
```diff
@@ -132,9 +132,10 @@ module.exports = {
     },
     mode: 'production',
     optimization: {
-        minimizer: [
-            new UglifyJsPlugin({ sourceMap: true, parallel: true, cache: true })
-        ]
+      minimizer: [new TerserPlugin({
+        cache: true,
+        parallel: true,
```
vincentfretin: Why did you remove `sourceMap: true`?

gary-bluenove: No good reason except it slows down the compilation. I'm in testing mode with this at the moment but I needed it in the develop branch for CI purposes. I need to also move the dep to devdep if we keep it. However I might just revert the commit if I don't see the improvements. At the moment it looks ×2 faster

vincentfretin: uglifyjs-webpack-plugin is unmaintained. Recent webpack version replaced uglifyjs-webpack-plugin  by terser-webpack-plugin. You should actually remove uglifyjs-webpack-plugin from the dependency, upgrade webpack, remove the the optimizer section in webpack.config.js.
I have the change in a old branch I never pushed. I can do it again monday if you want.
And you should commit the yarn.lock file when you add or update dependencies like that.

PR #741, opened 2019-06-10
Update packages

vincentfretin:
This is a huge change. This needs to be properly QAed before merge.
- react-redux 5.0.6 => 5.1.1
- react-select 2.1.1 => 2.4.4
- react-share 1.16.0 => 1.19.1 (minor update)
- redux-thunk 2.2.0 => 2.3.0
- babel 7.3.4 => 7.4.5 (We still use core-js 2 and not 3 because of some storybook dependencies like `@storybook/channels` is importing from core-js 2)
- @material-ui/core 3.9.0 => 3.9.3, @material-ui/lab 3.0.0-alpha.28 => 3.0.0-alpha.30 (for watson keywords slider)
- apollo-client 1.9.1 => 1.9.3 (minor update)
- react-apollo 1.4.12 => 1.4.16 (minor update)
- react-tweet-embed 1.0.8 => 1.2.2
- react-color 2.14.1 and 2.17.0 => 2.17.3
- react-hot-loader 4.3.11 => 4.11.0
- bootstrap 3.3.7 => 3.4.1
- draft-convert 2.1.4 => 2.1.5
- react 16.3.2 => 16.8.6
- storybook 4.1.13 => 4.1.18
- enzyme 3.4.1 => 3.10.0
- graphql-tag 2.4.2 => 2.10.1
- graphql-tools 1.1.0 => 1.2.3
- jquery 3.3.1 => 3.4.1 (pour l'attrapage)
- prettier-eslint 8.2.5 => 8.2.6 (minor update)
- prettier-eslint-cli 4.4.2 => 4.4.3 (minor update)
- prettier 1.9.1 => 1.9.2 (minor update)
- rc-slider 8.6.5 => 8.6.13 (used for vote sessions gauges)
- webpack 4.25.1 => 4.33.0 with terser instead of uglify for minification
- final-form 4.14.0 => 4.15.0
- react-dates 18.4.1 => 18.5.0
- searchkit 2.3.0 => 2.3.2

not updated:
- major update of prettier
- major update of eslint
- major update of redux and react-redux
- major update of apollo-client/react-apollo/apollo-upload-client


assembl/static2/package.json
```diff
@@ -54,14 +54,14 @@
...
     "enzyme": "^3.4.1",
-    "enzyme-adapter-react-16.3": "^1.4.1",
-    "eslint": "^4.6.1",
-    "eslint-config-airbnb": "^15.1.0",
-    "eslint-friendly-formatter": "^3.0.0",
-    "eslint-plugin-flowtype": "^2.35.1",
-    "eslint-plugin-import": "^2.7.0",
-    "eslint-plugin-jsx-a11y": "^5.1.1",
-    "eslint-plugin-react": "^7.3.0",
+    "enzyme-adapter-react-16": "^1.14.0",
```
pascalhuynh: I think, since we are updating libs versions, we can also use strict version for every library in the package.json file instead of `~` or `^` or `~>`. That way we are sure that libraries are 1:1 in every single local machine and remote machine

vincentfretin: I disagree. You should avoid pinning version in `package.json`. This is what `yarn.lock` is for. You can then use `yarn upgrade-interactive` to see the update and decide to update the packages.

vincentfretin: Libraries that pin their dependencies is a real problem like `searchkit` that pinned an old version of `lodash` or `react-tags` that pinned everything.

vincentfretin: FYI, for the `lodash` dependency in `searchkit`, this is overriden with yarn resolutions, see in `package.json` :  ` "resolutions": { "*/lodash": "^4.17.11" }`

gary-bluenove: personally I've never worked in an environment where the deps are NOT pinned. In the exampmle of searchkit and lodash, it might be for a VERY good reason that they have not bumped the version of lodash (probably not though). 

In essence we don't wan't a package to get bumped and it breaks our software. yes the .lock files prevent this but they should not be used as the dependence either. For example, static2 does not build with npm, only yarn because there is no package-lock.json so we have ENFORCED using yarn - that should never happen. 

My rule - fix the deps, do not enforce a build tool (yarn or npm) and do not check in .lock files - you don't need to if you're versions are fixed.

I'm putting in place renovate on the repositories for check for updates in order to stay up to date: https://renovatebot.com/

vincentfretin: Pinning in `package.json` is not enough because you don't pin versions of dependencies of dependencies. You need a lock file as well.

vincentfretin: The only way to properly freeze all versions is using `npm ci` (not `npm install`) with `package-lock.json` OR `yarn` with `yarn.lock`. We chose `yarn` for static2 before the implementation of the `package-lock.json` in npm 5 (we were using node 6 with npm 3 at that time, npm 5 wasn't released). Maybe we can switch now to `npm` instead of `yarn` for static2.

gary-bluenove: npm and yarn are a lot closer - I think the dev should be free to choose but we should commit both files and be clear how to build using either case. I think the fabric scripts update the version of npm on your machine and then use yarn for the build anyway. Let's make this outside the scope of this PR unless you want to commit a package-lock.json with it

vincentfretin: fabric install a specific version of node and npm yes, and it updates it if you don't have those versions. It's using wrongly `npm update` (should really be `npm ci`) for static, and `yarn` for static2. You need to chose between using `npm` or `yarn`, you can't commit both `package-lock.json` and `yarn.lock`, the developer will never do both, the files will be out of sync.

pascalhuynh: Well then, I don't want to block this PR so let's move on and we will adapt later with specific coding rules

vincentfretin: FYI We currently use the feature yarn workspaces that doesn't exist with npm, and yarn resolutions.

vincentfretin:
I will rebase and merge this monday, after you deployed current develop to prod.

vincentfretin:
Yes I will do it.

vincentfretin:
I rebased, updated again yarn.lock
but I kept searchkit 2.3.0 (2.3.2 introduced a warning on SearchBox input) and kept react-datepicker 1.5.0 (1.8.0 messed up the styles for hours)

vincentfretin:
Actually no with searchkit 2.3.0 I had the warning as well, so let's update it to searchkit 2.3.2.

vincentfretin:
I reverted last minute update of final-form https://github.com/assembl/assembl/commit/93b7b0585638c5d24077d1e83691c8ac66cc1ee5
because latest version of final-form updated the flow types and this fail for us
https://travis-ci.org/assembl/assembl/jobs/547302371

PR #743, opened 2019-06-11
Password reinforcement

MathieuVDP:
Update base_env configs to use password rules by default.

### DIFF
--- a/assembl/configs/base_env.rc
+++ b/assembl/configs/base_env.rc
@@ -173,3 +173,7 @@ theme_base =
 # on thread module
 # To be completed if needed
 attachment_allowed_mime_types = image/\w+ video/\w+ application/(pdf|msword) application/vnd\.(.*ms.*|.*officedocument.*)
+
+# Password requirements
+minimum_password_length = 9
+password_required_classes = [{"[a-z]": {"en": "lower-case letter"}, "[A-Z]": {"en": "upper-case letter"}, {"\\d": en: "digit"}, {"\\W": en: "special character"}]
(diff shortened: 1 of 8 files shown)


assembl/configs/base_env.rc
```diff
@@ -167,4 +167,6 @@ maxretry = 10
 
 # To isolate themes, each have been put in a bucket. Declare where the bucket is for access
 # ie. https://<bucket-url>, or on local machine: /opts/assembl_themes/<bucket>/
-theme_base =
\ No newline at end of file
+theme_base =
+minimum_password_length = 9
+password_required_classes = '{"[a-z]": {"en": "lower-case letter"}, "[A-Z]": {"en": "upper-case letter"}, "\\d": {"en": "digit"}, "\\W": {"en": "special character"}}'
```
ayyazdaniaryan: The regex escape looks off. Have you tried putting `\\d` instead? I'm certain that's what causes an issue with the tests not passing.

MathieuVDP: I suppose you mean `\d`. I did, but then the regex is wrong when I generate the local.ini, because it needs to be escaped there. And if I remove the single quotes, it doesn't work neither...
I don't know how to fix it right now...

assembl/graphql/user.py
```diff
@@ -172,13 +172,6 @@ def mutate(root, args, context, info):
                 if old_password == new_password:
                     raise Exception(u"004: The new password has to be different than the current password.")
 
-                from ..auth.password import verify_password
```
ayyazdaniaryan: No, do not remove this logic. It's the check we make for the new password vs all of the old passwords

MathieuVDP: This check is already done in `validate_password` in models/auth.py. This is duplicated, and I think the right place is in `validate_password`, which is called by `password_p`

ayyazdaniaryan: I see what you mean. Very well. I have added a new commit for an extra check that was done here, to be done in the downstream.

PR #744, opened 2019-06-12
Bug fixes from Bugherd

pascalhuynh:
Just one quick fix, the other ticket is still on WIP
* Fixes BH-301: display semantic analysis section when the preference is set


assembl/static2/js/app/components/navbar/navbar.jsx
```diff
@@ -223,15 +227,17 @@ export default compose(
             error: data.error,
             loading: data.loading
           },
-          logoData: null
+          logoData: null,
+          isSemanticAnalysisEnabled: null
```
vincentfretin: It may be better to have it at `false` if it's a boolean?

pascalhuynh: true that

PR #747, opened 2019-06-14
Update default configs with mimetypes

MathieuVDP:


assembl/configs/base_env.yaml
```diff
@@ -186,3 +186,8 @@ attachment_allowed_mime_types:
   - image/*
   - video/*
   - application/pdf
+  - application/msword
+  - application/vnd.openxmlformats-officedocument.wordprocessingml.document
+  - application/vnd.ms-powerpoint
+  - application/vnd.openxmlformats-officedocument.presentationml.presentation application/vnd.ms-excel
+  -	application/vnd.openxmlformats-officedocument.spreadsheetml.sheet
```
vincentfretin: The last two points are weird. Don't you need one mimetype per line?

MathieuVDP: Yes, I missed a newline, and there was a tab instead of space, I didn't catch it on atom. Fixed.

assembl/configs/base_env.yaml
```diff
@@ -186,3 +186,9 @@ attachment_allowed_mime_types:
   - image/*
   - video/*
   - application/pdf
+  - application/msword
+  - application/vnd.openxmlformats-officedocument.wordprocessingml.document
```
ayyazdaniaryan: You could also do a regex which will capture all of the list that you've just added, including perhaps other ms or office types that you may have missed:
```
r"application\/vnd\.(.*ms.*|.*officedocument.*)"
```

ayyazdaniaryan: But you'll have to make sure that the check downstream does a regex check

MathieuVDP: Good point, I'll implement this to get a shorter list.

PR #748, opened 2019-06-14
fix: display a supported wordcloud for IE 6-11

pascalhuynh:


assembl/static2/js/app/utils/globalFunctions.js
```diff
@@ -381,4 +381,8 @@ export const formatedTagList = (tags: Tags): Array<TagProps> =>
       }
       return result;
     }, [])
-    : []);
\ No newline at end of file
+    : []);
+
+// Function that checks wheter the browser in IE from v6 to v11
+// eslint-disable-next-line
+export const isIE = () => /*@cc_on!@*/ false || !!document.documentMode;
```
ayyazdaniaryan: Is this the best way? Common sense says use the `useragent`, and seems like `Trident` is the domain of IE 11 and below. https://stackoverflow.com/a/22242528

pascalhuynh: I got the info from here: https://stackoverflow.com/questions/9847580/how-to-detect-safari-chrome-ie-firefox-and-opera-browser. It got 1463 upvotes...

ayyazdaniaryan: Alright, looks like I was wrong. Useragent can be spoofed. Vas y

pascalhuynh:
@ayyazdaniaryan just updated the branch. Thanks for the review, ill update develop right after your approval.

pascalhuynh:
Thanks @ayyazdaniaryan will merge after travis check

PR #750, opened 2019-06-18
Bug fixes bh

cedrtang:
- BH179: fix tags when created/deleted and changing page

- BH315: fix tabs opacity to be more visible

- BH88: click on fullscreen images display modal

- BH257: when updating profile, username is not sent if not mandatory


vincentfretin:
Do you have commited a configs git submodule? Travis fail because of that. You need to remove it.

PR #756, opened 2019-06-20
Change beaker session to use secure cookies

gary-bluenove:


gary-bluenove:
So it turns out, after turning on the AWS ELB to use 443 to nginx and then generating a self signed cert and then playing with the proxy command in nginx that none of this is that important:

Add 1 line to beaker session and it sends the secure flag.

https://beaker.readthedocs.io/en/latest/modules/session.html

I think it's still good practice to use SSL throughout all the chain but that will also require more terraform and AMI changes

PR #758, opened 2019-06-21
Fun console message for curious devs

gary-bluenove:
Just a bit of fun. If we want to promote the company as tech driven, let's show them we know a thing or two


cedrtang:
LGTM

PR #759, opened 2019-06-21
add a scroll after sending a new multicolumn post

felixsaintgenis:


assembl/static2/js/app/components/debate/common/topPostForm.jsx
```diff
@@ -175,6 +175,9 @@ export class DumbTopPostForm extends React.Component<Props, State> {
               }
               this.resetForm();
               this.setState(submittingState(false));
+              if (ideaOnColumn) {
```
ayyazdaniaryan: `this.setState` is async, so don't do this scrolling action until setState is done. Either put it in a callback, or chain it in another `then`, to scroll to the top

PR #764, opened 2019-07-04
New thread menu

felixsaintgenis:
I open the PR to have the first reviews but I think this new menu will need another round of changes (wording and styling changes).

I still have key warning messages that I strangely cannot resolved.


assembl/static2/css/common.scss
```diff
@@ -84,7 +84,7 @@ textarea.placeholder {
 }
 
 .app-content {
-  margin: $navbar-height 0 0;
+  margin: 40px 0 0;
```
gary-bluenove: Fixed, always, mobile ?

felixsaintgenis: I don't know why but this grey space between the navbar and the header appears. 40px instead of $navbar-height which is 60px fix the problem. I didn't wanted to change the value of navbar-height because it is used multiple times in the project. 
Capture d’écran 2019-07-08 à 13 32 01

gary-bluenove: Ship it anyway!

PR #771, opened 2019-07-17
Check the file size before pushing and updating the index.html

gary-bluenove:


assembl/tasks/build.py
```diff
@@ -469,8 +469,20 @@ def push_wheelhouse(c, house=None):
...
     tmp_wheel_path = house or os.path.join(c.config.code_root, 'wheelhouse')
+
+    # Check assembl wheel is not zero bytes - don't push this!!
+    (version, num, commit_hash, commit_tag, branch) = git_version_data(c)
+    base_name = create_wheel_name(version, num, commit_hash=commit_hash)
+
+    assembl_wheel = os.path.join(tmp_wheel_path, base_name)
+    assembl_wheel_size = os.path.getsize(assembl_wheel)
+
+    if not assembl_wheel_size > 0:
```
cedrtang: @gary-bluenove Is there a reason you're not using `if assembl_wheel_size == 0:` ?

cedrtang:
Had just one question, otherwise LGTM

assembl/tasks/build.py
gary-bluenove: No good reason! Will change

PR #780, opened 2019-08-03
Devas 2035 export users

cedrtang:
Add administration page to export users data

### DIFF
--- a/assembl/tests/views_tests/test_api2.py
+++ b/assembl/tests/views_tests/test_api2.py
@@ -1154,14 +1154,14 @@ class TestUserExport(AbstractExport):
     ]
 
     participant1_data = ['A. Barking Loon', 'abloon@gmail.com', 'Test.Username', '2000-01-02 00:00', '2000-01-02 00:00', '2000-01-15 00:00',
-                        '1', '1', '1', '0', '2', '0', '1', '1', '0', '1', '0', '1', '1', '', 'Direction, HR', '[u\'emails\', u\'public_profiles\']']
+                        '1', '1', '1', '0', '2', '0', '1', '1', '0', '1', '0', '1', '1', '', 'HR', '[u\'emails\', u\'public_profiles\']']
 
     participant1_data_missing_one_post = ['A. Barking Loon', 'abloon@gmail.com', 'Test.Username', '2000-01-02 00:00', '2000-01-02 00:00', '2000-01-15 00:00',
-                        '0', '0', '1', '0', '1', '0', '0', '0', '0', '1', '0', '1', '1', '', 'Direction, HR', '[u\'emails\', u\'public_profiles\']']
+                        '0', '0', '1', '0', '1', '0', '0', '0', '0', '1', '0', '1', '1', '', 'HR', '[u\'emails\', u\'public_profiles\']']
 
     def test_base(self, test_session, test_app, discussion, user_language_preference_en_cookie, participant1_user, participant1_username, reply_post_2,
                 reply_1_sentiments, reply_2_sentiments, agent_status_in_discussion_user2_visits, post_related_to_sub_idea_1_participant2,
-                participant1_social_account_w_extra, participant1_social_account_w_extra2, agent_status_in_discussion_user1_visits, subidea_1):
+                participant1_social_account_w_extra, agent_status_in_discussion_user1_visits, subidea_1):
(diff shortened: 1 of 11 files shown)


assembl/views/api2/discussion.py
```diff
@@ -667,6 +667,143 @@ def multi_module_csv_export(request):
...
+        ).filter(m.Post.creation_date < end
+        ).order_by(m.Post.id).all()
+
+    sentiment_counts = db.query(m.Post.id, m.SentimentOfPost.actor_id, m.SentimentOfPost.type).join(m.SentimentOfPost).filter(
+        m.Post.id.in_([post.id for post in posts]),
+        m.SentimentOfPost.tombstone_condition()
+    ).group_by(m.Post.id, m.SentimentOfPost.actor_id, m.SentimentOfPost.type).all()
+
+    # Sort sentiments for better access depending of post/user
+    sentiments_given_by_user = defaultdict(lambda: defaultdict(int))
```
ayyazdaniaryan: So, these things can be calculated by using the `count` func and the `group_by` in the query above.

assembl/views/api2/discussion.py
```diff
@@ -667,6 +667,143 @@ def multi_module_csv_export(request):
...
+            for key, value in account.extra_data.iteritems():
+                if key not in fieldnames:
+                    fieldnames.append(key)
+                if key not in users[user.id].keys():
+                    users[user.id][key] = value
+                else:
+                    users[user.id][key] += ', ' + value
+
+    for post in posts:
+        if post.creator.id not in users.keys():
```
ayyazdaniaryan: `post.creator.id` is repeated a lot, can create a variable for it and call the var instead

assembl/views/api2/discussion.py
```diff
@@ -667,6 +667,143 @@ def multi_module_csv_export(request):
...
+                AGREE_RECEIVED: 0,
+                DISAGREE_RECEIVED: 0,
+                DONT_UNDERSTAND_RECEIVED: 0,
+                MORE_INFO_RECEIVED: 0,
+                TOP_POSTS: 0,
+                TOP_POST_REPLIES: 0,
+                POSTS: 0,
+                POSTS_REPLIES: 0,
+                TOTAL_POSTS: 0,
+                THEMATICS: []
```
ayyazdaniaryan: Please change this to `Idea`, we're trying to move away from the name `Thematic`, as this object type no longer exists in backend (and will be deprecated on front-end). Also, I would no only add the Idea title as you've done, I would also add the `idea_id`, so that consultants can create pivot tables based on that column.

assembl/views/api2/discussion.py
```diff
@@ -667,6 +667,143 @@ def multi_module_csv_export(request):
...
+            }
+
+        #Get SSO extra information
+        for account in user.social_accounts:
+            for key, value in account.extra_data.iteritems():
+                if key not in fieldnames:
+                    fieldnames.append(key)
```
ayyazdaniaryan: Is this a desired way? Or do you want to create one column for extra data, and serialize the object into the column in a more human readable format?

cedrtang: It's the format asked by Fred: to have one column per extra data

assembl/tests/views_tests/test_api2.py
```diff
@@ -1107,6 +1120,133 @@ def test_base(self, test_session, test_app, discussion, user_language_preference
...
+        for idx, value in enumerate(self.participant1_data):
+            # If date, remove the seconds to compare
+            if idx in [3, 4, 5]:
+                assert last_row[idx][:-3] == value
+            else:
+                assert last_row[idx] == value
+
+
+    def test_anon(self, test_session, test_app, discussion, user_language_preference_en_cookie, participant1_user, participant1_username, agent_status_in_discussion_user1_visits):
+        discussion.creation_date = datetime(year=1999, month=12, day=1)
```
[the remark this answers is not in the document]
cedrtang: `freeze_time` doesn't seem to work for my tests

assembl/views/api2/discussion.py
cedrtang: @ayyazdaniaryan If I was only looking for the count of sentiment per type per post, yes it would be easier. But I also have to determine the count of sentiment per type per user so it would require another query to get this info. Instead, I fetch all sentiments (only access DB once) then I process the data to get both per post and per user.

PR #788, opened 2019-08-20
Remove tags when we remove a post

tdesvenain:
fixes DEVAS-2195


tdesvenain:
i don't understand why travis error remains @ayyazdaniaryan

PR #791, opened 2019-08-21
Export contribution: remove invalid and control characters

cedrtang:


assembl/views/api2/discussion.py
```diff
@@ -861,7 +862,10 @@ def escapeit(r, data):
 
     if something:
         if isinstance(something, basestring):
-            return something.decode('utf-8')
+            # Clean non unicode characters like emojis
+            sanitized = something.decode('utf-8').encode('latin-1', errors='ignore').decode('latin-1')
```
ayyazdaniaryan: This is a very destructive act for non-latin based languages. So, by this act alone, you both need to comment that non-latin languages are out of the question, but you also need to set a filter now on the API caller for exports that use this `escapeit` function to bail out with an error, or do this in another way. 

Since Assembl supports multiple non-latin languages, such as Russian, Simplified and Traditional Chinese, and Japanese, I think this is not the best way to go about exports. The common use case won't error, as the export will likely be in French or English. BUT then the exports will be really meaningless in non-latin languages...

I suggest you do something like this instead:
```
def remove_emoji(string):
    emoji_pattern = re.compile("["
                           u"\U0001F600-\U0001F64F"  # emoticons
                           u"\U0001F300-\U0001F5FF"  # symbols & pictographs
                           u"\U0001F680-\U0001F6FF"  # transport & map symbols
                           u"\U0001F1E0-\U0001F1FF"  # flags (iOS)
                           u"\U00002702-\U000027B0"
                           u"\U000024C2-\U0001F251"
                           "]+", flags=re.UNICODE)
    return emoji_pattern.sub(r'', string)
```

(Taken from https://stackoverflow.com/questions/33404752/removing-emojis-from-a-string-in-python)

cedrtang: You're right, I didn't think of those ones. I ll fix it

cedrtang:
@ayyazdaniaryan I did a regex to remove emojis but different than yours as we re on python2 and a narrow build

PR #804, opened 2019-09-19
Remove this dead and dangerous route

gary-bluenove:
SNCF security breach. Doesn't actually appear to be used (GET route)


tdesvenain:
You should squash those commitd

PR #808, opened 2019-09-25
lib/sqla: use .ini configuration

ChloeTigre:
The non-RDS use case made use of some separate variables to build a
psycopg2 connection string.

The make_url function from sqlalchemy is here to parse entire
connection strings, provided they are valid, into URL objects that can
then be fed to SQLAlchemy.

I propose that we use properly formatted SQLAlchemy URLs and stop using
directly the separate variables.

RFC1738 URLs in the SQLAlchemy context need the port to be specified and
will not be considered valid otherwise. This is a limitation we have to
work with and I suggest we do so.

Example invalid connection string as used before:
`postgresql+psycopg2://%(db_user)s:%(db_password)s@%(db_host)s/%(db_database)s?sslmode=disable`

Example valid connection string as I propose we use:
`postgresql+psycopg2://%(db_user)s:%(db_password)s@%(db_host)s:%(db_port)s/%(db_database)s?sslmode=disable`

This requires the following definition in the DEFAULT section:

`db_port = 5432`

### DIFF
--- a/assembl/configs/base_env.rc
+++ b/assembl/configs/base_env.rc
@@ -2,11 +2,13 @@ ini_files = production.ini RANDOM:random.ini.tmpl:saml_random.ini.tmpl RC_DATA
 # The file that will hold the generated random keys.
 random_file = random.ini
 *db_host = localhost
+*db_port = 5432
 *db_user = assembl
 # TODO: If generated remotely, fetch in fabfile
 *db_password = assembl
 *db_database = assembl
 db_schema = public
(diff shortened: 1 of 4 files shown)


ChloeTigre:
Note: this PR implies a change in default configuration. I must add a `db_port`  parameter to a .rc file. @ayyazdaniaryan can you advise on this? (which file)

ChloeTigre:
Proposed a patch. See commit comments.

ChloeTigre:
Backend CI worked properly, apparently front-end build is broken for the moment.
I will merge if it's ok for you @ayyazdaniaryan

ayyazdaniaryan:
It's alright, the front-end crash isn't related to this PR. Go ahead for merge.

PR #809, opened 2019-09-26
sqla: more compatibility

ChloeTigre:
this adds support for a sqlalchemy.url_ro (for read replicas).
this also makes sqla more tolerant to non-updated configs which have
not set a proper sqlalchemy.url.


ChloeTigre:
K, merging into develop.

PR #812, opened 2019-10-15
Security report, do not disclose info

gary-bluenove:
Based upon Decathlon security report - when using a user that does not exist we tell the user this - we should send a generic message in order to give a hacker no clues. Same for the password - generic is the new black baby


gary-bluenove:
wow thanks - I did not know that existed - we need somewhere to find all this good info. I will remove the branch

PR #815, opened 2019-11-05
Hashtags filter for decathlon #DEVAS-2241 

tdesvenain:


assembl/alembic/versions/fc715ad23538_langstring_add_hashtags.py
```diff
@@ -0,0 +1,45 @@
...
+        op.add_column('langstring_entry',
+                      sa.Column('hashtags', postgresql.ARRAY(sa.String)),
+                      )
+
+    # Do stuff with the app's models here.
+    from assembl import models as m
+    db = m.get_session_maker()()
+    try:
+        with transaction.manager:
+            query = db.query(m.LangStringEntry).filter(m.LangStringEntry.value.contains(u'#'))
```
ayyazdaniaryan, line 35: This query might lead you down dangerous roads, might want to use a regex to be more clear:
```
query = db.query(m.LangStringEntry).filter(m.LangStringEntry.op('SIMILAR TO')(u'^#[\w\d]+'))
```

Or something to that effect

assembl/graphql/idea.py
```diff
@@ -354,8 +355,12 @@ class Meta:
                                           type=graphene.Boolean,
                                           required=False,
                                           description=docs.Idea.my_posts_and_answers,
-
-                                      ))  # use dotted name to avoid circular import  # noqa: E501
+                                      ),
+                                      hashtags=graphene.Argument(
```
ayyazdaniaryan, line 359: Can remove the empty space below

assembl/graphql/schema.py
```diff
@@ -159,6 +164,40 @@ def resolve_tags(self, args, context, info):
...
+        from sqlalchemy.orm import aliased
+
+        discussion_id = context.matchdict['discussion_id']
+        idea_id = args.get('idea_id')
+        idea_id = int(Node.from_global_id(idea_id)[1])
+        idea = get_query(models.Idea, context).get(idea_id)
+        idea_posts_query = idea.get_related_posts_query()
+        idea_posts_subquery = idea_posts_query.with_entities(models.Post.id).subquery('idea_posts')
+        all_posts = aliased(models.Post, name='all_posts')
+        hashtag_aggs = func.array_agg(models.LangStringEntry.hashtags)
```
ayyazdaniaryan: Okay, you're using this method very frequently, so why don't you make the hashtag a class property under `LangStringEntry` and return this. Little encapsulation that can go a long way

assembl/graphql/schema.py
```diff
@@ -159,6 +164,40 @@ def resolve_tags(self, args, context, info):
...
+            hashtag_aggs.label('hashtags'),
+        ).join(
+            idea_posts_subquery,
+            all_posts.id == idea_posts_subquery.c.id
+        ).outerjoin(
+            models.LangStringEntry,
+            models.LangStringEntry.langstring_id == all_posts.body_id
+        ).filter(
+            all_posts.discussion_id == discussion_id,  # optimisation
+            func.array_length(models.LangStringEntry.hashtags, 1) > 0,  # can't aggregate null or empty arrays
```
ayyazdaniaryan: This one too, is a filter that is used frequently now. You can make this a member of `LangStringEntry` and you can simply call it in, like we do with `get_discussion_condition()` on a `Post` type.

assembl/static2/js/app/components/debate/common/deletePostButton.jsx
```diff
@@ -87,4 +88,21 @@ DeletePostButton.defaultProps = {
...
+  withRouter,
+  graphql(deletePostMutation, {
+    name: 'deletePost',
+    options: (props) => {
+      const ideaId = props.questionId
+        || props.themeId
+        || (props.params
+          && (props.params.themeId || props.params.questionId));
+      return {
+        refetchQueries: [{
```
ayyazdaniaryan, line 101: A re-render of all visible posts on a single post deletion or not?

assembl/static2/js/app/components/debate/common/post/postBody.jsx
```diff
@@ -154,7 +167,7 @@ export const Html = (props: HtmlProps) => {
           extractState: extract.extractState || EMPTY_STRING,
           nature: extract.extractNature || EMPTY_STRING
         });
-        const wrapper = jQuery(`<annotation data-extractInfo='${extractInfo}'></annotation>`);
+        const wrapper = jQuery(`<annotation data-extractInfo='${extractInfo}'/>`);
```
ayyazdaniaryan: Really? 😬

assembl/static2/js/app/components/debate/common/postsFilter/policies.js
```diff
@@ -123,26 +123,41 @@ export const fullDisplayPolicy: PostsDisplayPolicy = {
 export const summaryDisplayPolicy: PostsDisplayPolicy = {
   labelMsgId: 'debate.thread.postsDisplay.summary',
   id: 'display-summary',
-  displayMode: 'summary'
+  displayMode: 'summary',
+  type: 'choice'
```
ayyazdaniaryan, line 127: 👍

assembl/static2/js/app/components/debate/common/postsFilter/thread/hashtagFilter.jsx
```diff
@@ -0,0 +1,47 @@
...
+  setPostsFilterHashtags: (hashtags: string[]) => dispatch(setThreadPostsFilterHashtags(hashtags))
+});
+
+export default compose(
+  connect(mapStateToProps, mapDispatchToProps),
+  withRouter,
+  graphql(HashtagsQuery, {
+    options: props => ({
+      variables: {
+        ideaId: props.themeId || (props.params && props.params.themeId) || ''
```
ayyazdaniaryan, line 27: Ahhh, is this the reason why you created multiple of these dumb components for each type of page? Otherwise, could have been an HOC element, that would transform the props according to page type, to keep things DRY, though at cost of readability

assembl/static2/js/app/components/debate/common/postsFilter/menuItem.jsx
```diff
@@ -23,7 +23,8 @@ class DumbPostsFilterMenuItem extends React.Component<PostsFilterMenuItemProps>
     const { inputName, inputType, item, selected } = this.props;
     return (
       <MenuItem id={`postsFilterItem-${item.id}`} onClick={this.handleClick}>
-        <input id={`postsFilterItem-${item.id}-input`} type={inputType} name={inputName} checked={selected} />&nbsp;
+        <input id={`postsFilterItem-${item.id}-input`} onChange={() => {}} type={inputType} name={inputName} checked={selected} />
+        &nbsp;
```
ayyazdaniaryan: Needed?

assembl/static2/js/app/utils/linkify.js
```diff
@@ -8,8 +8,14 @@ function addIframeForStoryChief(html: string): string {
   return html.replace(url, '<iframe class="synthesis-iframe" src="$2"></iframe>');
 }
 
-export function transformLinksInHtml(html: string): string {
+const HASHTAG_REGEX = /(\s|^|\n|>)(#[A-Za-z0-9\-_]+)/g;
```
ayyazdaniaryan, line 11: I'm not familiar with the caret in between two pipes, what's the significance of this?

assembl/static2/js/app/utils/translations.js
```diff
@@ -2677,6 +2682,10 @@ const Translations = {
         backToQuestions: "Back to questions"
       },
       thread: {
+        hashtagsFilter: {
+          noOption: "No tag has been used in this section for the moment",
+          placeholder: "Search your city, your country. For instance #Lille..."
```
ayyazdaniaryan: `"Search for your city, your country, etc. For instance, #Lille ..."`

assembl/static2/js/app/utils/translations.js
```diff
@@ -2677,6 +2682,10 @@ const Translations = {
         backToQuestions: "Back to questions"
       },
       thread: {
+        hashtagsFilter: {
+          noOption: "No tag has been used in this section for the moment",
```
ayyazdaniaryan: `No tags have been used ...`

assembl/tests/views_tests/test_graphql_idea_hashtags.py
```diff
@@ -0,0 +1,42 @@
+# -*- coding: utf-8 -*-
+from freezegun import freeze_time
+from graphql_relay import from_global_id
+
+from assembl.graphql.schema import Schema as schema
+from assembl.tests.fixtures.graphql import create_post_in_thread
+
+
+@freeze_time("2018-3-1")
+def test_idea_hashtags(graphql_request, request, test_session,
```
ayyazdaniaryan, line 10: You should also add a test for the edge case of only having a `#` with nothing else passed in. Eg.
```
post.body.entries[0].value = "The quick brown #Fox jumps over the lazy #".
```

assembl/alembic/versions/fc715ad23538_langstring_add_hashtags.py
tdesvenain: the filter can be very permissive : the regex used by analyse_hashtags_onupdate will be tested before any actual change. this is for optimization only

assembl/static2/js/app/components/debate/common/deletePostButton.jsx
tdesvenain: it only refreshes the hashtag select input options

assembl/static2/js/app/utils/linkify.js
tdesvenain: i mean here : a space, or **the start of the string** or a new line or a ">"

PR #820, opened 2019-11-25
Update flow config; fix file extension

trouba:
Flow errors:
I had some flow errors on commits, but it seemed to be on files that should't be checked by flow. 

According to [this article](https://medium.com/shopback-engineering/5-things-you-might-be-interested-in-your-flow-config-1ad35b023e93), we should probably not use [ignore] but [untyped] instead (I suspect it's because of this misunderstanding that the flowconfig file looked like this).

File extension:
[This support ticket](https://bluenove.atlassian.net/browse/SUP-542) made me realise that if the file extension is in uppercase (PDF, JPG, ...) we don't recognize it and use a default image. This should fix it


ayyazdaniaryan (approved):
LGTM - wait for travis to pass on the frontend parts at least.

PR #821, opened 2019-11-26
Refactor filters dropdown

trouba:
This PR refactors the dropdown filter component for multiples reasons:

- this tickets shows that there are usability issues https://bluenove.atlassian.net/jira/servicedesk/projects/SUP/queues/custom/1/SUP-541
- the component used a navigation react-component encapsulating filters in <a>, creating a very bad html semantic


assembl/static2/js/app/components/debate/common/postsFilter/filtersCategoryLabel.jsx
```diff
@@ -0,0 +1,28 @@
+// @flow
+import * as React from 'react';
+import { Translate } from 'react-redux-i18n';
+
+type Props = {
+  labelMsgId: string
+};
+
+class FiltersCategoryLabel extends React.Component<Props> {
```
ayyazdaniaryan: Please change this to a functional component, as you're not using state and other hooks, no class is needed.

ayyazdaniaryan:
```
const FiltersCategoryLabel = ({labelMsgId}: Props) => {
   const categoryLabelStyle = {
      fontWeight: 'bold',
      textAlign: 'left',
      fontSize: '14px',
      padding: '20px 0 5px 20px',
      color: '#777'
    };
    return (
      <div style={categoryLabelStyle}>
        <strong>
          <Translate value={labelMsgId} />
        </strong>
      </div>
    );
  }
};
```

assembl/static2/js/app/components/debate/common/postsFilter/filtersCategoryLabel.jsx
```diff
@@ -0,0 +1,28 @@
...
+  render() {
+    const categoryLabelStyle = {
```
ayyazdaniaryan: Hmmm...adhoc css in js can be a bit dangerous if during refactoring of the styling in the css misses the rules here. I'm not sure how to feel about this...

ayyazdaniaryan: Okay no, please move this css into the css folder and use classnames. Adhoc changes in CSS will cause a ton of headaches downstream, especially with the ambition of cleanup of CSS once the frontend V2 is removed from V2.

For sake of reduced comments, please move all css-in-js to the css folder, to the appropriate place and use classnames.

ayyazdaniaryan (changes requested):
In V2 alone, we're not yet ready for CSS-in-JS. So, I ask that you move them to the css folder for now. There's work in progress right now to clean up the css by @felixsaintgenis. Moving the CSS to JS will only inhibit this work.

ayyazdaniaryan:
Otherwise, good work 👍

assembl/static2/js/app/components/debate/common/postsFilter/filtersCategoryLabel.jsx
trouba: What is the appropriate place ? 
the css is split up between files, sometimes a same class is in multiple places, and there is no indications of good practice on where this should go. I'd rather not create new classes and keep styles locales than add some classes on random files

trouba: "Adhoc changes in CSS will cause a ton of headaches downstream" < ?

trouba: I also found hundred of occurrences of style used directly on a component, often with a className, so I think we're pretty safe if i use some local style without classname on a component ?

ayyazdaniaryan: We didn't have pattern of putting entire css of a component inline in the `style`. We did use `style` for dynamic css such as colour changes, position changes, etc. You should aim to create an scss class following the (not so great) pattern in the css folder: create a new file for your components such as `filter.scss` and add your css as a class there. Then use the classnames. Webpack will load your css for you. You just have to ensure you `@import` your new file in `_assembl_base_styles.scss` for webpack to load it up

assembl/static2/js/app/components/debate/common/postsFilter/filtersCategoryLabel.jsx
trouba: Ok, tried it, I can't, there are too many embedded classes at different places that conflict with my css if I put it in a separate file, so I will close this PR for now and I will come back to it when the css refacto is done

assembl/static2/js/app/components/debate/common/postsFilter/filtersCategoryLabel.jsx
trouba: Ok, tried it, I can't, there are too many embedded classes at different places that conflict with my css if I put it in a separate file, so I will close this PR for now and I will come back to it when the css refacto is done

